PR #229, opened 2023-04-13
feat: create irc request

liorzilp:


mtsitrin (commented):
U actually use this file? it's ignite based configuration

config.yml
mtsitrin, line 1: U actually use this file? it's ignite based configuration

liorzilp: I think `ignite chain serve` using it. I'll revisit it later on.

ItzhakBokris (approved):

PR #241, opened 2023-04-24
feat: added /healthcheck endpoint that return 200OK if hub's height progress…

mtsitrin:
…ed in the last minute

# PR Standards

## Opening a pull request should be able to meet the following requirements:

---

For Author:

- [x]  Targeted PR against correct branch
- [x]  Linked to Github issue with discussion and accepted design
- [x]  Targets only one github issue

---

For Reviewer:

- [ ]  confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ]  Reviewers assigned
- [ ]  confirmed all author checklist items have been addressed

--- 

After reviewer approval:

- [ ]  In case targets main branch, PR should be squashed and merged.
- [ ]  In case PR targets a release branch, PR should be rebased.


liorzilp:
Please check: https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/07-invariants.md
I think this is the right way to do it

app/healthcheck.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+const progressCheckPeriod = 1 * time.Minute
+
+func HealthcheckRegister(clientCtx client.Context, r *mux.Router) {
+	go RunHealthCheck(clientCtx)
+	r.HandleFunc("/healthcheck", HealthcheckRequestHandlerFn(clientCtx)).Methods("GET")
+}
+
+func getLatestHeight(clientCtx client.Context) (int64, error) {
```
mtsitrin, line 25: FYI, i've took this from the `/block` API implementatiob

liorzilp: Invariants can be checked manually via [messages](https://github.com/cosmos/cosmos-sdk/blob/main/docs/docs/building-modules/02-messages-and-queries.md), but most often they are checked automatically at the end of each block.
So it can also be used as a trigger. Like updating the time at the beginning of the block and trigger checking if there was no update for some time.
I am good with the two options, up to you.

omritoptix (approved):

PR #369, opened 2023-10-26
feat: implement streamer module

mtsitrin:
# PR Standards

## Opening a pull request should be able to meet the following requirements:

---

For Author:

- [ ]  Targeted PR against correct branch
- [ ]  Linked to Github issue with discussion and accepted design
- [ ]  Targets only one github issue

---

For Reviewer:

- [ ]  confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ]  Reviewers assigned
- [ ]  confirmed all author checklist items have been addressed

---;

After reviewer approval:

- [ ]  In case targets main branch, PR should be squashed and merged.
- [ ]  In case PR targets a release branch, PR should be rebased.


x/streamer/keeper/hooks.go
```diff
@@ -0,0 +1,73 @@
...
+func (k Keeper) BeforeEpochStart(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	return nil
+}
+
+// AfterEpochEnd is the epoch end hook.
+func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	streams := k.GetUpcomingStreams(ctx)
+	// move to active if it's correct epoch and start time reached
+	for _, stream := range streams {
+		if epochIdentifier != stream.DistrEpochIdentifier {
```
omritoptix, line 20: If created a stream to start at 13:05 (and distribute every hour) , and now it’s 12:55 if I understand correctly it will only start at the next time the hour epic will fire, i.e at 14? If I understand correctly we're gonna remove streams which are supposed to start based on their epoch, though their start time is not necessarily related.

x/streamer/keeper/distribute.go
```diff
@@ -0,0 +1,174 @@
...
+// the distrInfo struct. It also updates the stream for the distribution.
+func (k Keeper) distributeInternal(
+	ctx sdk.Context, stream types.Stream,
+) (sdk.Coins, error) {
+	totalDistrCoins := sdk.NewCoins()
+	remainCoins := stream.Coins.Sub(stream.DistributedCoins...)
+	remainEpochs := uint64(stream.NumEpochsPaidOver - stream.FilledEpochs)
+
+	for _, coin := range remainCoins {
+		epochAmt := coin.Amount.Quo(sdk.NewInt(int64(remainEpochs)))
```
omritoptix, line 85: If the amount of tokens is lower than epochs  (i.e 29/30) we won't divide the remaining balance. What is the desired logic here?  seems like it's just gonna stay and "die" with the stream?

x/streamer/keeper/distribute.go
```diff
@@ -0,0 +1,174 @@
...
+	totalDistributedCoins := sdk.Coins{}
+	for _, stream := range streams {
+		streamDistributedCoins, err := k.distributeInternal(ctx, stream)
+		if err != nil {
+			return nil, err
+		}
+		totalDistributedCoins = totalDistributedCoins.Add(streamDistributedCoins...)
+	}
+
+	k.checkFinishDistribution(ctx, streams)
```
omritoptix, line 145: why iterate again and not just check it in the same loop vs doing 2 loops? 
probably in the `updateStreamPostDistribute` the checkDistribution can be done and then no need for the comparison with the +1 in the epochs (line 154).

x/streamer/keeper/distribute.go
```diff
@@ -0,0 +1,174 @@
...
+	return totalDistributedCoins, nil
```
omritoptix, line 146: Why returning totalDistributedCoins if not using it or logging it? If just for test purposes ?

x/streamer/client/cli/flags.go
```diff
@@ -0,0 +1,23 @@
...
+	FlagEpochs          = "epochs"
+	FlagOwner           = "owner"
+)
+
+// FlagSetCreateStream returns flags for creating gauges.
+func FlagSetCreateStream() *flag.FlagSet {
+	fs := flag.NewFlagSet("", flag.ContinueOnError)
+
+	fs.String(FlagStartTime, "", "Timestamp to begin distribution")
+	fs.String(FlagEpochIdentifier, "day", "Epoch identifier to begin distribution (e.g. 'day', 'week')")
```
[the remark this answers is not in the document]
mtsitrin: yes.
the supported epochs are defined in the epochs module params on genesis

x/streamer/keeper/stream.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	if !coins.IsAllPositive() {
+		return 0, fmt.Errorf("all coins %s must be positive", coins)
+	}
+
+	moduleBalance := k.bk.GetAllBalances(ctx, authtypes.NewModuleAddress(types.ModuleName))
+	spendedCoins := k.GetModuleToDistributeCoins(ctx)
```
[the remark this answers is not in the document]
mtsitrin: renamed. it's the tokens already already allocated to other streams

x/streamer/keeper/distribute.go
```diff
@@ -0,0 +1,174 @@
...
+	return coins
+}
+
+// getToDistributeCoinsFromStreams returns coins that have not been distributed yet from the provided streams
+func (k Keeper) getToDistributeCoinsFromStreams(streams []types.Stream) sdk.Coins {
+	coins := sdk.Coins{}
+	distributed := sdk.Coins{}
+
+	for _, stream := range streams {
+		coins = coins.Add(stream.Coins.Sort()...)
```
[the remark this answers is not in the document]
mtsitrin: the `coins` interface expects sorted coins
found out in UT

x/streamer/keeper/distribute.go
mtsitrin: the stream gonna get to finished state, so those token will be calculated in the "free" token of the streamer... they won't be stuck on the finished stream

x/streamer/keeper/distribute.go
mtsitrin: currently yeah, I can log it as well

x/streamer/keeper/stream.go
```diff
@@ -0,0 +1,197 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// SetStreamWithRefKey takes a single stream and assigns a key.
+// Takes combinedKey (the keyPrefix for upcoming, active, or finished streams combined with stream start time) and adds a reference to the respective stream ID.
+// If this stream is active or upcoming, creates reference between the denom and stream ID.
+func (k Keeper) SetStreamWithRefKey(ctx sdk.Context, stream *types.Stream) error {
```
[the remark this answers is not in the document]
mtsitrin: the module keeps the finished stream in the store as sell

x/streamer/keeper/hooks.go
omritoptix: I see. not sure it's a desired behavior. IMO I'd expect a stream to start at the relevant hour defined. 
If the epoch is every week for example? could it be postponed in a few days until the week epoch will end?

x/streamer/keeper/stream.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	if !coins.IsAllPositive() {
+		return 0, fmt.Errorf("all coins %s must be positive", coins)
+	}
+
+	moduleBalance := k.bk.GetAllBalances(ctx, authtypes.NewModuleAddress(types.ModuleName))
+	spendedCoins := k.GetModuleToDistributeCoins(ctx)
+
+	if !coins.IsAllLTE(moduleBalance.Sub(spendedCoins...)) {
```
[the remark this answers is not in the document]
omritoptix: @mtsitrin ?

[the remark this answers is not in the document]
mtsitrin: yes, it's `smaller or equal amount`
anyway I've added UT to validate it, u can see

x/streamer/keeper/hooks.go
mtsitrin: well if it's a week, it should distributed every week.
the count is not relative to the start time. it's by based on epoch triggers.

so the stream will distribute on the next epoch after the start time. make sense for me.
The only fix is that I'll change the stream to be `Active` when time come, regardless of the epoch

omritoptix (approved):

PR #424, opened 2023-11-23
feat: streamer additional gov proposals

mtsitrin:
- [x] gov proposal to stop existing stream 
- [x] gov proposal to modify distribution weights of existing stream
- [x] gov proposal to replace distribution weights of existing stream


x/streamer/client/cli/tx_pause_stream.go
```diff
@@ -0,0 +1,58 @@
...
+		Use:   "stop-stream-proposal streamID [flags]",
+		Short: "proposal to stop an exisiting stream",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			streamID, err := strconv.ParseUint(args[0], 10, 64)
```
omritoptix, line 28: prob worth checking also if the stream exists before a proposal is sent (also applied to tx_replace and tx_update proposals)

x/streamer/keeper/keeper_replace_update_distribution.go
```diff
@@ -0,0 +1,79 @@
...
+		recordsMap[existingRecord.GaugeId] = existingRecord
+	}
+
+	for _, record := range records {
+		recordsMap[record.GaugeId] = record
+	}
+
+	newRecords := []types.DistrRecord{}
+	for _, val := range recordsMap {
+		if !val.Weight.Equal(sdk.ZeroInt()) {
```
omritoptix, line 57: what is it's minus value? i.e weight = -3?

omritoptix (commented):
1. File called tx_stop_stream. the term stop is a bit confusing as it seems like it can be started again. IMO a better term can be found e.g terminate/etc. 
2. I think  moving a stopped stream  to “finished” state can be confusing when querying a stream as it may be thought of as it was ongoing and completed. Possibly terminated/deactivated etc more suitable. related to my previous comment. 
3. What about the balance of the tokens in the streams once it’s prematurely terminated? maybe missed that part of the logic.

x/streamer/client/cli/tx_pauseStream.go
```diff
@@ -0,0 +1,58 @@
...
+		Use:   "stop-stream-proposal streamID [flags]",
+		Short: "proposal to stop an exisiting stream",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+
+			streamID, err := strconv.ParseUint(args[0], 10, 64)
```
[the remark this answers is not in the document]
mtsitrin: it doesn't have access to the app in this stage. it's just creates the proposal message actually.
the validation logic happens in the handler

x/streamer/client/cli/tx_pause_stream.go
mtsitrin: it doesn't have access to the app in this stage. it's just creates the proposal message actually.
the validation logic happens in the handler

x/streamer/keeper/keeper_replace_update_distribution.go
mtsitrin: it's valdiated in `distrInfo, err := types.NewDistrInfo(newRecords)`

mtsitrin:
> File called tx_stop_stream. the term stop is a bit confusing as it seems like it can be started again. IMO a better term can 

be found e.g terminate/etc.
will change

> I think moving a stopped stream to “finished” state can be confusing when querying a stream as it may be thought of as it was ongoing and completed. Possibly terminated/deactivated etc more suitable. related to my previous comment.

I don't want to introduce additional changes in the key store for it.
u will see from the finished stream data that's it been terminated (`filled epochs < num epoch to dist`)

> What about the balance of the tokens in the streams once it’s prematurely terminated? maybe missed that part of the logic.

not sure what u mean.
the tokens for this stream are "released" for further calculations (see `TestStopStream_ModuleToDistributeCoins`)

omritoptix (commented):
the cli file is still named `tx_pause_stream` probably better to change to `tx_terminate_stream` and also the `Use` and `Short` fields of the command.

omritoptix (approved):

PR #453, opened 2023-12-13
feat(test): add new integration test suite utilities

VictorTrustyDev:
This is the new Integration test suite utilities I developed, hope this will enhance quality of Dymension production.

There are testcases which are falling, I marked them with `fixme` because as my expectation, the testcase should pass, for example about denom validation.

I also prepared contract deployment methods those are ready to be used in case you want to enable contract creation in the future.

_The integration test suite, developed by Escan team, with knowledge from test suites of Cosmos-SDK, ibc-go and Evmos-SDK._

### DIFF
--- a/removed.txt
+++ b/removed.txt
@@ -1,0 +1,1 @@
+Content was removed


integration_test_util/types/temporary.go
```diff
@@ -0,0 +1,43 @@
...
+	return &TemporaryHolder{}
+}
+
+func (h *TemporaryHolder) AddTempFile(file string) {
+	if len(file) < 1 {
+		return
+	}
+	if !strings.HasPrefix(file, "/tmp/") {
+		panic(fmt.Sprintf("temp file must be in '/tmp': %s", file))
+	}
```
VictorTrustyDev, line 25: Won't work correctly with Windows, feel free to enhance if anyone in the team using Windows.

PR #480, opened 2024-01-12
fix: Add support for delayedack middleware to also delay ack and timeout packets

omritoptix:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


x/delayedack/hooks.go
```diff
@@ -32,24 +32,37 @@ func (im IBCMiddleware) FinalizeRollappPackets(ctx sdk.Context, rollappID string
 	// Get the packets for the rollapp until height
 	logger.Debug("Finalizing IBC rollapp packets", "rollappID", rollappID, "state end height", stateEndHeight, "num packets", len(rollappPendingPackets))
 	for _, rollappPacket := range rollappPendingPackets {
-		logger.Debug("Finalizing IBC rollapp packet", "rollappID", rollappID, "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel())
+		logger.Debug("Finalizing IBC rollapp packet", "rollappID", rollappID, "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel(), "type", rollappPacket.Type)
 		// Update the packet status
 		im.keeper.UpdateRollappPacketStatus(ctx, rollappID, rollappPacket, types.RollappPacket_ACCEPTED)
```
mtsitrin, line 37: I think it's better to move it after the switch case, and set according to the error received
```
// Update the packet status
if err != nil {
	im.keeper.UpdateRollappPacketStatus(ctx, rollappID, rollappPacket, types.RollappPacket_REJECTED)
} else {
	im.keeper.UpdateRollappPacketStatus(ctx, rollappID, rollappPacket, types.RollappPacket_ACCEPTED)
}
```

proto/dymension/delayedack/rollapp_packet.proto
```diff
@@ -8,14 +8,22 @@ option go_package = "github.com/dymensionxyz/dymension/x/delayedack/types";
 
 message RollappPacket {
     ibc.core.channel.v1.Packet packet = 1;
+    bytes acknowledgement = 2;
+    bool isTimeout = 3;
```
mtsitrin, line 12: not used

proto/dymension/delayedack/rollapp_packet.proto
```diff
@@ -8,14 +8,22 @@ option go_package = "github.com/dymensionxyz/dymension/x/delayedack/types";
...
       ACCEPTED = 1;
       REJECTED = 2;
     }
-    Status status = 2;
-    uint64 ProofHeight = 3;
-    string error = 4;
-    bytes relayer = 5;
+    Status status = 4;
+    uint64 ProofHeight = 5;
+    string error = 6;
```
mtsitrin, line 20: not used

x/delayedack/hooks.go
```diff
@@ -32,24 +32,37 @@ func (im IBCMiddleware) FinalizeRollappPackets(ctx sdk.Context, rollappID string
...
+		// Call the relevant callback for each packet
+		switch rollappPacket.Type {
+		case types.RollappPacket_ON_RECV:
+			logger.Debug("Calling OnRecvPacket", "rollappID", rollappID, "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel())
+			ack := im.app.OnRecvPacket(ctx, *rollappPacket.Packet, rollappPacket.Relayer)
+			// Write the acknowledgement to the chain only if it is synchronous
+			if ack != nil {
+				_, chanCap, err := im.keeper.LookupModuleByChannel(ctx, rollappPacket.Packet.DestinationPort, rollappPacket.Packet.DestinationChannel)
+				if err != nil {
+					logger.Error("Error looking up module by channel", "rollappID", rollappID, "error", "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel(), "error", err.Error())
```
mtsitrin, line 47: bad key-value for logger. redundent `"error"`

proto/dymension/delayedack/rollapp_packet.proto
omritoptix: Added usage for the 'error' field and fixed a bug where the methods weren't wrapped with `ApplyFuncIfNoError`

proto/dymension/delayedack/rollapp_packet.proto
mtsitrin: the `isTimeout` flag not used as I see it. 
timeout is set by the type (`RollappPacket_ON_TIMEOUT`)

x/delayedack/hooks.go
mtsitrin: yeah but do u want to mark it as ACCEPTED?
In normal case, if the app would return err, the relayer would retry

in our case the packet will be marked as accepted and forgotten (which might be fine if by design, but probably need to be marked somehow)

x/delayedack/hooks.go
```diff
@@ -32,24 +33,70 @@ func (im IBCMiddleware) FinalizeRollappPackets(ctx sdk.Context, rollappID string
...
-		ack := im.app.OnRecvPacket(ctx, *rollappPacket.Packet, rollappPacket.Relayer)
-		// Write the acknowledgement to the chain only if it is synchronous
-		if ack != nil {
-			_, chanCap, err := im.keeper.LookupModuleByChannel(ctx, rollappPacket.Packet.DestinationPort, rollappPacket.Packet.DestinationChannel)
+		rollappPacket := im.keeper.UpdateRollappPacketStatus(ctx, rollappID, rollappPacket, types.RollappPacket_ACCEPTED)
+		// Call the relevant callback for each packet
+		switch rollappPacket.Type {
+		case types.RollappPacket_ON_RECV:
+			logger.Debug("Calling OnRecvPacket", "rollappID", rollappID, "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel())
+			ack := im.app.OnRecvPacket(ctx, *rollappPacket.Packet, rollappPacket.Relayer)
```
mtsitrin, line 43: this call should be under the wrappedFunc as well

x/delayedack/hooks.go
```diff
@@ -32,24 +33,70 @@ func (im IBCMiddleware) FinalizeRollappPackets(ctx sdk.Context, rollappID string
...
+						return err
+					}
+					return nil
+				}
+				err := osmoutils.ApplyFuncIfNoError(ctx, wrappedFunc)
+				if err != nil {
+					logger.Error("Error writing acknowledgement", "rollappID", rollappID, "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel(), "error", err.Error())
+					// Update the packet with the error
+					rollappPacket.Error = err.Error()
+					im.keeper.SetRollappPacket(ctx, rollappID, rollappPacket)
```
mtsitrin, line 62: why u call `SetRollappPacket` instead of Update on errors?
it will create new rollapp instance with Pending Status

x/delayedack/ibc_middleware.go
```diff
@@ -163,7 +164,66 @@ func (im IBCMiddleware) OnAcknowledgementPacket(
...
+	ibcClientLatestHeight := clientState.GetLatestHeight()
+	finalizedHeight, err := im.keeper.GetRollappFinalizedHeight(ctx, chainID)
+	if err == nil && finalizedHeight >= ibcClientLatestHeight.GetRevisionHeight() {
+		logger.Debug("Skipping IBC transfer OnAcknowledgementPacket as the packet proof height is already finalized")
+		return im.app.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+	// Run the underlying app's OnAcknowledgementPacket callback
+	// with cache context to avoid state changes and report the acknowledgement result.
+	// Only save the packet if the underlying app's callback succeeds.
+	cacheCtx, _ := ctx.CacheContext()
```
mtsitrin, line 208: why u need it? u already run it with cacheTx by the caller

x/delayedack/hooks.go
omritoptix: IMO it should be marked as FINALIZED - as it is finalized.
as I previously wrote I think ACCEPTED is confusing term here.
so as I see it, the packet is finalized, but it could have errors.

as for the relayer - if we're talking about application layer error which are not part of out-of-sync problem (e.g LC headers didn't yet arrive - which is not the case in our example) - by default the relayer will not retry (i.e if you simply run `rly tx transfer` command). It's true that if you restart it will try to get all non-relayed packets but if the packet itself is maleformed or it's data invalid - retry won't help unless the relayer code or the blockchain state changes.

My point is that as I see it mostly application layer errors, specifically for ics-20 which are mostly around malformed data or lack of balances wouldn't change with relayer retry (which eventually get exhausted and the packet transfer would fail) so I think it's better to say: "Hi, this packet is finalized, but we couldn't apply it as it was malformed/not enough balance".

re forgotten - I think our UI should reflect any errors in the packet by looking at the error field.

x/delayedack/hooks.go
omritoptix: it shouldn't as the error is not part of the key generated for the packet so it should override existing one

proto/dymension/delayedack/rollapp_packet.proto
omritoptix: True! not necessary.

x/delayedack/ibc_middleware.go
omritoptix: Yea but than the caller controls it.
after it succeeds with cache context, it will actually run it without - which will make the ack run before the packet is finalized.
I want to force it to always run with cache context by the caller and run it with non-cached context only from the hook. 
Maybe there is a better way to do it?

x/delayedack/hooks.go
mtsitrin: I think the status=pending is part of the key

x/delayedack/ibc_middleware.go
mtsitrin: oh I see what u did here
what about the case err != nil?
we don't want to write the cacheCtx?

x/delayedack/ibc_middleware.go
```diff
@@ -163,7 +164,66 @@ func (im IBCMiddleware) OnAcknowledgementPacket(
...
+	err = im.app.OnAcknowledgementPacket(cacheCtx, packet, acknowledgement, relayer)
+	switch err {
```
mtsitrin, line 211: why u need switch case and not single if statement?

x/delayedack/ibc_middleware.go
omritoptix: can you elaborate? not sure what you mean by writing the cacheCtx.

omritoptix:
There is an upgrade hander required for FP because of the packet struct change. opened an issue: https://github.com/dymensionxyz/dymension/issues/529 
Should be fine for blumbus though

mtsitrin (approved):

PR #499, opened 2024-01-22
helper functions for delayed ack handling

mtsitrin:
* checking errors on delateyed packets and using `RollappPacket_REJECTED` on failed packets
* moved common code to wrapper functions


omritoptix:
changed moved to https://github.com/dymensionxyz/dymension/pull/559

PR #535, opened 2024-02-05
feat: delete finalized rollapp packets 

omritoptix:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

- [x]  Wrote unit and integration tests

---

For Reviewer:

- [ ]  Reviewers assigned

---

After reviewer approval:


x/delayedack/types/params.go
```diff
@@ -0,0 +1,52 @@
...
+// ParamSetPairs get the params.ParamSet
+func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
+	return paramtypes.ParamSetPairs{
+		paramtypes.NewParamSetPair(KeyEpochIdentifier, &p.EpochIdentifier, func(_ interface{}) error { return nil }),
+	}
+}
+
+// Validate validates the set of params
+func (p Params) Validate() error {
+	return nil
```
mtsitrin, line 45: epochidentifier != ""

x/delayedack/keeper/hooks.go
```diff
@@ -28,3 +34,38 @@ func (k eibcHooks) AfterDemandOrderFulfilled(ctx sdk.Context, demandOrder *eibct
...
+}
+
+// AfterEpochEnd is the epoch end hook.
+// We want to clean up the demand orders that are with underlying packet status which are finalized.
+func (e epochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumber int64) error {
+	if epochIdentifier != e.GetParams(ctx).EpochIdentifier {
+		return nil
+	}
+	// Get all rollapp packets with status FINALIZED and delete them
+	finalizedRollappPackets := e.ListRollappPacketsByStatus(ctx, commontypes.Status_FINALIZED, 0)
```
mtsitrin, line 65: will be more efficent to delete in the iterator
otherwise u iterate+unmarshal, and iterate again

x/common/types/rollapp_packet.go
mtsitrin, line 1: IMO move to `types/events.go`

x/delayedack/keeper/hooks.go
```diff
@@ -28,3 +35,40 @@ func (k eibcHooks) AfterDemandOrderFulfilled(ctx sdk.Context, demandOrder *eibct
...
+	for _, finalizedPacket := range finalizedRollappPackets {
+		finalizedPacketCopy := finalizedPacket
+		_ = osmoutils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
```
mtsitrin, line 69: log errors?

x/delayedack/keeper/rollapp_packet.go
```diff
@@ -114,35 +120,44 @@ func (k *Keeper) UpdateRollappPacketWithStatus(ctx sdk.Context, rollappPacket co
...
 	for ; iterator.Valid(); iterator.Next() {
 		var val commontypes.RollappPacket
-		k.cdc.MustUnmarshal(iterator.Value(), &val)
-		if val.ProofHeight <= maxProofHeight {
+		err := k.cdc.Unmarshal(iterator.Value(), &val)
+		if err != nil {
+			logger.Error("Failed to unmarshal rollapp packet", "error", err)
+			continue
+		}
+		if maxProofHeight == 0 || val.ProofHeight <= maxProofHeight {
```
mtsitrin, line 160: why not storing the PENDING keys by proofheight as well? to avoid iterating over all the PENDING packets when finalizing

x/delayedack/keeper/rollapp_packet.go
```diff
@@ -153,17 +168,38 @@ func (k Keeper) ListRollappPendingPackets(
...
 
 	// Iterate over the range from lastProofHeight to proofHeight
-	iterator := sdk.KVStorePrefixIterator(store, []byte{})
+	iterator := sdk.KVStorePrefixIterator(store, commontypes.AllRollappPacketKeyPrefix)
 	defer iterator.Close() // nolint: errcheck
 
 	for ; iterator.Valid(); iterator.Next() {
 		var val commontypes.RollappPacket
-		k.cdc.MustUnmarshal(iterator.Value(), &val)
+		err := k.cdc.Unmarshal(iterator.Value(), &val)
```
mtsitrin, line 179: why u don't want to panic here? corrupted store is blocker, isn't it?

x/eibc/types/demand_order.go
```diff
@@ -67,6 +71,17 @@ func (m *DemandOrder) Validate() error {
 	return nil
 }
 
+func (m *DemandOrder) GetEvents() []sdk.Attribute {
```
mtsitrin, line 74: events.go?

x/common/types/rollapp_packet.go
omritoptix: it has `GetEvents` which is a method on the `RollappPacket` struct. If more methods will be needed for `RollappPacket` stuct they will also go there.

x/delayedack/keeper/hooks.go
omritoptix: from the godoc of the `ApplyFuncIfNoError`:

```
// This function lets you run the function f, but if theres an error or panic
// drop the state machine change and log the error.
```

x/delayedack/keeper/rollapp_packet.go
omritoptix: yea actually I think there is even a bug here which is worse. There is an assumption that the packets are ordered by proof height (which I think is a left over from pre-keys change) and on the first violation of proofHeight the loop breaks. which is bad .

x/delayedack/keeper/rollapp_packet.go
omritoptix: yea we can panic here. this method is only called from init genesis.

x/eibc/types/demand_order.go
omritoptix: also same as before as it's part of this struct and I find it cleaner to define those methods on structs hence keeping it as part of the struct file. 
if you think it's better to move to events lets talk maybe it is I just don't see the reason.

x/delayedack/keeper/rollapp_packet.go
mtsitrin: don't u want to add the proofheight to the iterator?

mtsitrin (approved):

x/delayedack/keeper/rollapp_packet.go
omritoptix: it's iterating by order of proofHeight so once proofHeight > MaxProofHeight I just break the loop. 
trying to introduce endKey for the iterator is not straight forward as the proofHeight is in the middle of the rollappPacketKey bytes so can't be used as startKey/endKey.

mtsitrin (approved):

PR #629, opened 2024-03-05
feat: freeze rollapp after fraud

mtsitrin:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


proto/dymension/rollapp/state_status.proto
```diff
@@ -13,4 +13,6 @@ enum StateStatus {
   STATE_STATUS_RECEIVED = 1; 
   // STATE_STATUS_FINALIZED defines a rollapp state where the the "Dispute Period" has ended and this state is considered final
   STATE_STATUS_FINALIZED = 2; 
+  // STATE_STATUS_DISPUTED defines a rollapp state where the state update transaction was disputed by a participant
+  STATE_STATUS_DISPUTED = 3;
```
omritoptix, line 17: In common.type we call it "REVERTED". I think we should keep the same terminology. (and also possibly change the RECEIVED to PENDING).

I'd just use the commontypes.STATUS for all the statuses

x/rollapp/keeper/fraud_proposal.go
```diff
@@ -24,20 +24,27 @@ func (k Keeper) HandleFraud(ctx sdk.Context, rollappID, clientId string, height
...
 
-	// slash the sequencer
+	// slash the sequencer, clean delayed packets
 	err = k.hooks.FraudSubmitted(ctx, rollappID, height, seqAddr)
 	if err != nil {
 		return err
 	}
 
-	//FIXME: make sure the clientId corresponds to the rollappID
+	//mark the rollapp as frozen. revert all pending states to finalized
```
omritoptix, line 37: we call it "jailed" or "frozen"? terminology wise I think "frozen" is more suitable since "jailed" sounds like a temporary condition and also it adheres to the IBC client / Validator terminology (i.e frozen/jailed respectively)

x/rollapp/keeper/fraud_proposal.go
```diff
@@ -14,7 +14,7 @@ import (
 // HandleFraud handles the fraud evidence submitted by the user.
 func (k Keeper) HandleFraud(ctx sdk.Context, rollappID, clientId string, height uint64, seqAddr string) error {
 	// Get the rollapp from the store
-	_, found := k.GetRollapp(ctx, rollappID)
+	rollapp, found := k.GetRollapp(ctx, rollappID)
```
omritoptix, line 17: what if a rollapp is already frozen? prob want to error on that.

x/rollapp/keeper/fraud_proposal.go
```diff
@@ -24,20 +24,27 @@ func (k Keeper) HandleFraud(ctx sdk.Context, rollappID, clientId string, height
 		return err
 	}
 
-	//TODO: mark the rollapp as frozen (if immutable) or mark the fraud height to allow overwriting
-
 	if stateInfo.Sequencer != seqAddr {
```
omritoptix, line 27: Prob should also check the stateInfo is not finalized.

x/rollapp/keeper/fraud_proposal.go
mtsitrin: changed to frozen in proto

x/rollapp/keeper/fraud_proposal.go
mtsitrin: what happens if fraud submitted for height 100, accepted
but there's a fraud on height 90 as well, which is submitted later?

omritoptix: good point! seems that currently when iterating over states and finalizing them we simply check if the rollapp is frozen or not. 
in order to suport this use case we should probably also change to check at which height it's frozen and only mark those states as "REVERTED" and the rest as finalized (even if the rollapp is frozen) as we're saying it's frozen since specific height. 
need to make sure also that the IBC client is frozen from that height. 

So I'd say for your question check if frozen, what's the frozen height and only accepts disputes with height < frozen height.

mtsitrin:
left tasks:
* check for freeze height on double fraud submission
* move packets from PENDING to REVERTED
* UT

x/rollapp/keeper/fraud_handler.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	//mark the rollapp as frozen. revert all pending states to finalized
+	rollapp.Frozen = true
+	k.SetRollapp(ctx, rollapp)
+
+	lastFinalizedIdx, _ := k.GetLatestFinalizedStateIndex(ctx, rollappID)
+	finalizedStateInfo, _ := k.GetStateInfo(ctx, rollappID, lastFinalizedIdx.Index)
+
+	lastFinalizedHeight := finalizedStateInfo.StartHeight + finalizedStateInfo.NumBlocks
```
omritoptix, line 57: 1. I believe this is not the height you want. this is the height on the rollapp. This height can be much bigger than the currrent block height.  Ideally I think what you need here to get the latest finalized height on the hub (not on the rollapp) which can be taken by `currentHubHeight - disputePeriodInBlocks`. 
1. Also I believe the calc is wrong here. -1 is missing (if you have `StartHeight` 1 and `NumBlocks` 1 , the latest finalized height is 1 not 2)

x/rollapp/keeper/fraud_handler.go
```diff
@@ -0,0 +1,121 @@
...
+
+	//iterate over all height from the last finalized height to the current height
```
omritoptix, line 59: possibly cleaner to move the state revert  logic to a separate function (maybe on the rollapp struct)

omritoptix (approved):

PR #638, opened 2024-03-07
feat(evm): integrate virtual frontier bank contract

VictorTrustyDev:
Closes: #627

For the detailed information about the VFC, visit: https://github.com/dymensionxyz/ethermint/pull/11.
This PR integrate the feature into Dymension Hub.

Features:
- Virtual Frontier Bank Contract
- Auto deploy contract
  - For Dymension chains at begin epoch **day**, for IBC denom only (default)
  - For local-net, at block 1, for native denom

Incompatible issue:
- Dymension uses default proto of Cosmos, which is BaseAccount, not Ethermint's default is EthAccount so the expected VFBC contract code hash can not be set, leading to loss of mapping to the deployed bytecode.
  - The problem was solved by this [work around](https://github.com/dymensionxyz/ethermint/pull/11/commits/f5ecd0f9d6e44ba51299a07fc781f03ea5c2f2c8)
  - Currently still work without problem but unsure.
  - **TODO: In the future, if enable contract creation for DYM, need to migrate all the VFBC to use EthAccount and set the corresponding code hash and code.**

TODO:
- Upgrade Ethermint version upon [PR](https://github.com/dymensionxyz/ethermint/pull/11) merged.
- Generate swagger.
- Consider remove [this method](https://github.com/dymensionxyz/dymension/blob/eedd4102e1416aa16ad69a02b6b415ceab956ce1/x/evm/keeper/epoch_hooks.go#L58-L65), was added for testing purpose.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -3,6 +3,7 @@ package app
 import (
 	"encoding/json"
 	"fmt"
+	evmclient "github.com/evmos/ethermint/x/evm/client"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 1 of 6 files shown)


VictorTrustyDev:
Closed due to [integration spec changed](https://github.com/dymensionxyz/dymension/issues/658).

PR #641, opened 2024-03-09
feat: rollapp ibc state validation upon state update 

### LINKED ISSUE #258: add IBC's `RedundantRelayDecorator` ante handler
IBC repo provides the `RedundantRelayDecorator` ante handler:
```
// RedundantRelayDecorator returns an error if a multiMsg tx only contains packet messages (Recv, Ack, Timeout) and additional update messages
// and all packet messages are redundant. If the transaction is just a single UpdateClient message, or the multimsg transaction
// contains some other message type, then the antedecorator returns no error and continues processing to ensure these transactions
// are included. This will ensure that relayers do not waste fees on multiMsg transactions when another relayer has already submitted
// all packets, by rejecting the tx at the mempool layer.
```

We probably should add this handler to the hub's ante handlers

### LINKED ISSUE #643: Validate rollapp IBC state update against current rollapp state

srene:
## Description

----

The  objective of this PR is to prevent potential vector attacks that utilises other cosmos chains with the same chain id of an existing rollappid to enable malicious ibc-transfers to other chains that are not the rollapp itself.

For that, we need to prevent  that any ibc-transfer can use the chain id of an existing rollapp, without coming from the real rollapp, authenticating the rollapp id. In the following we list 3 potential solutions:

Solution 1:  Reject  channels creation with same id of rollapp id

Main drawback: it is not possible to prevent channel creations with the chain id equal to a  rollappid,  that is not yet registered in the hub, but it can be known beforehand by an attacker.

Solution 2:   Reject light clients update for a channel with a rollapp id equal to a registered rollapp id, when it does not come from the registered sequencer after the rollapp is created. By doing that, ibc will fail to accept packets from a channel that does not match rollapp sequencer info.

Main drawback: Headers received from an attacker channel until the first state info update from the valid rollapp sequencer, will be accepted as valid, being a vulnerability.

Solution 3: Checking the next validator hash in light client headers for the channel, corresponds to the same next validator hash created with a single validator, which is the sequencer registered for the rollapp in the state info.

Main drawback: The solution requires authenticating the rollapp id per ibc packet.

In this PR the solution 3 has been implemented, since even though is not the most efficient, it does not have the vulnerability issues of the first two.

Note that this is a short-term solution built with the assumptions of having a single honest sequencer for a rollapp.

Closes #643 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:

### DIFF
--- a/x/delayedack/ibc_middleware.go
+++ b/x/delayedack/ibc_middleware.go
@@ -128,6 +128,7 @@ func (im IBCMiddleware) OnRecvPacket(
 	err = im.keeper.ValidateRollappId(ctx, rollappID, packet.DestinationPort, packet.DestinationChannel)
 	if err != nil {
 		logger.Error("Failed to validate rollappID", "rollappID", rollappID, "err", err)
+		return channeltypes.NewErrorAcknowledgement(err)
 	}
 
 	proofHeight, err := im.GetProofHeight(ctx, packet)
@@ -194,6 +195,7 @@ func (im IBCMiddleware) OnAcknowledgementPacket(
 	err = im.keeper.ValidateRollappId(ctx, rollappID, packet.DestinationPort, packet.DestinationChannel)
 	if err != nil {
 		logger.Error("Failed to validate rollappID", "rollappID", rollappID, "err", err)
+		return err
 	}
 
 	proofHeight, err := im.GetProofHeight(ctx, packet)
@@ -263,6 +265,7 @@ func (im IBCMiddleware) OnTimeoutPacket(
 	err = im.keeper.ValidateRollappId(ctx, rollappID, packet.DestinationPort, packet.DestinationChannel)
 	if err != nil {
 		logger.Error("Failed to validate rollappID", "rollappID", rollappID, "err", err)
+		return err
 	}
 
 	proofHeight, err := im.GetProofHeight(ctx, packet)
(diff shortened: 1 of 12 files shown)


x/delayedack/ibc_middleware.go
```diff
@@ -250,12 +260,16 @@ func (im IBCMiddleware) OnTimeoutPacket(
 		return im.app.OnTimeoutPacket(ctx, packet, relayer)
 	}
 
+	err = im.keeper.ValidateRollappId(ctx, rollappID, packet.DestinationPort, packet.DestinationChannel)
+	if err != nil {
+		logger.Error("Failed to validate rollappID", "rollappID", rollappID, "err", err)
```
omritoptix, line 265: neeed to return here `NewErrorAcknowlegment`

omritoptix (commented):
if you fix what I wrote the tests don't pass

mtsitrin:
* I think u referenced the wrong issue here.
* Why running this code per IBC packets, and not on the stateUpdate itself?
The validation itself is not `packet` specific, no reason to run it per packet IMO

omritoptix:
@mtsitrin is correct. better running it per UpdateState.
I believe this could be handled using anteHandler if not better way found.
In case this check fails the updateState should be rejected which should also protect against malicious actors trying to pretend to be a rollapp chain.

srene:
added right issue number in the description

omritoptix (approved):

PR #650, opened 2024-03-11
fix: fix grace period finalization test

### LINKED ISSUE #649: current tests does not check if state infos are finalized correctly
With the recent modifications of removing finalized state infos from the BlockHeightToFinalizationQueue, there is no test that is checking state infos are finalized correctly

srene:
## Description
----

With the recent modifications of removing finalized state infos from the BlockHeightToFinalizationQueue, no current tests are checking state infos are finalized correctly. In this PR the TestUpdateState is updated to check that the state infos submitted to the hub get finalized accordingly with the specified dispute period

Closes #649

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:

### DIFF
--- a/x/rollapp/keeper/msg_server_update_state_test.go
+++ b/x/rollapp/keeper/msg_server_update_state_test.go
@@ -114,20 +114,19 @@ func (suite *RollappTestSuite) TestUpdateState() {
 		goCtx = sdk.WrapSDKContext(suite.Ctx)
 
 		// calc new updateState
-		expectedLatestStateInfoIndex, found := suite.App.RollappKeeper.GetLatestStateInfoIndex(suite.Ctx, rollapp.GetRollappId())
+		latestStateInfoIndex, found := suite.App.RollappKeeper.GetLatestStateInfoIndex(suite.Ctx, rollapp.GetRollappId())
 		suite.Require().EqualValues(true, found)
 		// verify index
-		suite.Require().EqualValues(i+1, expectedLatestStateInfoIndex.Index)
+		suite.Require().EqualValues(i+1, latestStateInfoIndex.Index)
 		// load last state info
-
-		expectedStateInfo, found := suite.App.RollappKeeper.GetStateInfo(suite.Ctx, rollapp.GetRollappId(), expectedLatestStateInfoIndex.GetIndex())
+		expectedStateInfo, found := suite.App.RollappKeeper.GetStateInfo(suite.Ctx, rollapp.GetRollappId(), latestStateInfoIndex.GetIndex())
 		suite.Require().EqualValues(true, found)
 
 		// verify finalization queue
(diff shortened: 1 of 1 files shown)


x/rollapp/keeper/msg_server_update_state_test.go
```diff
@@ -147,55 +146,31 @@ func (suite *RollappTestSuite) TestUpdateState() {
 		suite.Require().Nil(err)
 
 		// end block
-		responseEndBlock := suite.App.EndBlocker(suite.Ctx, abci.RequestEndBlock{Height: suite.Ctx.BlockHeight()})
+		suite.App.EndBlocker(suite.Ctx, abci.RequestEndBlock{Height: suite.Ctx.BlockHeight()})
+
+		for i := uint64(0); i <= expectedLatestStateInfoIndex.Index; i++ {
```
omritoptix, line 151: calling it `expectedLatest..` an using it as a var all over the code is a bit confusing. Usually expected is used for comparison.

x/rollapp/keeper/msg_server_update_state_test.go
```diff
@@ -147,55 +146,31 @@ func (suite *RollappTestSuite) TestUpdateState() {
...
+			expectedStateInfo, found := suite.App.RollappKeeper.GetStateInfo(suite.Ctx, rollapp.GetRollappId(), i)
```
omritoptix, line 152: same here with the `expected` var name.
also because it's an index which is incremented by 1, is there a reason it shouldn't be found?
If not, I'd add `suite.Require().EqualTrue(found)`

x/rollapp/keeper/msg_server_update_state_test.go
```diff
@@ -147,55 +146,31 @@ func (suite *RollappTestSuite) TestUpdateState() {
...
+			if found {
+				if uint64(suite.Ctx.BlockHeight()) > disputePeriodInBlocks {
```
omritoptix, line 154: the blokHeight AFAIU shouldn't change between loop iteratins. why is not checked before the loop?

x/rollapp/keeper/msg_server_update_state_test.go
srene: changed

x/rollapp/keeper/msg_server_update_state_test.go
srene: found removed

x/rollapp/keeper/msg_server_update_state_test.go
srene: true. moved outside the loop

omritoptix (approved):

PR #668, opened 2024-03-14
feat(evm): integrate virtual frontier bank contract

VictorTrustyDev:
Closes: #627
Closes: #658 

For the detailed information about the VFC, visit: https://github.com/dymensionxyz/ethermint/pull/11.
This PR integrate the feature into Dymension Hub.

Features:
- Virtual Frontier Bank Contract
- Auto deploy contract
  - For Dymension chains, after new denom metadata created via gov (made by `x/denommetadata`)
    - IBC denom only
  - For new chains (eg: local-net), at genesis block, for the existing IBC denom in bank metadata

Incompatible issue:
- Dymension uses default proto of Cosmos, which is BaseAccount, not Ethermint's default is EthAccount so the expected VFBC contract code hash can not be set, leading to loss of mapping to the deployed bytecode.
  - The problem was solved by this [work around](https://github.com/dymensionxyz/ethermint/pull/11/commits/f5ecd0f9d6e44ba51299a07fc781f03ea5c2f2c8)
  - Currently still work without problem but unsure.
  - **TODO: In the future, if enable contract creation for DYM, need to migrate all the VFBC to use EthAccount and set the corresponding code hash and code.**

TODO:
- [x] Upgrade Ethermint version upon [PR](https://github.com/dymensionxyz/ethermint/pull/11) merged.
- [ ] ~~Generate swagger.~~ _(skipped because no command to do it atm)_
- [x] Remove [this method](https://github.com/dymensionxyz/dymension/blob/19895079ea9cefccb34284b9aa8f132018b27622/x/denommetadata/genesis.go#L14-L73), which was added for testing purpose.
- [x] Remove [this method](https://github.com/dymensionxyz/dymension/pull/668/commits/beb894d7a5749324b6e244ddfc57cc718b64950c), which was added for testing purpose.


app/app.go
```diff
@@ -685,7 +695,8 @@ func New(
 		AddRoute(ibcclienttypes.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper)).
 		AddRoute(streamermoduletypes.RouterKey, streamermodule.NewStreamerProposalHandler(app.StreamerKeeper)).
 		AddRoute(rollappmoduletypes.RouterKey, rollappmodule.NewRollappProposalHandler(&app.RollappKeeper)).
-		AddRoute(denommetadatamoduletypes.RouterKey, denommetadatamodule.NewDenomMetadataProposalHandler(&app.DenomMetadataKeeper))
+		AddRoute(denommetadatamoduletypes.RouterKey, denommetadatamodule.NewDenomMetadataProposalHandler(app.DenomMetadataKeeper)).
+		AddRoute(evmtypes.RouterKey, evm.NewEvmProposalHandler(app.EvmKeeper))
```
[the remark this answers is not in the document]
VictorTrustyDev: to active/deactive VFBC contracts

x/denommetadata/keeper/hooks.go
```diff
@@ -0,0 +1,38 @@
...
+
+// NewVirtualFrontierBankContractRegistrationHook returns the DenomMetadataHooks for VFBC registration
+func NewVirtualFrontierBankContractRegistrationHook(evmKeeper evmkeeper.Keeper) VirtualFrontierBankContractRegistrationHook {
+	return VirtualFrontierBankContractRegistrationHook{
+		evmKeeper: evmKeeper,
+	}
+}
+
+func (v VirtualFrontierBankContractRegistrationHook) AfterDenomMetadataCreation(ctx sdk.Context, newDenomMetadata banktypes.Metadata) error {
+	if strings.HasPrefix(newDenomMetadata.Base, "ibc/") { // only deploy for IBC denom.
```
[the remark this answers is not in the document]
VictorTrustyDev: I had never seen before. Should it be considered?

VictorTrustyDev:
@omritoptix should we remove [this function](https://github.com/dymensionxyz/dymension/blob/19895079ea9cefccb34284b9aa8f132018b27622/x/denommetadata/genesis.go#L14-L73)? Used to deploy a VFBC for the local-net. Or any better way to do? I think should have a VFBC for local-net is convenient for development purpose without extra actions to deploy one (like connect IBC, submit gov register denom metadata, vote)

omritoptix:
@VictorTrustyDev  how about just doing it with an external bash script like we're setting most things? (e.g pools, incentives etc)

x/denommetadata/keeper/hooks.go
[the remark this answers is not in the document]
omritoptix: not sure. how about we just do lowercase the `newDenomMetadata.Base` just in case?

VictorTrustyDev:
but contract deployment triggers via gov

omritoptix:
the gov voting is 60 seconds so you can create a script which can be ran after the chain starts (optionally) which sends a gov and votes it right away. look at `pool_bootstrap.sh` for example of bootstraping pools.

x/denommetadata/keeper/hooks.go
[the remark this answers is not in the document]
VictorTrustyDev: [here](https://github.com/dymensionxyz/dymension/pull/668/commits/7efdd2d7a899ad67bf3a82787b80aace30425d78)

VictorTrustyDev:
ok I removed the code

VictorTrustyDev:
> @VictorTrustyDev how about just doing it with an external bash script like we're setting most things? (e.g pools, incentives etc)

@omritoptix 
VFBC deployment can only be triggered via gov `create` denom-metadata, while localnet setup already have the denom-metadata setup so can not `gov` to create one more time.
Beside that, there is no other token on-chain, like IBC, so `gov` to register a ghost denom would not help testing because no account have balance of that ghost token.
So I put back the code to deploy at genesis [here](https://github.com/dymensionxyz/dymension/pull/668/commits/beb894d7a5749324b6e244ddfc57cc718b64950c).

x/denommetadata/keeper/hooks.go
mtsitrin, line 1: the implemenatation needs to be part of the `x/vfc` and not of `x/denommetada`

The `denommetada` module should be "free" of VFC reference

x/denommetadata/module.go
```diff
@@ -131,7 +140,7 @@ func (am AppModule) RegisterInvariants(ir sdk.InvariantRegistry) {
 // InitGenesis performs the module's genesis initialization.
 // Returns an empty ValidatorUpdate array.
 func (am AppModule) InitGenesis(ctx sdk.Context, cdc codec.JSONCodec, gs json.RawMessage) []abci.ValidatorUpdate {
-
+	_ = am.evmKeeper.DeployVirtualFrontierBankContractForBankDenomMetadataRecord(ctx, am.evmKeeper.GetParams(ctx).EvmDenom)
```
mtsitrin, line 143: why it's part of the `denommetadata` module?

x/denommetadata/keeper/hooks.go
VictorTrustyDev: Allow me repeat again, `vfc` is not a module, it is a part of the `x/evm` module.

x/denommetadata/module.go
VictorTrustyDev: I place it there so Tza can develop his part. Can remove when PR merge.

x/denommetadata/keeper/hooks.go
VictorTrustyDev: I already [moved](https://github.com/dymensionxyz/dymension/pull/668/commits/35a3e27831a4fe963ba1c9ecd74c84966cfbe0c7) the hook part to `x/vfc/hooks` folder.

x/denommetadata/module.go
omritoptix: @VictorTrustyDev can you please remove? @ItzhakBokris tested. seems good. 

possibly we'd wanna write init genesis for the vfc in case of starting from a genesis with multiple denom metadatas to deploy vfc for them. wdyt?

VictorTrustyDev: @omritoptix problem is no account have balance to transfer

VictorTrustyDev: FYI: removed

VictorTrustyDev: @omritoptix @mtsitrin I have just added
[`x/denommetadata` triggers hook at genesis for created bank denom metadata](https://github.com/dymensionxyz/dymension/pull/668/commits/8a7c0e42ed67ccd739262f2bd1892da0b7731aac)
Ofc the filter still applied (`ibc/*` only), wdyt?

omritoptix (approved):

PR #718, opened 2024-03-21
fix(rollapp): EIP155 rollapp keys can be overwritten

### LINKED ISSUE #717: EIP155 keys owned by other rollapps can be overwritten
the ParseChainID function trims whitespace characters from the chain ID before checking whether the chain ID is EVM-compatible. This is problematic because malicious rollapp creators can overwrite the EIP155 key storage into their own rollapps using the same chain ID with extra whitespaces.

srene:
## Description

the ParseChainID function trims whitespace characters from the chain ID before checking whether the chain ID is EVM-compatible. This is problematic because malicious rollapp creators can overwrite the EIP155 key storage into their own rollapps using the same chain ID with extra whitespaces. Using chain ID with extra space will pass the existing rollapp  checks, but it will lead to the same EIP155 key after parsing the chain id.

This PR is aimed at solving this by checking there is no existing rollapp with the same EIP155 key instead of checking by rollapp name.

This PR also refactors how the chain ID is parsed and checked, creating a common struct for that, with the chain id and the eip155 id in case is an EVM compatible id.

Closes #717 

----

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:

### DIFF
--- a/x/rollapp/keeper/msg_server_create_rollapp.go
+++ b/x/rollapp/keeper/msg_server_create_rollapp.go
@@ -18,14 +18,8 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 		return nil, err
 	}
 
-	// error already checked in ValidateBasic
-	rollappID, eip155, _ := types.GetValidEIP155ChainId(msg.RollappId)
-
-	if eip155 == nil {
-		return nil, types.ErrInvalidRollappID
-	}
 	// check to see if the RollappId has been registered before
-	if _, isFound := k.GetRollappByEIP155(ctx, eip155.Uint64()); isFound {
+	if _, isFound := k.GetRollapp(ctx, msg.RollappId); isFound {
 		return nil, types.ErrRollappExists
 	}
 
--- a/x/rollapp/keeper/msg_server_create_rollapp.go
+++ b/x/rollapp/keeper/msg_server_create_rollapp.go
@@ -14,10 +14,6 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 		return nil, types.ErrRollappsDisabled
 	}
 
-	if err := msg.ValidateBasic(); err != nil {
-		return nil, err
-	}
-
 	rollappId, err := types.NewChainID(msg.RollappId)
 	if err != nil {
 		return nil, err
--- a/x/rollapp/keeper/msg_server_create_rollapp.go
+++ b/x/rollapp/keeper/msg_server_create_rollapp.go
@@ -18,7 +18,7 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 		return nil, err
 	}
 
-	//error already checked in ValidateBasic
+	// error already checked in ValidateBasic
 	eip155, _ := types.ParseChainID(msg.RollappId)
 
 	if eip155 == nil {
--- a/x/rollapp/keeper/msg_server_create_rollapp.go
+++ b/x/rollapp/keeper/msg_server_create_rollapp.go
@@ -18,8 +18,13 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 		return nil, err
 	}
 
+	rollappId, err := types.NewChainID(msg.RollappId)
+	if err != nil {
+		return nil, err
+	}
+
 	// check to see if the RollappId has been registered before
-	if _, isFound := k.GetRollapp(ctx, msg.RollappId); isFound {
+	if _, isFound := k.GetRollapp(ctx, rollappId.ChainID); isFound {
 		return nil, types.ErrRollappExists
 	}
 
--- a/x/rollapp/keeper/msg_server_create_rollapp.go
+++ b/x/rollapp/keeper/msg_server_create_rollapp.go
@@ -18,10 +18,8 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 		return nil, err
 	}
 
-	eip155, err := types.ParseChainID(msg.RollappId)
-	if err != nil {
-		return nil, err
-	}
+	//error already checked in ValidateBasic
+	eip155, _ := types.ParseChainID(msg.RollappId)
 
 	if eip155 == nil {
 		return nil, types.ErrInvalidRollappID
--- a/x/rollapp/keeper/msg_server_create_rollapp.go
+++ b/x/rollapp/keeper/msg_server_create_rollapp.go
@@ -24,13 +24,13 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 	}
 
 	// check to see if the RollappId has been registered before
-	if _, isFound := k.GetRollapp(ctx, rollappId.ChainID); isFound {
+	if _, isFound := k.GetRollapp(ctx, rollappId.GetChainID()); isFound {
 		return nil, types.ErrRollappExists
 	}
 
 	if rollappId.IsEIP155() {
 		// check to see if the RollappId has been registered before with same key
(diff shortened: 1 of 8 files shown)


danwt (commented):
I think you're code is correct but the whole thing with the chain id isn't really dealt with cleanly and will probably become quite annoying in the future.

I think we need a function

getValidChainID(string) (string,error) 

which can use

func ParseChainID(chainID string) (*big.Int, error)
https://github.com/dymensionxyz/dymension/blob/4ee6b752dae868b760cb85e2f2533a261038a3ef/x/rollapp/types/chain_id.go#L35-L60
which should be renamed to ParseEVMChainID

and we need to delete the dead code here

https://github.com/dymensionxyz/dymension/blob/4ee6b752dae868b760cb85e2f2533a261038a3ef/x/rollapp/types/chain_id.go#L26-L33

Then we need to use the validated string here
https://github.com/dymensionxyz/dymension/blob/0cebdd327c68620cb09725370130ea582a12dacd/x/rollapp/keeper/msg_server_create_rollapp.go#L52
 instead of using msg.RollappId

srene:
code refactored with a new struct that parses and validates  chain id, and creates an eip155 key in case it is evm compatible. this new code blocks the ability to overwrite the eip155 keys with rollappids with extra whitespace, as stated in the audit.

mtsitrin (changes requested):
This PR does'nt fix the issue. the whitespace is not the issue here.

```
	rollappId := "rollapp_1234-1" 
	badrollappId := "dummy_1234-1" 
```

will overwrite the `1234` eip key with the 2nd rollapp

srene:
> This PR does'nt fix the issue. the whitespace is not the issue here.
> 
> ```
> 	rollappId := "rollapp_1234-1" 
> 	badrollappId := "dummy_1234-1" 
> ```
> 
> will overwrite the `1234` eip key with the 2nd rollapp

True. it appears there was more than one way to overwrite the key, and not only the one described in the audit. should be fixed now. thanks

x/rollapp/types/chain_id.go
```diff
@@ -23,28 +23,35 @@ var (
...
+	chainID := strings.TrimSpace(id)
 
 	if chainID == "" {
-		return nil, errorsmod.Wrapf(ErrInvalidRollappID, "chain-id cannot be empty")
+		return ChainID{}, errorsmod.Wrapf(ErrInvalidRollappID, "chain-id cannot be empty")
 	}
 
 	if len(chainID) > 48 {
-		return nil, errorsmod.Wrapf(ErrInvalidRollappID, "rollapp-id '%s' cannot exceed 48 chars", chainID)
+		return ChainID{}, errorsmod.Wrapf(ErrInvalidRollappID, "rollapp-id '%s' cannot exceed 48 chars", chainID)
```
danwt, line 39:
```suggestion
				return ChainID{}, errorsmod.Wrapf(ErrInvalidRollappID, "exceeds 48 chars: %s: len: %d", chainID, len(chainID))
```

x/rollapp/types/chain_id.go
```diff
@@ -23,28 +23,35 @@ var (
...
-// ParseChainID parses a string chain identifier's epoch to an Ethereum-compatible
-// chain-id in *big.Int format. The function returns an error if the chain-id has an invalid format
-func ParseChainID(chainID string) (*big.Int, error) {
-	chainID = strings.TrimSpace(chainID)
+func NewChainID(id string) (ChainID, error) {
+	chainID := strings.TrimSpace(id)
 
 	if chainID == "" {
-		return nil, errorsmod.Wrapf(ErrInvalidRollappID, "chain-id cannot be empty")
+		return ChainID{}, errorsmod.Wrapf(ErrInvalidRollappID, "chain-id cannot be empty")
```
danwt, line 35:
```suggestion
		return ChainID{}, errorsmod.Wrapf(ErrInvalidRollappID, "empty")
```

x/rollapp/types/chain_id.go
```diff
@@ -23,28 +23,35 @@ var (
...
-func IsValidChainID(chainID string) bool {
-	if len(chainID) > 48 {
-		return false
-	}
-
-	return ethermintChainID.MatchString(chainID)
+type ChainID struct {
+	ChainID  string
+	EIP155ID *big.Int
 }
```
danwt, line 29: can we add a method to get the eip155id as uint64, rather than doing the conversion manually in loads of places

x/rollapp/keeper/msg_server_create_rollapp_test.go
```diff
@@ -117,3 +150,77 @@ func (suite *RollappTestSuite) TestCreateRollappWhenDisabled() {
...
+	badrollapp := types.MsgCreateRollapp{
+		Creator:               alice,
+		RollappId:             badrollappId,
+		MaxSequencers:         1,
+		PermissionedAddresses: []string{},
+	}
+	_, err = suite.msgServer.CreateRollapp(goCtx, &badrollapp)
+	// it should not be possible to register rollapp name with same EIP155 Key
+	suite.Require().ErrorIs(err, types.ErrRollappExists)
+}
```
danwt, line 226: this is a huge duplication, can we have a table test

or use t.Run()

x/rollapp/keeper/msg_server_create_rollapp_test.go
```diff
@@ -107,6 +108,38 @@ func (suite *RollappTestSuite) TestCreateRollappAlreadyExists() {
...
+	// rollapp is the rollapp to create
+	rollapp := types.MsgCreateRollapp{
+		Creator:               alice,
+		RollappId:             rollappName,
+		MaxSequencers:         1,
+		PermissionedAddresses: []string{},
+	}
+	_, err := suite.msgServer.CreateRollapp(goCtx, &rollapp)
+	suite.Require().ErrorIs(err, types.ErrInvalidRollappID)
+}
```
danwt, line 141: table test or use t.Run() subtests

danwt (changes requested):
The struct is a good idea

Can we additionally document the GetRollapp function to better explain what exactly is a rollappId, and why there are two different keys?

https://github.com/dymensionxyz/dymension/blob/a8780eb3915b7a28c75320f8f5a7e0c516ab5cd9/x/rollapp/keeper/rollapp.go#L47-L63

x/rollapp/types/chain_id.go
srene: done

x/rollapp/keeper/msg_server_create_rollapp_test.go
srene: done

x/rollapp/keeper/msg_server_create_rollapp_test.go
srene: done

srene:
> The struct is a good idea
> 
> Can we additionally document the GetRollapp function to better explain what exactly is a rollappId, and why there are two different keys?
> 
> https://github.com/dymensionxyz/dymension/blob/a8780eb3915b7a28c75320f8f5a7e0c516ab5cd9/x/rollapp/keeper/rollapp.go#L47-L63

comments were added in the function

x/rollapp/keeper/msg_server_create_rollapp.go
```diff
@@ -14,11 +14,27 @@ func (k msgServer) CreateRollapp(goCtx context.Context, msg *types.MsgCreateRoll
 		return nil, types.ErrRollappsDisabled
 	}
 
+	if err := msg.ValidateBasic(); err != nil {
```
mtsitrin, line 17: this redundent as the `msg.ValidateBasic()` already called by the SDK before getting into this function

srene: fixed

mtsitrin (approved):

Issue #739, opened 2024-03-27
finalization errors could lead to corrupted state


omritoptix:
As it runs in being block, it doesn't revert on error. 
We need to wrap this part in module.go in applyFuncIfNoError:

```
func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
	err := am.keeper.FinalizeQueue(ctx)
```

danwt:
just fyi you can probably do 'open with github' in the IDE (command palette) to get a link easily

PR #745, opened 2024-03-27
fix(rollapp): use ApplyFuncIfNoError in finalize end blocker

danwt:
## Description

In this PR, we protect against errors in endblock
Additionally, we reduce some defensive programming

----

Closes https://github.com/dymensionxyz/dymension/issues/739

PR review checkboxes:

I have...

- [x]  Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


x/rollapp/keeper/state_info.go
```diff
@@ -34,6 +36,17 @@ func (k Keeper) GetStateInfo(
 	return val, true
 }
 
+func (k Keeper) MustGetStateInfo(ctx sdk.Context,
```
omritoptix, line 39: why not changing `GetStateInfo` all together to panic? because of the cascading changes required all over?

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -7,39 +7,47 @@ import (
...
 			k.SetLatestFinalizedStateIndex(ctx, stateInfoIndex)
 			// call the after-update-state hook
 			keeperHooks := k.GetHooks()
-			err = keeperHooks.AfterStateFinalized(ctx, stateInfoIndex.RollappId, &stateInfo)
+			err := keeperHooks.AfterStateFinalized(ctx, stateInfoIndex.RollappId, &stateInfo)
 			if err != nil {
-				// Failed to call finalization dependent event like ibc packet finalization, invariant breaking. can't proceed
-				return fmt.Errorf("error calling finalized state finalized: rollappID %s, stateInfo: %+v, error %s",
-					stateInfoIndex.RollappId, stateInfo, err.Error())
+				panic(fmt.Errorf("invariant broken: after state finalized must succeed: %w", err))
```
omritoptix, line 50: why is this a broken invariant? if the after state finalized doens't succeed, it's not necessarily means an invariant was broken as I see it. in that case it means we failed to finalize e.g rollapp packets so we return an error which will cause revert the changes so far.

x/rollapp/keeper/state_info.go
mtsitrin: I think it make sense as u have queries for stateInfo, which u probably don't want to panic for non existing index

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -7,39 +7,47 @@ import (
...
+		// we trap at this granularity because we want to avoid iterating inside the
+		// wrapped function (above), and we want the function to be atomic, so
+		// we cannot trap inside finalizePending
+		func(ctx sdk.Context) error {
+			k.finalizePending(ctx, pendingFinalizationQueue)
+			return nil
+		})
+}
+
+func (k Keeper) finalizePending(ctx sdk.Context, pendingFinalizationQueue []types.BlockHeightToFinalizationQueue) {
```
mtsitrin, line 32: why u panic in this func instead of returning err?
it's much cleaner. u use cacheCtx on error also, don't have to panic

mtsitrin (changes requested):

x/rollapp/keeper/state_info.go
danwt: what @mtsitrin said

x/rollapp/keeper/block_height_to_finalization_queue.go
danwt: Fair point. I was putting too much trust in the pre existing comment saying 'invariant breaking'
I will change this one to return an error

x/rollapp/keeper/block_height_to_finalization_queue.go
danwt: I have changed the 
`keeperHooks.AfterStateFinalized(ctx, stateInfoIndex.RollappId, &stateInfo)`
to be an error

But I'll keep the panic for the other one

I think it's a good idea to be disciplined about broken invariants, when they break, we should panic, because it means we have a logic bug and can't recover

Does it make sense?

omritoptix (approved):

mtsitrin (approved):

Issue #750, opened 2024-03-28
sync validations between different `token metadata` componentts 


mtsitrin:
currently we have a few places that use the token metadata:
1. `x/denommetadata`
2. `vfc`
3. `x/rollapp` on rollapp registration

the validation scheme should be the same, as they all need to pass for successful registration.
--------
e.g this case, passes the `rollapp creation` but fails on `vfc` registration

```
[
    {
        "description": "rollapp rollappevm_1234-1 native token",
        "base": "urax",
        "denom_units": [
            {
                "denom": "urax",
                "exponent": 0
            }
        ],
        "name": "rollappevm_1234-1 urax",
        "symbol": "URAX",
        "display": "urax"
    }
]
```

PR #752, opened 2024-03-28
Fix vfc hook metadata.Validate() AfterDenomMetadataCreation

trinitys7:
## Description

----
When we trigger genesis event with only 1 denom_unit, the denommeta data will something like :
```
{
        "base": "ibc/FECACB927EB3102CCCB240FFB3B6FCCEEB8D944C6FEA8DFF079650FEFF59781D",
        "denom_units": [
            {
                "denom": "ibc/FECACB927EB3102CCCB240FFB3B6FCCEEB8D944C6FEA8DFF079650FEFF59781D",
                "exponent": 0
            }
        ],
        "symbol": "URX",
        "display": "urax"
    }
```

The denommetadata will no longer validated because `denom_units` does not contain  `display`. 
This PR will fix that problem by saving `ibc denom` to `display`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/cosmos/cosmos-sdk v0.46.16
 	github.com/cosmos/ibc-apps/middleware/packet-forward-middleware/v6 v6.1.2-0.20240228222509-2b38263c41d3
 	github.com/cosmos/ibc-go/v6 v6.2.1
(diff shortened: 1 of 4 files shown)


mtsitrin:
As discussed, it's not the way to solve it.
will be treated in https://github.com/dymensionxyz/dymension/issues/750

in the meanwhile, the rollapp token registered on creation should include denom_unit for both the base denom and the display denom

PR #766, opened 2024-03-29
fix(delayedack): rollapp packet finalization not filtered by rollapp 

### LINKED ISSUE #757: `FinalizeRollappPackets` does not filter by rollappId, causing packets to be incorrectly processed

zale144:
## Description

----

Closes #757 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


x/delayedack/rollapp_hooks.go
```diff
@@ -32,79 +36,147 @@ func (im IBCMiddleware) FinalizeRollappPackets(ctx sdk.Context, rollappID string
...
-			logger.Debug("Calling OnTimeoutPacket", "rollappID", rollappID, "sequence", rollappPacket.Packet.GetSequence(), "destination channel", rollappPacket.Packet.GetDestChannel())
-			wrappedFunc := func(ctx sdk.Context) error {
-				err := im.IBCModule.OnTimeoutPacket(ctx, *rollappPacket.Packet, rollappPacket.Relayer)
-				if err != nil {
-					return err
-				}
-				return nil
-			}
-			err := osmoutils.ApplyFuncIfNoError(ctx, wrappedFunc)
+		if rollappPacket.RollappId != rollappID {
```
omritoptix, line 49: I think much more efficeint method would be to only iterate on the relevant rollapp packets from the first place vs iterating everything and filter after.
This can be done by expanding the `ListRollappPacketByStatus` to get an optional rollapp_id arguement or in another way. The rollapp key is built in such way that it should allow it as it's built using `/status/rollapp-id`

zale144: @omritoptix 
It's actually built like this, so `status/proof_height/rollapp_id/...`

```
result := append(statusPrefix, keySeparatorBytes...)
result = append(result, proofHeightBytes...)
result = append(result, keySeparatorBytes...)
result = append(result, rollappIdBytes...)
result = append(result, keySeparatorBytes...)
result = append(result, packetUIDBytes...)
```
so we get a key like this: `“\0\x02/\0\0\0\0\0\0k\b/rolintdev_5832-2/channel-1\0\0\0\0\0\0\0\x14”`

I think we would need to also put the proof height in the prefix, or just rearrange the parts of the key.

x/delayedack/keeper/rollapp_packet.go
```diff
@@ -120,61 +124,33 @@ func (k *Keeper) UpdateRollappPacketWithStatus(ctx sdk.Context, rollappPacket co
...
 
 	for ; iterator.Valid(); iterator.Next() {
 		var val commontypes.RollappPacket
 		k.cdc.MustUnmarshal(iterator.Value(), &val)
-		if maxProofHeight == 0 || val.ProofHeight <= maxProofHeight {
+
+		if listFilter.filter == nil || listFilter.filter(val) {
 			list = append(list, val)
-		} else {
+		} else if listFilter.stopOnFirstMatch {
```
omritoptix, line 140: imo `stopOnFirstMatch` is a confusing name as it doesn’t actually stop on first match but on first "Unmatch" in that sense.

x/delayedack/keeper/hooks.go
```diff
@@ -52,21 +52,18 @@ func (k Keeper) GetEpochHooks() epochstypes.EpochHooks {
...
+func (e epochHooks) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, _ int64) error {
 	if epochIdentifier != e.GetParams(ctx).EpochIdentifier {
 		return nil
 	}
 	// Get all rollapp packets with status != PENDING and delete them
-	finalizedRollappPackets := e.ListRollappPacketsByStatus(ctx, commontypes.Status_FINALIZED, 0)
-	revertedRollappPackets := e.ListRollappPacketsByStatus(ctx, commontypes.Status_REVERTED, 0)
-	toDeletePackats := append(finalizedRollappPackets, revertedRollappPackets...)
-	for _, packet := range toDeletePackats {
+	toDeletePackets := e.ListRollappPackets(ctx, ByNotStatus(commontypes.Status_PENDING))
```
omritoptix, line 64: I’m afraid it will be very inefficient as we need to iterate over all packets every interval where the majority will be pending packets assuming epoch > grace period. 
Original epoch planned was 1hr.

x/delayedack/keeper/invariants.go
```diff
@@ -6,108 +6,102 @@ import (
...
+	if checkFrozen && !rollapp.Frozen {
+		return
+	}
+
+	// will stay 0 if no state is found
+	// but will still check packets
+	var latestFinalizedHeight uint64
 
-		rollapps := k.rollappKeeper.GetAllRollapps(ctx)
+	defer func() {
```
omritoptix, line 62: why deferring vs just putting at the end of the func?

x/delayedack/keeper/invariants.go
```diff
@@ -6,108 +6,102 @@ import (
...
+	if !found {
+		return
+	}
 
-		for _, rollapp := range rollapps {
-			if !rollapp.Frozen {
-				continue
-			}
-			frozenRollappsFinalizedHeight[rollapp.RollappId] = 0
+	latestFinalizedStateInfo, found := k.rollappKeeper.GetStateInfo(ctx, rollapp.RollappId, latestFinalizedStateIndex.Index)
```
omritoptix, line 72: Id say this is invariant breaking (i.e index without matching stateInfo) so maybe worth wrapping in `Must`.

zale144: agreed, I tried my best to stick with the original logic

x/delayedack/keeper/invariants.go
zale144: in the original implementation, the packets are always checked, event if no state was found. This is just a refactor, I wanted the logic to stay the same, assuming it is implemented as intended

x/delayedack/keeper/hooks.go
zale144: We have one packet key which starts with the rollappID now. My idea of narrowing down the number of packets for most use cases is if we first filter by rollappID prefix and then by status. If we get a subset of packets by RollApp, I wasn't sure, but I thought it might be smaller than a subset by status - FINALIZED + REVERTED.
As it was, we fetch all FINALIZED and REVERTED rollapps and then loop over them. 

I'll think of a better way

x/delayedack/keeper/rollapp_packet.go
zale144: my bad

x/delayedack/keeper/rollapp_packet.go
```diff
@@ -120,61 +124,46 @@ func (k *Keeper) UpdateRollappPacketWithStatus(ctx sdk.Context, rollappPacket co
...
-	store := ctx.KVStore(k.storeKey)
-
-	// Iterate over the range from lastProofHeight to proofHeight
-	iterator := sdk.KVStorePrefixIterator(store, commontypes.AllRollappPacketKeyPrefix)
+func (k Keeper) iterateOverRollappPacketsPerStatus(
+	iterator sdk.Iterator,
+	filter filterFunc,
+	breakOnMismatch bool,
+) (list []commontypes.RollappPacket) {
 	defer iterator.Close() // nolint: errcheck
```
mtsitrin, line 145: iterator better be closed by caller and not callee

x/common/types/key_rollapp_packet.go
```diff
@@ -27,30 +25,49 @@ var (
...
 // RollappPacketKey constructs a key for a specific RollappPacket
 func RollappPacketKey(rollappPacket *RollappPacket) []byte {
+	rollappIdBytes := []byte(rollappPacket.RollappId)
 	// Get the relevant key prefix based on the packet status
-	statusPrefix := MustGetStatusBytes(rollappPacket.Status)
+	statusBytes := mustGetStatusBytes(rollappPacket.Status)
 	// Build the key bytes repr. Convert each uint64 to big endian bytes to ensure lexicographic ordering.
-	keySeparatorBytes := []byte(KeySeparator)
-	rollappIdBytes := []byte(rollappPacket.RollappId)
+	keySeparatorBytes := []byte(keySeparator)
```
mtsitrin, line 32: can be global const as well

x/delayedack/keeper/rollapp_packets_list_filter.go
```diff
@@ -0,0 +1,58 @@
+package keeper
+
+import commontypes "github.com/dymensionxyz/dymension/v3/x/common/types"
+
+type filterFunc func(val commontypes.RollappPacket) bool
+
+type rollappPacketListFilter struct {
+	prefixes        [][]byte
```
mtsitrin, line 8: why u need to support multiple prefixes?
IMO it's more clearer to leave `rollappPacketListFilter` for single query

x/delayedack/keeper/rollapp_packets_list_filter.go
```diff
@@ -0,0 +1,58 @@
...
+}
+func ByRollappIDAndStatusAndMaxHeight(
+	rollappID string,
+	maxProofHeight uint64,
+	breakOnMismatch bool,
+	status ...commontypes.Status,
+) rollappPacketListFilter {
+	filter := ByRollappIDAndStatus(rollappID, status...)
+	filter.breakOnMismatch = breakOnMismatch
+	filter.filter = func(val commontypes.RollappPacket) bool {
```
mtsitrin, line 30: why u need filter for it?
can't u use `status/rollappID/proofHeight` for proof height iterator?
otherwise u get all the heights and filtering afterwards

x/delayedack/genesis.go
```diff
@@ -20,6 +20,6 @@ func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState)
 // ExportGenesis returns the module's exported genesis
 func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	return &types.GenesisState{
-		RollappPackets: k.GetAllRollappPackets(ctx),
+		RollappPackets: k.ListRollappPackets(ctx, keeper.AllRollappPackets()),
```
mtsitrin, line 23: the `keeper.AllRollappPackets()` feels redundent. same as passing nothing.

x/common/types/key_rollapp_packet.go
```diff
@@ -27,30 +25,49 @@ var (
...
+// by status and rollappID:
+// "status/rollappID/"
+// by only status:
+// "status/"
+func RollappPacketStatusAndRollappIDKey(rollappPacket *RollappPacket) (result []byte) {
+	rollappIdBytes := []byte(rollappPacket.RollappId)
+	statusPrefix := mustGetStatusBytes(rollappPacket.Status)
+	keySeparatorBytes := []byte(keySeparator)
+	result = append(statusPrefix, keySeparatorBytes...)
+	if rollappPacket.RollappId == "" {
```
mtsitrin, line 61: I would separate to `RollappPacketByStatusKey`
and call it `IteratorKey`, as it's key expected to use as iterator

omritoptix (dismissed):

x/delayedack/keeper/rollapp_packets_list_filter.go
omritoptix: i believe you can't as `status/rollappID/proofHeight` will give you all packets at height and you need all packets until height.

x/delayedack/keeper/rollapp_packets_list_filter.go
omritoptix: for filtering for both `REVERTED` and `FINALIZED` with one filter

x/delayedack/keeper/rollapp_packets_list_filter.go
mtsitrin: if `proofHeight` is ordered, should be possible to iterate `until height`

omritoptix: > otherwise u get all the heights and filtering afterwards

I believe the iterator gets a batch from the db every time (and not all elements),  and it does so only when  `next()` element is the last in that batch. 

in that sense, you'll make the same amount of db calls assuming it gets the height by order, which I believe it does (as you're breaking from the iterator when you reached max height). 

If that's the case, than the only mis-optimization I see here is holding more objects in memory than necessary (i.e the last batch could potentially contain less objects if the height was limited)

Not sure it's worth the effort of changing,  optimization wise.

hoangdv2429 (approved):

lacsomot (approved):

omritoptix (dismissed):

x/delayedack/keeper/rollapp_packets_list_filter.go
mtsitrin, line 1: the filters in this file require UT

x/common/types/key_rollapp_packet.go
```diff
@@ -40,15 +37,46 @@ func RollappPacketKey(rollappPacket *RollappPacket) []byte {
...
 
+func RollappPacketByStatusByRollappIDMaxProofHeightPrefixes(rollappID string, status Status, proofHeight uint64) ([]byte, []byte) {
+	return RollappPacketByStatusByRollappIDByMaxProofHeightPrefix(rollappID, status, 0),
+		RollappPacketByStatusByRollappIDByMaxProofHeightPrefix(rollappID, status, proofHeight+1) // inclusive end
+}
+
+// RollappPacketByStatusByRollappIDByMaxProofHeightPrefix constructs a key prefix for a specific RollappPacket
+// by rollappID, status and proofHeight:
+// "rollappID/status/proofHeight"
+func RollappPacketByStatusByRollappIDByMaxProofHeightPrefix(rollappID string, status Status, proofHeight uint64) []byte {
```
mtsitrin, line 57:
```suggestion
func RollappPacketByStatusByRollappIDByProofHeightPrefix(rollappID string, status Status, proofHeight uint64) []byte {
```

x/common/types/key_rollapp_packet.go
```diff
@@ -40,15 +37,46 @@ func RollappPacketKey(rollappPacket *RollappPacket) []byte {
...
 	result = append(result, rollappIdBytes...)
 	result = append(result, keySeparatorBytes...)
+	result = append(result, proofHeightBytes...)
+	result = append(result, keySeparatorBytes...)
 	result = append(result, packetUIDBytes...)
 
 	return result
 }
 
+func RollappPacketByStatusByRollappIDMaxProofHeightPrefixes(rollappID string, status Status, proofHeight uint64) ([]byte, []byte) {
```
mtsitrin, line 49: IMO this func is confusing here
it's logic can be done in the caller filter (`PendingByRollappIDByMaxHeight`)

x/common/types/key_rollapp_packet.go
```diff
@@ -23,14 +18,16 @@ var (
 	FinalizedRollappPacketKeyPrefix = []byte{0x00, 0x02}
 	// RevertedRollappPacketKeyPrefix is the prefix for reverted rollapp packets
 	RevertedRollappPacketKeyPrefix = []byte{0x00, 0x03}
+	// keySeparatorBytes is used to separate the rollapp packet key parts
+	keySeparatorBytes = []byte("/")
 )
 
 // RollappPacketKey constructs a key for a specific RollappPacket
+// status/rollappID/proofHeight/packetUID
 func RollappPacketKey(rollappPacket *RollappPacket) []byte {
```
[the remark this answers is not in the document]
zale144: To build the actual key? Good thinking

mtsitrin:
great job @zale144

x/common/types/key_rollapp_packet.go
```diff
@@ -25,36 +25,18 @@ var (
 // RollappPacketKey constructs a key for a specific RollappPacket
 // status/rollappID/proofHeight/packetUID
 func RollappPacketKey(rollappPacket *RollappPacket) []byte {
-	// Get the relevant key prefix based on the packet status
-	statusPrefix := MustGetStatusBytes(rollappPacket.Status)
-	// Build the key bytes repr. Convert each uint64 to big endian bytes to ensure lexicographic ordering.
-	rollappIdBytes := []byte(rollappPacket.RollappId)
-	proofHeightBytes := sdk.Uint64ToBigEndian(rollappPacket.ProofHeight)
-	// Build the packetUID from the destination channel and sequence number.
+	srppPrefix := RollappPacketByStatusByRollappIDByProofHeightPrefix(rollappPacket.RollappId, rollappPacket.Status, rollappPacket.ProofHeight)
```
omritoptix, line 28: srpp?

mtsitrin (approved):

omritoptix (commented):
I think it's better to create black box testing for the filter so I'd check it actually filters as expected vs create the bytes as expected. 

also I'd put it in a `keeper_test` package vs `keeper` package to enfoce black box tests.

PR #772, opened 2024-03-31
fix: removed rollapp related logic from init genesis and missing params fix

omritoptix:
## Description

This PR:
1. removes all import of rollapp genesis state as most likely rollapps will need to hard-fork to run on the new chain so importing previous rollapp state doesn't make sense. still in research
2. fixes a missing param initialization both on delayed ack and with eIBC params
3. Removes the CLI tests for rollapp and sequencer which were dependent on initGenesis. (https://github.com/dymensionxyz/dymension/issues/773)
----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


go.sum
```diff
@@ -483,7 +483,7 @@ github.com/dymensionxyz/ethermint v0.22.0-dymension-v0.4.1 h1:J31ASxmDVm7ytjmk9T
 github.com/dymensionxyz/ethermint v0.22.0-dymension-v0.4.1/go.mod h1:O2J61ZwM0Vdms6pRa1fb43pwmCuNRctro3AB90WlOc0=
 github.com/dymensionxyz/osmosis/osmomath v0.0.6-dym-v0.0.1 h1:59ZE3Ocrn04MUpb5VJgfi24eDcnQ9VBfCSw0Mx1n7OI=
 github.com/dymensionxyz/osmosis/osmomath v0.0.6-dym-v0.0.1/go.mod h1:2idySYJxP5YfMAEeSeqD8e7fSchfsI4jn7XFHJgNUsM=
-github.com/dymensionxyz/osmosis/v15 v15.2.1-0.20240310110736-a0811b5943a5 h1:fdu2lHVWNsACjqA5A63Y8sSVV8UiIoDKjOScUDTivOc=
+github.com/dymensionxyz/osmosis/v15 v15.2.1-0.20240310110736-a0811b5943a5 h1:1SSfSpblYEtcWOKttIDDpuaUOtThFHdWzU/9a42rITc=
```
omritoptix, line 486:
```suggestion
github.com/dymensionxyz/osmosis/v15 v15.2.1-0.20240310110736-a0811b5943a5 h1:fdu2lHVWNsACjqA5A63Y8sSVV8UiIoDKjOScUDTivOc=
```

hoangdv2429 (dismissed):
I think we should open another issue to keep track of what to do with additional infos from exportGenesis.

x/delayedack/genesis.go
```diff
@@ -9,17 +9,12 @@ import (
...
-			panic(err)
-		}
-	}
 }
 
 // ExportGenesis returns the module's exported genesis
 func ExportGenesis(ctx sdk.Context, k keeper.Keeper) *types.GenesisState {
 	return &types.GenesisState{
+		Params:         k.GetParams(ctx),
 		RollappPackets: k.GetAllRollappPackets(ctx),
```
srene, line 18: shouldn't we export the same data used in the import, and remove unused variables from genesistate?

x/delayedack/types/genesis.go
```diff
@@ -1,17 +1,12 @@
...
-
 // DefaultIndex is the default global index
 const DefaultIndex uint64 = 1
 
 // DefaultGenesis returns the default genesis state
 func DefaultGenesis() *GenesisState {
 	return &GenesisState{
-		Params:         DefaultParams(),
-		RollappPackets: []commontypes.RollappPacket{},
+		Params: DefaultParams(),
```
srene, line 9: shouldn't we remove rollapppackets from genesistate proto?

x/eibc/genesis.go
```diff
@@ -9,16 +9,6 @@ import (
...
-		demandOrderCopy := demandOrder
-		err := k.SetDemandOrder(ctx, &demandOrderCopy)
-		if err != nil {
-			panic(err)
-		}
-
-	}
 }
 
 // ExportGenesis returns the module's exported genesis
```
srene, line 14: same here, not sure why the export genesis is not updated in the same way.

srene (commented):
hi, im not sure why the export genesis of the modules is not updated in the same way of the init genesis. shouldn't the exported info match what is supposed to be imported?

hoangdv2429:
> hi, im not sure why the export genesis of the modules is not updated in the same way of the init genesis. shouldn't the exported info match what is supposed to be imported?

I think we're still discussing on how we should import/handle those data. That why we still export it @srene. Is that right @omritoptix

x/delayedack/genesis.go
mtsitrin: We didn't want to break the `Export` functionality, as it might be useful for other things,
regardless of the ability to init new chain from it.

x/delayedack/types/genesis.go
mtsitrin: prefer not to change proto files, as it requires store migration on upgrade.
we also **will support** hub forking in the future, once we'll research the required flow from RollApp prespective

mtsitrin:
need to add validation in the `genesisState` structs.
otherwise, one can provide gensis file with garbage, and it will be just ignored and run successfully.
IMO genesis file should be 100% correct.
if no rollapps/sequencers/packets expected, it should be validated

omritoptix:
> I think we should open another issue to keep track of what to do with additional infos from exportGenesis.
@hoangdv2429 yea we have research issue about it.

x/sequencer/genesis.go
```diff
@@ -10,6 +10,15 @@ import (
 // state.
 func InitGenesis(ctx sdk.Context, k keeper.Keeper, genState types.GenesisState) {
 	k.SetParams(ctx, genState.Params)
+	// Validate all other genesis fields are empty
+	genesisFields := []int{
```
hoangdv2429, line 14: can we just do 
`if len(genState.SequencerList) != 0 {
        return fmt.Errorf("invalid genesis state: SequencerList should be empty")
   }`

omritoptix: it's in the init genesis, which doesn't except error to be retuned, but panicing on failure.

hoangdv2429: Then we can panic instead of returning error maybe

omritoptix: ha you mean that genesisFields is not needed because it's one field. gotcha.

hoangdv2429 (approved):

x/rollapp/genesis.go
```diff
@@ -9,28 +9,20 @@ import (
...
-	for _, elem := range genState.LatestFinalizedStateIndexList {
-		k.SetLatestFinalizedStateIndex(ctx, elem)
+	k.SetParams(ctx, genState.Params)
+	// Validate all other genesis fields are empty
+	genesisFields := []int{
+		len(genState.RollappList),
+		len(genState.StateInfoList),
+		len(genState.LatestStateInfoIndexList),
+		len(genState.LatestFinalizedStateIndexList),
+		len(genState.BlockHeightToFinalizationQueueList),
```
zale144, line 19:
```suggestion
	genesisFieldsLen := len(genState.RollappList) +
		len(genState.StateInfoList) +
		len(genState.LatestStateInfoIndexList) +
		len(genState.LatestFinalizedStateIndexList) +
		len(genState.BlockHeightToFinalizationQueueList)
```

x/rollapp/genesis.go
```diff
@@ -9,28 +9,20 @@ import (
...
 	}
-	// Set all the blockHeightToFinalizationQueue
-	for _, elem := range genState.BlockHeightToFinalizationQueueList {
-		k.SetBlockHeightToFinalizationQueue(ctx, elem)
+	for _, fieldLength := range genesisFields {
+		if fieldLength != 0 {
+			panic("Only params can be initialized at genesis")
+		}
```
zale144, line 24:
```suggestion
	if genesisFieldsLen != 0 {
		panic("Only params can be initialized at genesis")
	}
```

zale144 (approved):
Minor suggestion

PR #780, opened 2024-04-01
fix(ante): Add missing ante handler

### LINKED ISSUE #755: Add missing `anteHandlers` methods in `newLegacyCosmosAnteHandlerEip712`
2 points here:

1. The `newLegacyCosmosAnteHandlerEip712` looks to handle Cosmos transactions, same as `newCosmosAnteHandler` with the only difference that legacy handler can handle transactions signed with Ledger. If they must be equivalent in everything else, probably the legacy handler should also use `ante.NewExtensionOptionsDecorator` and `delayedack.NewIBCProofHeightDecorator` (these two are used in "normal" handler but not in "legacy" one)

2. please double-check that the `ante.NewExtensionOptionsDecorator` is intended to be configured like it is now. The checker is nil and, by looking at the config structure, someone could think that it allows all extensions but it forbids any of them

trinitys7:
## Description

----

Closes #755 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:

### DIFF
--- a/app/ante/handlers.go
+++ b/app/ante/handlers.go
@@ -64,6 +64,7 @@ func newLegacyCosmosAnteHandlerEip712(options HandlerOptions) sdk.AnteHandler {
 		// Note: signature verification uses EIP instead of the cosmos signature validator
 		NewLegacyEip712SigVerificationDecorator(options.AccountKeeper, options.SignModeHandler),
 		ante.NewIncrementSequenceDecorator(options.AccountKeeper),
+		delayedack.NewIBCProofHeightDecorator(),
 		ibcante.NewRedundantRelayDecorator(options.IBCKeeper),
 		ethante.NewGasWantedDecorator(options.EvmKeeper, options.FeeMarketKeeper),
 	)


omritoptix (approved):

omritoptix (changes requested):
The issue talks also about `NewExtensionOptionsDecorator`. any reason we didn't include it?

trinitys7:
Currently, the app set ExtensionOptionChecker to [nil](https://github.com/dymensionxyz/dymension/blob/8aaca2247210ad31649e150c120ddb9efe45579d/app/app.go#L965). If we add it, that would make the LegacyCosmosAnteHandlerEip712 AnteHandler get `unknown extension options error`.
The LegacyCosmosAnteHandlerEip712 is an extension options, so we shouldn't add NewExtensionOptionsDecorator  into it

zale144 (approved):

omritoptix (approved):

PR #781, opened 2024-04-01
fix(rollapp): fix `RollappFinalizedStateInvariant` on fraudulent and state revert

### LINKED ISSUE #759: `RollappFinalizedStateInvariant` will break due to reverted states
> In x/rollapp/keeper/invariants.go:241-244, the
> RollappFinalizedStateInvariant invariant ensures that the finalized states are
> Status_FINALIZED. This is problematic because the invariant does not allow
> Status_REVERTED status, which can happen when reverting a fraudulent state, as seen in
> x/rollapp/keeper/fraud_handler.go:109.
> Consequently, the RollappFinalizedStateInvariant invariant will break incorrectly
> for reverted states, triggering a chain halt

Not sure if this is actually hapenning as reverted states should be registered as reverted but need to check.

hoangdv2429:
## Description

Closes #759 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:

### DIFF
--- a/x/rollapp/keeper/invariants.go
+++ b/x/rollapp/keeper/invariants.go
@@ -238,8 +238,8 @@ func RollappFinalizedStateInvariant(k Keeper) sdk.Invariant {
 					broken = true
 				}
 
-				if stateInfo.Status != commontypes.Status_FINALIZED {
-					msg += fmt.Sprintf("rollapp (%s) have stateInfo at index %d not finalized\n", rollapp.RollappId, i)
+				if stateInfo.Status != commontypes.Status_FINALIZED && stateInfo.Status != commontypes.Status_REVERTED {
+					msg += fmt.Sprintf("rollapp (%s) have stateInfo at index %d not finalized or reverted\n", rollapp.RollappId, i)
 					broken = true
 				}
 			}
--- a/x/rollapp/keeper/invariants.go
+++ b/x/rollapp/keeper/invariants.go
@@ -238,7 +238,7 @@ func RollappFinalizedStateInvariant(k Keeper) sdk.Invariant {
 					broken = true
 				}
 
-				if stateInfo.Status != commontypes.Status_FINALIZED && stateInfo.Status != commontypes.Status_REVERTED {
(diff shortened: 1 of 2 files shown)


x/rollapp/keeper/invariants.go
```diff
@@ -238,8 +238,8 @@ func RollappFinalizedStateInvariant(k Keeper) sdk.Invariant {
 					broken = true
 				}
 
-				if stateInfo.Status != commontypes.Status_FINALIZED {
-					msg += fmt.Sprintf("rollapp (%s) have stateInfo at index %d not finalized\n", rollapp.RollappId, i)
+				if stateInfo.Status != commontypes.Status_FINALIZED && stateInfo.Status != commontypes.Status_REVERTED {
```
omritoptix, line 241:
```suggestion
				if stateInfo.Status == commontypes.Status_PENDING {
```

omritoptix:
@hoangdv2429 also can we add a test for that?

x/rollapp/keeper/invariants.go
```diff
@@ -238,8 +238,8 @@ func RollappFinalizedStateInvariant(k Keeper) sdk.Invariant {
 					broken = true
 				}
 
-				if stateInfo.Status != commontypes.Status_FINALIZED {
-					msg += fmt.Sprintf("rollapp (%s) have stateInfo at index %d not finalized\n", rollapp.RollappId, i)
+				if stateInfo.Status == commontypes.Status_PENDING {
```
mtsitrin, line 241: @omritoptix I think it misses the purpose of this invariant.
all the states until the finalization index expected to be finalized.
I don't see a scenario where it's not

hoangdv2429:
Rollapp when received fraudulent will be frozen, finalized state will stay that way and pending states will be reverted. So no invariant is being broken.

PR #860, opened 2024-04-23
feat(delayedack): Add type filter for delayedack query

### LINKED ISSUE #850: Add ability to filter delayedack by type
currently its only by rollapp id and status

zale144:
## Description

Add delayedack filter by type

----

Closes #850 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


x/delayedack/client/cli/query.go
```diff
@@ -84,20 +79,33 @@ func CmdGetPacketsByRollapp() *cobra.Command {
...
 				if !ok {
 					// Handle error: statusStr is not a valid commontypes.Status
 					return fmt.Errorf("invalid status: %s", statusStr)
 				}
-				req.Status = status
+				req.Status = commontypes.Status(status)
+			}
+
+			if len(args) > 2 {
+				typeStr := strings.ToUpper(args[2])
```
trinitys7, line 96: As I understand in the code, in order to filter by type, we need to filter it by status first. We should make the code separate between status and type.

zale144: in `packets-by-rollapp` the status and type are optional filters. In `packets-by-status` the type is optional. But you still have `packets-by-type` so you can use that to filter by type alone, granted the status will be PENDING by default.

trinitys7: What I mean is like, when you assert like that, the filter must always be status first, and then after that it can be type. So it's kinda annoying if I want to filter just by type

omritoptix:
@zale144 I would avoid the refactor of moving types to the types common for 2 reasons:
1. I don't think it's a common `type` around the system (unlike `status`) as it's rollapp packet specific.
2. I would like to avoid all the refactors to simplify this pr and it's reviewing. 

if there is a way to avoid it and minimize changes would be preferred imo unless I'm missing something.

mtsitrin (approved):

mtsitrin (approved):

mtsitrin (approved):

PR #900, opened 2024-05-24
feat: Register IBC denom on transfer

zale144:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


danwt:
Please could you make and link an issue, and cross link the rdk pr?

danwt:
e2e tests fail, is this the version mismatch that @omritoptix mentioned in today's standup?

app/app.go
```diff
@@ -749,7 +758,13 @@ func New(
...
 		packetforwardkeeper.DefaultRefundTransferPacketTimeoutTimestamp,
 	)
-	transferStack = delayedackmodule.NewIBCMiddleware(transferStack, app.DelayedAckKeeper)
+	middleware := delayedackmodule.NewIBCMiddleware(transferStack, app.DelayedAckKeeper, app.RollappKeeper)
+	transferStack = middleware
+	app.TransferInjectKeeper.SetMiddleware(
+		transferStack,
+		middleware,
+	)
+	transferStack = app.TransferInjectKeeper
```
danwt, line 767: I find this section a bit hard to read
any way to restructure, rename stuff to make it more readable?

x/delayedack/ibc_middleware.go
```diff
@@ -52,18 +51,29 @@ func (im IBCMiddleware) OnRecvPacket(
...
 	if err != nil {
 		logger.Error("Failed to extract rollapp id from packet", "err", err)
 		return channeltypes.NewErrorAcknowledgement(err)
 	}
 
+	if rollapp == nil {
+		return im.IBCModule.OnRecvPacket(ctx, packet, relayer)
+	}
+
+	rollappID := rollapp.RollappId
```
danwt, line 69: why was this added?

x/delayedack/ibc_middleware.go
```diff
@@ -134,17 +144,23 @@ func (im IBCMiddleware) OnAcknowledgementPacket(
...
 	if err != nil {
 		logger.Error("Failed to extract rollapp id from channel", "err", err)
 		return err
 	}
 
+	if rollapp == nil {
+		return im.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+
+	rollappID := rollapp.RollappId
```
danwt, line 157: ditto: why was this added?

x/delayedack/ibc_middleware.go
```diff
@@ -220,18 +236,29 @@ func (im IBCMiddleware) OnTimeoutPacket(
...
 	if err != nil {
 		logger.Error("Failed to extract rollapp id from channel", "err", err)
 		return err
 	}
 
+	if rollapp == nil {
+		return im.IBCModule.OnTimeoutPacket(ctx, packet, relayer)
+	}
+
+	rollappID := rollapp.RollappId
```
danwt, line 254: ditto

x/delayedack/ibc_middleware.go
```diff
@@ -285,66 +312,6 @@ func (im IBCMiddleware) OnTimeoutPacket(
...
-	if rollapp.ChannelId != rollappChannelOnHub {
-		return "", &data, errorsmod.Wrapf(
-			rollapptypes.ErrMismatchedChannelID,
-			"channel id mismatch: expect: %s: got: %s", rollapp.ChannelId, rollappChannelOnHub,
-		)
-	}
-
-	return chainID, &data, nil
-}
-
```
danwt, line 347: nice change ❤️

x/rollapp/keeper/util.go
```diff
@@ -0,0 +1,78 @@
...
+func (k Keeper) ExtractRollappAndTransferPacketFromData(
+	ctx sdk.Context,
+	data []byte,
+	rollappPortOnHub string,
+	rollappChannelOnHub string,
+) (*types.Rollapp, *transfertypes.FungibleTokenPacketData, error) {
+	// no-op if the packet is not a fungible token packet
+	packet := new(transfertypes.FungibleTokenPacketData)
+	if err := types.ModuleCdc.UnmarshalJSON(data, packet); err != nil {
+		return nil, nil, errorsmod.Wrapf(err, "failed to unmarshal packet data")
```
danwt, line 24: no failed to

x/rollapp/keeper/util.go
```diff
@@ -0,0 +1,78 @@
...
+	rollapp, err := k.ExtractRollappFromChannel(ctx, rollappPortOnHub, rollappChannelOnHub)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return rollapp, packet, nil
+}
+
+// ExtractRollappFromChannel extracts the rollapp from the IBC port and channel
+func (k Keeper) ExtractRollappFromChannel(
```
danwt, line 36: can it be clear in the docstring that nil will be returned if it's not found

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+func (t *Keeper) SetMiddleware(
+	ibc porttypes.IBCModule,
+	ics porttypes.ICS4Wrapper,
+) {
+	t.IBCModule = ibc
+	t.ICS4Wrapper = ics
+}
```
danwt, line 48: This keeper is also itself middleware right (or not?)

Can we make it clear in that case that this module and wrapper are the 'next' middleware(s), I suppose

Maybe better to have two setters

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	ctx sdk.Context,
+	chanCap *capabilitytypes.Capability,
+	destinationPort string, destinationChannel string,
+	timeoutHeight clienttypes.Height,
+	timeoutTimestamp uint64,
+	data []byte,
+) (sequence uint64, err error) {
+	rollapp, err := t.rollappKeeper.ExtractRollappFromChannel(ctx, destinationPort, destinationChannel)
+	if err != nil {
+		return 0, fmt.Errorf("cannot extract rollapp id from packet: %w", err)
```
danwt, line 61: no cannot

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	// TODO: consider that other chains also might want this feature
+	if rollapp == nil {
+		return t.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	packet := new(transfertypes.FungibleTokenPacketData)
+	if err = types.ModuleCdc.UnmarshalJSON(data, packet); err != nil {
+		return 0, errorsmod.Wrapf(errortypes.ErrUnknownRequest, "cannot unmarshal ICS-20 transfer packet data: %s", err.Error())
```
danwt, line 70: no cannot

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// check if the rollapp already contains the denom metadata
+	if ContainsFunc(rollapp.TokenMetadata, func(dm *rtypes.TokenMetadata) bool { return dm.Base == packet.Denom }) {
+		return t.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	denomMetadata, ok := t.bankKeeper.GetDenomMetaData(ctx, packet.Denom)
+	if !ok {
+		return 0, errorsmod.Wrapf(errortypes.ErrInvalidType, "denom metadata not found")
```
danwt, line 80: .ErrNotFound

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	packet.Memo, err = types.AddDenomMetadataToMemo(packet.Memo, denomMetadata)
+	if err != nil {
+		return 0, errorsmod.Wrapf(errortypes.ErrInvalidType, "cannot create injected value")
```
danwt, line 85: not a useful message

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	data, err = types.ModuleCdc.MarshalJSON(packet)
+	if err != nil {
+		return 0, errorsmod.Wrapf(errortypes.ErrInvalidType, "cannot marshal ICS-20 transfer packet data")
```
danwt, line 90: no cannot

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+// OnAcknowledgementPacket adds the token metadata to the rollapp if it doesn't exist
+func (t *Keeper) OnAcknowledgementPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	acknowledgement []byte,
+	relayer sdk.AccAddress,
+) error {
+	var data transfertypes.FungibleTokenPacketData
+	if err := types.ModuleCdc.UnmarshalJSON(packet.GetData(), &data); err != nil {
+		return errorsmod.Wrapf(errortypes.ErrUnknownRequest, "cannot unmarshal ICS-20 transfer packet data: %s", err.Error())
```
danwt, line 105: no cannot

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	if errors.Is(err, types.ErrMemoUnmarshal) || errors.Is(err, types.ErrMemoDMEmpty) {
+		return t.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+	if err != nil {
+		return err
+	}
+
+	rollapp, err := t.rollappKeeper.ExtractRollappFromChannel(ctx, packet.SourcePort, packet.SourceChannel)
+	if err != nil {
+		return fmt.Errorf("cannot extract rollapp id from packet: %w", err)
```
danwt, line 122: no cannot

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if rollapp == nil {
+		return t.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+
+	dm := packetMetaData.DenomMetadata
+
+	if ContainsFunc(rollapp.TokenMetadata, func(tm *rtypes.TokenMetadata) bool { return tm.Base == dm.Base }) {
+		return t.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
```
danwt, line 132: can you add a a comment explaining why total equality is not used

danwt: also add a comment explaining why it short circuits
(I understand why, but for future reading)

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	if rollapp == nil {
+		return t.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	packet := new(transfertypes.FungibleTokenPacketData)
+	if err = types.ModuleCdc.UnmarshalJSON(data, packet); err != nil {
+		return 0, errorsmod.Wrapf(errortypes.ErrUnknownRequest, "cannot unmarshal ICS-20 transfer packet data: %s", err.Error())
+	}
+
+	// check if the rollapp already contains the denom metadata
```
danwt, line 73: not a very useful comment

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	if ContainsFunc(rollapp.TokenMetadata, func(dm *rtypes.TokenMetadata) bool { return dm.Base == packet.Denom }) {
```
danwt, line 74: please DRY out this function with the acknowledgment one and document/explain it

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+		return t.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
```
danwt, line 75: have you thought about if you need to check if memo already exists here too? ⚠️

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	dm := packetMetaData.DenomMetadata
+
+	if ContainsFunc(rollapp.TokenMetadata, func(tm *rtypes.TokenMetadata) bool { return tm.Base == dm.Base }) {
+		return t.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+
+	// check if hub has the denom metadata
+	if !t.bankKeeper.HasDenomMetaData(ctx, dm.Base) {
+		return errorsmod.Wrapf(errortypes.ErrInvalidType, "denom metadata not found")
+	}
```
danwt, line 137: why is this block here?
can you add a comment explaining
why would this happen
why we would return an error and the impact of that

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/dymensionxyz/dymension/v3/x/transferinject/types"
+)
+
+type Keeper struct {
+	cdc codec.BinaryCodec
+	porttypes.IBCModule
+	porttypes.ICS4Wrapper
+	rollappKeeper types.RollappKeeper
+	bankKeeper    types.BankKeeper
+}
```
danwt, line 28: With the chosen protoco, it is possible to send the denom metadata UNTIL the ack comes back

which migh result in duplicated sends

Can you document this somewhere in a comment and state that we're ok with that


(cc @omritoptix I guess we're ok with it right?)

x/transferinject/types/expected_keepers.go
```diff
@@ -0,0 +1,23 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/bank/types"
+
+	rollapptypes "github.com/dymensionxyz/dymension/v3/x/rollapp/types"
+)
+
+// BankKeeper defines the expected interface needed
+type BankKeeper interface {
+	GetDenomMetaData(ctx sdk.Context, denom string) (types.Metadata, bool)
+	HasDenomMetaData(ctx sdk.Context, denom string) bool
```
danwt, line 13: it's a nit, but here's only one usage of Has
Could we just have `Get`?

x/transferinject/types/packet_metadata.go
```diff
@@ -0,0 +1,64 @@
...
+func (p PacketMetadata) ValidateBasic() error {
+	return p.DenomMetadata.Validate()
+}
+
+const memoObjectKeyDM = "denom_metadata"
+
+var (
+	ErrMemoUnmarshal          = fmt.Errorf("unmarshal memo")
+	ErrDenomMetadataUnmarshal = fmt.Errorf("unmarshal denom metadata")
+	ErrMemoDMEmpty            = fmt.Errorf("memo denom metadata field is missing")
```
danwt, line 23: please dont abbreviate to dm

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	cdc codec.BinaryCodec,
+	rollappKeeper types.RollappKeeper,
+	bankKeeper types.BankKeeper,
+) *Keeper {
+	return &Keeper{
+		cdc:           cdc,
+		rollappKeeper: rollappKeeper,
+		bankKeeper:    bankKeeper,
+	}
+}
```
danwt, line 40: can you add a overview docstring here or just in the file explaining at a high level what this package and keeper does and is for

danwt (changes requested):
**Architecturally:**
I wonder if we can do better by decoupling the memo protection from the denom metadata concepts. We have here a package called 'transferinject' which is quite abstract and general sounding but it only specifically supports this denom metadata memo stuff

Happy to merge like that and improve later though

**Security and tests**

Need to test and define correct behavior for
1) memo not included and denom meta not included (happy A)
2) memo not included and denom meta already sent and acked (happy B)
2.5) memo not included and denom meta already sent but not acked 
3) memo included and denom meta not included (attack)
4) memo included and denom meta already sent and acked (attack)

etc

danwt:
also please stop adding 'failed to' etc to errors 😄

app/app.go
```diff
@@ -655,6 +644,26 @@ func New(
...
+		app.RollappKeeper,
+		app.BankKeeper,
+	)
+
+	// Create Transfer Keepers
+	app.TransferKeeper = ibctransferkeeper.NewKeeper(
+		appCodec,
+		keys[ibctransfertypes.StoreKey],
+		app.GetSubspace(ibctransfertypes.ModuleName),
+		app.TransferInjectKeeper,
```
mtsitrin, line 658: the ICS4Wrapper should be left `app.IBCKeeper.ChannelKeeper` 
as this is the layer that provides the writing capability for packets and acks

x/transferinject/keeper/keeper.go
```diff
@@ -0,0 +1,168 @@
...
+	transfertypes "github.com/cosmos/ibc-go/v6/modules/apps/transfer/types"
+	clienttypes "github.com/cosmos/ibc-go/v6/modules/core/02-client/types"
+	channeltypes "github.com/cosmos/ibc-go/v6/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/v6/modules/core/05-port/types"
+
+	rtypes "github.com/dymensionxyz/dymension/v3/x/rollapp/types"
+	"github.com/dymensionxyz/dymension/v3/x/transferinject/types"
+)
+
+type Keeper struct {
```
mtsitrin, line 22: I don't think u need a keeper. u don;t manage store and not exposing any APIs.

IMO u need 
```go
// Middleware implements the ICS26 Module interface
type Middleware interface {
    porttypes.IBCModule // middleware has acccess to an underlying application which may be wrapped by more middleware
    ics4Wrapper: ICS4Wrapper // middleware has access to ICS4Wrapper which may be core IBC Channel Handler or a higher-level middleware that wraps this middleware.
}
```

app/app.go
```diff
@@ -362,8 +363,10 @@ type App struct {
 	EIBCKeeper      eibckeeper.Keeper
 
 	// this line is used by starport scaffolding # stargate/app/keeperDeclaration
-	DelayedAckKeeper    delayedackkeeper.Keeper
-	DenomMetadataKeeper *denommetadatamodulekeeper.Keeper
+	DelayedAckKeeper     delayedackkeeper.Keeper
+	DenomMetadataKeeper  *denommetadatamodulekeeper.Keeper
+	TransferInjectKeeper *transferinjectkeeper.Keeper
```
mtsitrin, line 368: u don't need two transferKeepers
u probably need the IBC transfer keeper, and pass it your middleware as ICS4 wrapper

app/app.go
mtsitrin: I would stick with the `transferStack` notation. easier to read IMO

x/delayedack/ibc_middleware.go
```diff
@@ -23,14 +20,16 @@ var _ porttypes.Middleware = &IBCMiddleware{}
 // IBCMiddleware implements the ICS26 callbacks
 type IBCMiddleware struct {
 	porttypes.IBCModule
-	keeper keeper.Keeper
+	keeper.Keeper
```
mtsitrin, line 23: not sure why u prefer to embed the keeper it this way
the `DelayedAck` keeper provides "store access" service to the middleware
IMO was more explicit and better readable before

x/delayedack/ibc_middleware.go
zale144: transient change

x/transferinject/keeper/keeper.go
zale144: got it, will replace with 'failed to'

x/transferinject/types/expected_keepers.go
zale144: fine

zale144:
> also please stop adding 'failed to' etc to errors 😄

ikr, it's a habit. Probably coming from some baggage from my upbringing. "I'm not only an error, but also a failure" 😭

x/delayedack/ibc_middleware.go
zale144: the idea was to be able to remove the IBC4 wrapper below, starting line 288

app/app.go
```diff
@@ -625,7 +627,6 @@ func New(
 		keys[rollappmoduletypes.MemStoreKey],
 		app.GetSubspace(rollappmoduletypes.ModuleName),
 		app.IBCKeeper.ClientKeeper,
-		app.TransferKeeper,
```
mtsitrin, line 628: missing `SetTransferKeeper` instead?

app/app.go
mtsitrin, line 1: not sure where the `SendPacket` of the middleware is actually wired.
the `transferStack` is used for incoming packets, not outgoing

app/app.go
```diff
@@ -749,7 +750,9 @@ func New(
 		packetforwardkeeper.DefaultForwardTransferPacketTimeoutTimestamp,
 		packetforwardkeeper.DefaultRefundTransferPacketTimeoutTimestamp,
 	)
-	transferStack = delayedackmodule.NewIBCMiddleware(transferStack, app.DelayedAckKeeper)
+	middleware := delayedackmodule.NewIBCMiddleware(transferStack, app.DelayedAckKeeper, app.RollappKeeper)
```
mtsitrin, line 753: can have more descriptive name

x/rollapp/keeper/keeper.go
```diff
@@ -160,6 +159,13 @@ func (k Keeper) mintRollappGenesisTokens(ctx sdk.Context, rollapp types.Rollapp)
 	return nil
 }
 
+func (k *Keeper) SetTransferKeeper(tk types.TransferKeeper) {
```
mtsitrin, line 162: add comment that this is a required call

app/app.go
zale144: `SendPacket` is part of ICS4Wrapper, which is wired through delayedack

mtsitrin: not sure I got it.
when one transfers hub->rollapp, it submits `MsgTransfer` which is handled by the `transferKeeper`

zale144: yeah, me neither.

but, yeah, `transferKeeper` then calls `ics4Wrapper.SendPacket`, and then we end up in our transferStack

mtsitrin: but transferKeeper gets `app.IBCKeeper.ChannelKeeper,` as it's `ics4Wrapper`

app/app.go
danwt: was talking more about the middleware

x/delayedack/ibc_middleware.go
danwt: what's that 😄

x/transferinject/ibc_middleware_test.go
```diff
@@ -0,0 +1,461 @@
...
+				},
+			},
+			data: types.ModuleCdc.MustMarshalJSON(&transfertypes.FungibleTokenPacketData{
+				Denom: "adym",
+			}),
+			wantSentData: types.ModuleCdc.MustMarshalJSON(&transfertypes.FungibleTokenPacketData{
+				Denom: "adym",
+				Memo:  types.AddDenomMetadataToMemo("", validDenomMetadata),
+			}),
+			wantErr: "",
```
danwt, line 52: you dont need these lines 
`wantErr: "",`

x/transferinject/ibc_middleware_test.go
```diff
@@ -0,0 +1,461 @@
...
+				ICS4Wrapper: &mockICS4Wrapper{},
+				rollappKeeper: &mockRollappKeeper{
+					returnRollapp: &rollapptypes.Rollapp{},
+				},
+				bankKeeper: mockBankKeeper{
+					returnMetadata: validDenomMetadata,
+				},
+			},
+			data: types.ModuleCdc.MustMarshalJSON(&transfertypes.FungibleTokenPacketData{
+				Denom: "adym",
```
danwt, line 46: why not put the types.ModuleCdc.MustMarshalJSON calls once, in the execution part, and make the data field the &transfertypes.FungibleTokenPacketDat type?
Less characters oeverall

x/transferinject/ibc_middleware_test.go
```diff
@@ -0,0 +1,461 @@
...
+	channeltypes "github.com/cosmos/ibc-go/v6/modules/core/04-channel/types"
+	porttypes "github.com/cosmos/ibc-go/v6/modules/core/05-port/types"
+	"github.com/stretchr/testify/require"
+
+	rollapptypes "github.com/dymensionxyz/dymension/v3/x/rollapp/types"
+	"github.com/dymensionxyz/dymension/v3/x/transferinject"
+	"github.com/dymensionxyz/dymension/v3/x/transferinject/types"
+)
+
+func TestIBCMiddleware_SendPacket(t *testing.T) {
```
danwt, line 21: need to use errorsmod.IsOf or errors.Is instead of string comparison for all these errors

danwt: Some of the subtest names could be better 
This one is OK: `name: "return early: rollapp already has token metadata",`
Because it tells me what is being tested (the fact that we catch it if it already has data)

But the rest of them dont really tell me what's being tested, they just say what happens or what is returned

x/transferinject/ibc_middleware_test.go
```diff
@@ -0,0 +1,461 @@
...
+		}, {
+			Denom:    "DYM",
+			Exponent: 18,
+		},
+	},
+}
+
+func tokenMetadata(metadata banktypes.Metadata) *rollapptypes.TokenMetadata {
+	return transferinject.DenomToTokenMetadata(&metadata)
+}
```
danwt, line 405: is this useful? it doesnt save much and adds an indirection

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	if hasDenom(rollapp.TokenMetadata, packet.Denom) {
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	denomMetadata, ok := m.bankKeeper.GetDenomMetaData(ctx, packet.Denom)
+	if !ok {
+		return 0, errorsmod.Wrapf(errortypes.ErrNotFound, "denom metadata not found")
```
danwt, line 92: great job here in making use of the existing sdk errors ⭐

x/transferinject/types/packet_metadata.go
```diff
@@ -0,0 +1,57 @@
...
+	DenomMetadata *types.Metadata `json:"denom_metadata"`
+}
+
+func (p PacketMetadata) ValidateBasic() error {
+	return p.DenomMetadata.Validate()
+}
+
+var (
+	ErrMemoUnmarshal          = fmt.Errorf("unmarshal memo")
+	ErrMemoDenomMetadataEmpty = fmt.Errorf("memo denom metadata field is missing")
```
danwt, line 26: arguably you could wrap sdkerrors.ErrInvalidType instead

x/transferinject/types/packet_metadata.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/x/bank/types"
+)
+
+// MemoData represents the structure of the memo with user and hub metadata
+type MemoData struct {
+	UserMemo       string          `json:"user_memo"`
+	PacketMetadata *PacketMetadata `json:"packet_metadata"`
+}
```
danwt, line 14: This is good but I have a suggestion that will bring us more inline with recommended practices, and be more maintainable

First of all, taking care of the user memo should be out of scope of this middleware
We should have an additional middleware (can be very lightweight, just 1 file) which always executes first, which takes the user memo and puts it in a json `{"user":<xxx>"}`

Then, your middleware can assume the memo is already json (map[string]any)
Then your middleware should simply add another key to this json

{
  "user": "foobar",
  "transferinject" : {
    "denom_metadata" : ...
  },
  "other_module" : ...
}


I feel quite strongly about this change tbh, especially because I will already add another memo type in the work I'm doing
And it's not unlikely we will add more later

This is also best practice according to https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer#using-the-memo-field

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+	data []byte,
+) (sequence uint64, err error) {
+	packet := new(transfertypes.FungibleTokenPacketData)
+	if err = types.ModuleCdc.UnmarshalJSON(data, packet); err != nil {
+		return 0, errorsmod.Wrapf(errortypes.ErrUnknownRequest, "unmarshal ICS-20 transfer packet data: %s", err.Error())
+	}
+
+	if types.MemoAlreadyHasPacketMetadata(packet.Memo) {
+		return 0, errorsmod.Wrapf(errortypes.ErrUnauthorized, "denom metadata already exists in memo")
+	}
```
danwt, line 75: just curious why this is above the 
if rollapp == nil {

?

x/transferinject/ibc_middleware.go
danwt, line 92: when can this occur? needs a comment

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+// OnAcknowledgementPacket adds the token metadata to the rollapp if it doesn't exist
+func (m *IBCMiddleware) OnAcknowledgementPacket(
+	ctx sdk.Context,
+	packet channeltypes.Packet,
+	acknowledgement []byte,
+	relayer sdk.AccAddress,
+) error {
+	var ack channeltypes.Acknowledgement
+	if err := types.ModuleCdc.UnmarshalJSON(acknowledgement, &ack); err != nil {
+		return errorsmod.Wrapf(rtypes.ErrUnknownRequest, "unmarshal ICS-20 transfer packet acknowledgement: %v", err)
```
danwt, line 114: fyi, there is an sdk error for failing to unmarshal json
but lets leave it for now

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	memoData, err := types.ParseMemoData(data.Memo)
+	if errors.Is(err, types.ErrMemoUnmarshal) || errors.Is(err, types.ErrMemoDenomMetadataEmpty) {
+		return m.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+	if err != nil {
+		return err
+	}
+
```
danwt, line 137: I don't think there should be an additional if len==0 check
If it's empty, the types.ParseMemoData should return some appropriate error which can then be checked with errors.Is

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+	if memoData.PacketMetadata == nil || memoData.PacketMetadata.DenomMetadata == nil {
+		return m.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
```
danwt, line 140: explain why packetMetadata can be not nil but .DenomMetadata can have a value

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+
+	rollapp, err := m.rollappKeeper.ExtractRollappFromChannel(ctx, packet.SourcePort, packet.SourceChannel)
+	if err != nil {
+		return errorsmod.Wrapf(errortypes.ErrNotFound, "extract rollapp id from packet: %s", err.Error())
```
danwt, line 144: ~You lose type info here by dumping the error string~
~let's leave it like this for now~
~I will take a look at the sdk error code and try to come up with something better~

danwt: its not really a 'not found error' here
just return the error

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,193 @@
...
+	if hasDenom(rollapp.TokenMetadata, dm.Base) {
+		return m.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
+	}
+
+	// add the new token metadata to the rollapp
+	rollapp.TokenMetadata = append(rollapp.TokenMetadata, DenomToTokenMetadata(dm))
+
+	m.rollappKeeper.SetRollapp(ctx, *rollapp)
+
+	return m.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
```
danwt, line 161:
```
	dm := memoData.PacketMetadata.DenomMetadata

	if !hasDenom(rollapp.TokenMetadata, dm.Base) {

		// add the new token metadata to the rollapp
		rollapp.TokenMetadata = append(rollapp.TokenMetadata, DenomToTokenMetadata(dm))

		m.rollappKeeper.SetRollapp(ctx, *rollapp)
	}

	return m.IBCModule.OnAcknowledgementPacket(ctx, packet, acknowledgement, relayer)
```

danwt (changes requested):
Good changes, but I've still a couple requests

x/transferinject/ibc_middleware.go
zale144: well if we don't find the rollapp, it means we're sending to a non-rollapp chain, so we don't include the metadata, but that's a simplistic way to check if the other chain supports this middleware. In the future it would be nice to consider finding a nicer way to discriminate the receiving end's support for this middleware

app/app.go
```diff
@@ -761,7 +764,7 @@ func New(
 	app.RollappKeeper.SetHooks(rollappmoduletypes.NewMultiRollappHooks(
 		// insert rollapp hooks receivers here
 		app.SequencerKeeper.RollappHooks(),
-		transferStack.(delayedackmodule.IBCMiddleware),
+		delayedAckMiddleware,
```
mtsitrin, line 767: `transferStack`

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,199 @@
...
+	// meaning, find a better way to check if the receiving chain supports this middleware
+	if rollapp == nil {
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	if hasDenom(rollapp.TokenMetadata, packet.Denom) {
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	// get the denom metadata from the bank keeper, if it doesn't exist, return an error
```
mtsitrin, line 79: Is this feature discussed in the ADR? I think it can be limiting for the future
@omritoptix wdyt

app/app.go
zale144: what's the difference between `transferStack.(delayedackmodule.IBCMiddleware)` and `delayedAckMiddleware` ?

danwt (dismissed):
nice work!
please make sure Michaels compat comment is resolved before merging

x/transferinject/ibc_middleware.go
danwt: ok thanks

x/transferinject/ibc_middleware.go
danwt: >be promiscuous with the hub and have each rollapp handle it based on it's limitations.
@omritoptix can you expand this? Not sure I got you
Thanks

zale144: Ok, now if denom metadata is not found, we don't reject but just move on with the middleware chain

danwt (approved):

app/app.go
mtsitrin: my bad

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,196 @@
...
+	if err != nil {
+		return errorsmod.Wrapf(errortypes.ErrInvalidRequest, "extract rollapp from packet: %s", err.Error())
+	}
+	if rollapp == nil {
+		return errorsmod.Wrapf(errortypes.ErrNotFound, "rollapp not found")
+	}
+
+	if !hasDenom(rollapp.TokenMetadata, dm.Base) {
+		// add the new token metadata to the rollapp
+		rollapp.TokenMetadata = append(rollapp.TokenMetadata, DenomToTokenMetadata(dm))
```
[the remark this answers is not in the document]
mtsitrin: I think 2 makes sense

x/transferinject/ibc_middleware.go
danwt: did you push the change?

danwt (changes requested):

x/transferinject/ibc_middleware.go
zale144: yeah

danwt (approved):
approved pending resolution to Michaels comments

proto/dymension/rollapp/rollapp.proto
```diff
@@ -41,12 +41,14 @@ message Rollapp {
   repeated string permissionedAddresses = 5;
   // tokenMetadata is a list of TokenMetadata that are registered on this rollapp
   repeated TokenMetadata tokenMetadata = 6;
+  // registeredDenoms is a list of registered denoms on this rollapp
+  repeated string registeredDenoms = 7;
```
mtsitrin, line 45: let make this variable `=10`, to avoid compatibility issues
otherwise, store migration needed

mtsitrin (changes requested):

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,165 @@
...
+	if rollapp == nil {
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	// Check if the rollapp already contains the denom metadata by matching the base of the denom metadata.
+	// At the first match, we assume that the rollapp already contains the metadata.
+	// It would be technically possible to have a race condition where the denom metadata is added to the rollapp
+	// from another packet before this packet is acknowledged.
+	if Contains(rollapp.RegisteredDenoms, packet.Denom) {
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
```
danwt, line 77: here you use packet.Denom but in the ack code you use base
Can it be consistent?

proto/dymension/rollapp/rollapp.proto
```diff
@@ -41,12 +41,14 @@ message Rollapp {
   repeated string permissionedAddresses = 5;
   // tokenMetadata is a list of TokenMetadata that are registered on this rollapp
   repeated TokenMetadata tokenMetadata = 6;
+  // registeredDenoms is a list of registered denoms on this rollapp
```
danwt, line 44: >list of registered denoms

Is it a base or id or what?

danwt (changes requested):

danwt:
Now that we only refer to the denom by a short string, what do we do if it get's updated in the bank? We need to ship the update or what?

Is this specced out/documented?

zale144:
> Now that we only refer to the denom by a short string, what do we do if it get's updated in the bank? We need to ship the update or what?
> 
> Is this specced out/documented?

The base is the key, which doesn't get updated, but if anything else is updated, the base/key still references the same record in the bank

danwt:
So the rollapp wouldn't necessarily get the update then?

x/transferinject/ibc_middleware.go
```diff
@@ -0,0 +1,165 @@
...
+	// meaning, find a better way to check if the receiving chain supports this middleware
+	if rollapp == nil {
+		return m.ICS4Wrapper.SendPacket(ctx, chanCap, destinationPort, destinationChannel, timeoutHeight, timeoutTimestamp, data)
+	}
+
+	// Check if the rollapp already contains the denom metadata by matching the base of the denom metadata.
+	// At the first match, we assume that the rollapp already contains the metadata.
+	// It would be technically possible to have a race condition where the denom metadata is added to the rollapp
+	// from another packet before this packet is acknowledged.
+	if Contains(rollapp.RegisteredDenoms, packet.Denom) {
```
mtsitrin, line 76: @zale144 don't u need to validate it's not a token sourced from the rollapp? otherwise u register the rollapp's token as it represented on the hub, makes it wierdly duplicate record on the rollapp

x/transferinject/types/packet_metadata.go
```diff
@@ -0,0 +1,74 @@
...
+
+var (
+	ErrMemoTransferInjectEmpty         = fmt.Errorf("memo 'transferinject' is missing")
+	ErrMemoTransferInjectAlreadyExists = errorsmod.Wrapf(errortypes.ErrUnauthorized, "'transferinject' already exists in memo")
+)
+
+func ParsePacketMetadata(input string) (*TransferInject, error) {
+	bz := []byte(input)
+
+	var memo MemoData
```
mtsitrin, line 35: @zale144 why sometimes u unmarshal into `MemoData` and sometimes into `memoMap := make(map[string]any)`

x/transferinject/ibc_middleware.go
zale144: the denom here is used as a key/reference, as it's found/represented on the hub (same one in the ack handler), just as an optimization, to avoid sending denom metadata in every packet. the point is to deduplicate on the hub.

mtsitrin: the entire middleware should work only for tokens other than tokens sourced on the rollapp.

danwt:
Conflicts

danwt:
Is this ready to be merged @zale144 ?
Did you document the fact that this will not support updates that do not change the base?

zale144:
> Is this ready to be merged @zale144 ? Did you document the fact that this will not support updates that do not change the base?

@danwt I don't see why not. Just added the bit about updates to the ADR

mtsitrin (approved):

PR #915, opened 2024-06-05
feat(eibc): update demand order

### LINKED ISSUE #902: Modifable demand order
Allow to modify an existing demand order

We should add new Tx, called `UpdateDemandOrder ` which allows to change the fee of the demand order
Modifying the demand order still needs to comply with validation rules:
- order is PENDING
- fee cannot be larger than amount
- ~~fee cannot be smaller than bridging fee (can’t be “losing” demand order)~~
- Only the `demandOrder.Recipient` can modify the order

After changing the demand order, a dedicated event for demand order change will be emitted

When submitting `MsgFulfillOrder` , the submitter **will include the expected fee**.
It mitigates the scenario where the fee price reduces after `MsgFulfillOrder` was sent **but before it was executed.**

ADR: https://www.notion.so/dymension/ADR-x-eIBC-Fee-Market-87b8e1ec78664413ba7a056852f5af77?pvs=4

mtsitrin:
Closes #902 

ADR: https://www.notion.so/dymension/ADR-x-eIBC-Fee-Market-87b8e1ec78664413ba7a056852f5af77?pvs=4

## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to update demand orders.
  - Added expected fee validation when fulfilling orders.

- **Improvements**
  - Enhanced the order fulfillment process with additional fee validations and better error handling.

- **Bug Fixes**
  - Corrected typos and improved error messages for better clarity.

- **CLI Updates**
  - Renamed commands and updated descriptions for better usability.
  - Added new command for updating demand orders.


x/eibc/keeper/msg_server.go
```diff
@@ -23,25 +26,17 @@ var _ types.MsgServer = msgServer{}
...
-	// Check that the msg is valid
+
 	err := msg.ValidateBasic()
 	if err != nil {
 		return nil, err
 	}
-	// Check that the order exists in status PENDING
-	demandOrder, err := m.GetDemandOrder(ctx, commontypes.Status_PENDING, msg.OrderId)
+
+	demandOrder, err := m.ValidateOrderIsMutable(ctx, msg.OrderId)
```
zale144, line 35: I would still suggest separating the retrieval from the validation of the order

x/eibc/keeper/msg_server.go
```diff
@@ -66,3 +61,74 @@ func (m msgServer) FulfillOrder(goCtx context.Context, msg *types.MsgFulfillOrde
...
+func (m msgServer) UpdateDemandOrder(goCtx context.Context, msg *types.MsgUpdateDemandOrder) (*types.MsgUpdateDemandOrderResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	err := msg.ValidateBasic()
+	if err != nil {
+		return nil, err
+	}
+
+	// Check that the order exists in status PENDING
+	demandOrder, err := m.ValidateOrderIsMutable(ctx, msg.OrderId)
```
zale144, line 75: same as above

x/eibc/types/errors.go
```diff
@@ -15,9 +15,7 @@ var (
 	ErrInvalidRecipientAddress      = errorsmod.Register(ModuleName, 8, "Invalid recipient address")
 	ErrBlockedAddress               = errorsmod.Register(ModuleName, 9, "Can't purchase demand order for recipient with blocked address")
 	ErrDemandAlreadyFulfilled       = errorsmod.Register(ModuleName, 10, "Demand order already fulfilled")
-	ErrNegativeTimeoutFee           = errorsmod.Register(ModuleName, 11, "Timeout fee must be greater than or equal to 0")
-	ErrTooMuchTimeoutFee            = errorsmod.Register(ModuleName, 12, "Timeout fee must be less than or equal to the total amount")
-	ErrNegativeErrAckFee            = errorsmod.Register(ModuleName, 13, "Error acknowledgement fee must be greater than or equal to 0")
-	ErrTooMuchErrAckFee             = errorsmod.Register(ModuleName, 14, "Error acknowledgement fee must be less than or equal to the total amount")
-	ErrDemandOrderNotProfitable     = errorsmod.Register(ModuleName, 15, "Demand order not profitable")
+	ErrTooMuchFee                   = errorsmod.Register(ModuleName, 11, "Fee must be less than or equal to the total amount")
```
zale144, line 18:
```suggestion
	ErrFeeTooHigh                   = errorsmod.Register(ModuleName, 11, "Fee must be less than or equal to the total amount")
```

proto/dymension/eibc/tx.proto
```diff
@@ -1,19 +1,32 @@
...
     rpc FulfillOrder(MsgFulfillOrder) returns (MsgFulfillOrderResponse) {}
+    rpc UpdateDemandOrder(MsgUpdateDemandOrder) returns (MsgUpdateDemandOrderResponse) {}
 }
 
 // MsgFulfillOrder defines the FulfillOrder request type.
 message MsgFulfillOrder {
+    option (cosmos.msg.v1.signer) = "fulfiller_address";
     // fulfiller_address is the bech32-encoded address of the account which the message was sent from.
     string fulfiller_address = 1;
     string order_id = 2;
```
zale144, line 19: should we not also have the original order fee included here, for comparing with the current one, in case it was updated in the meantime?

zale144 (commented):
Nice work! Left some comments

x/eibc/keeper/msg_server.go
```diff
@@ -23,25 +26,25 @@ var _ types.MsgServer = msgServer{}
...
-	// Check that the msg is valid
+
 	err := msg.ValidateBasic()
 	if err != nil {
 		return nil, err
 	}
-	// Check that the order exists in status PENDING
-	demandOrder, err := m.GetDemandOrder(ctx, commontypes.Status_PENDING, msg.OrderId)
+
+	demandOrder, err := m.GetMutableOrder(ctx, msg.OrderId)
```
omritoptix, line 35: I understand the need to wrap the logic of checking if the demand order is relevant,  but I think that `Mutable` is confusing term here as it sounds like there are demand orders, which are active, but immutable - which is not the case.

"mutable" is simply pending and not fulfilled. 

maybe "active" is a better name (as we return error of `DemandOrderInactive` also) or another name which doesn't imply specific attirbute of a demand order.

x/eibc/keeper/msg_server.go
```diff
@@ -23,25 +26,25 @@ var _ types.MsgServer = msgServer{}
...
 	}
-	// Check that the order is not fulfilled yet
-	if demandOrder.IsFulfilled {
-		return nil, types.ErrDemandAlreadyFulfilled
-	}
 
-	// Check the underlying packet is still relevant (i.e not expired, rejected, reverted)
-	if demandOrder.TrackingPacketStatus != commontypes.Status_PENDING {
-		return nil, types.ErrDemandOrderInactive
+	// Check that the demand order fee is higher than the minimum fee
```
omritoptix, line 40:
```suggestion
	// Check that the fulfiller expected fee is equal to the demand order fee
```

x/eibc/types/tx.go
```diff
@@ -59,13 +95,25 @@ func (m *MsgFulfillOrder) isValidOrderId(orderId string) bool {
...
+	_, err := sdk.AccAddressFromBech32(address)
+	if err != nil {
 		return err
 	}
-	return nil
-}
 
-func (m *MsgFulfillOrder) GetFulfillerBech32Address() []byte {
-	return sdk.MustAccAddressFromBech32(m.FulfillerAddress)
+	if fee != "" {
```
omritoptix, line 107: seems like if `fee=""` we pass validation

x/eibc/keeper/msg_server.go
```diff
@@ -66,3 +69,66 @@ func (m msgServer) FulfillOrder(goCtx context.Context, msg *types.MsgFulfillOrde
...
+	}
+
+	// calculate the new price: transferTotal - newFee - bridgingFee
+	newFeeInt, _ := sdk.NewIntFromString(msg.NewFee)
+	transferTotal, _ := sdk.NewIntFromString(data.Amount)
+	newPrice := transferTotal.Sub(newFeeInt)
+	if newPrice.IsNegative() {
+		return nil, types.ErrFeeTooHigh
+	}
+	bridgingFee := m.dack.BridgingFeeFromAmt(ctx, transferTotal)
```
omritoptix, line 111: bridging fee should only apply to OnRecv packets

omritoptix (changes requested):

x/eibc/types/tx.go
mtsitrin: we do, as we want to support zero fee orders
fee is `sdk.Coins` which when empty is "" string

x/eibc/keeper/msg_server.go
omritoptix: still didn't approve as didn't see this was addressed in the pr fixes.

x/eibc/types/demand_order.go
```diff
@@ -89,6 +98,18 @@ func (m *DemandOrder) GetRecipientBech32Address() sdk.AccAddress {
...
+	// Check that the order is not fulfilled yet
+	if m.IsFulfilled {
+		return ErrDemandAlreadyFulfilled
+	}
+	// Check the underlying packet is still relevant (i.e not expired, rejected, reverted)
+	if m.TrackingPacketStatus != commontypes.Status_PENDING {
+		return ErrDemandOrderInactive
+	}
+	return nil
+}
```
danwt, line 111: Can we have any of

```
Open order
Unexecuted order
Outstanding order
Live order
```

as terminology instead of active

Also, bit weird to return errors here

They aren't really errors, they're just statuses.

Just make an enum

x/eibc/keeper/msg_server.go
```diff
@@ -23,25 +26,25 @@ var _ types.MsgServer = msgServer{}
...
+	for _, coin := range demandOrder.Fee {
+		expectedFee, _ := sdk.NewIntFromString(msg.ExpectedFee)
+		if !coin.Amount.Equal(expectedFee) {
+			return nil, types.ErrExpectedFeeNotMet
+		}
 	}
+
 	// Check for blocked address
 	if m.bk.BlockedAddr(demandOrder.GetRecipientBech32Address()) {
 		return nil, types.ErrBlockedAddress
```
danwt, line 50: I know it's not your PR but why is this here, rather than before demand order creation?

x/eibc/keeper/msg_server.go
```diff
@@ -66,3 +69,66 @@ func (m msgServer) FulfillOrder(goCtx context.Context, msg *types.MsgFulfillOrde
...
+	demandOrder, err := m.GetActiveOrder(ctx, msg.OrderId)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check that the submitter is the expected recipient of the order
+	submitter := msg.GetSubmitterAddr()
+	if !submitter.Equals(demandOrder.GetRecipientBech32Address()) {
+		return nil, errorsmod.Wrap(sdkerrors.ErrUnauthorized, "only the recipient can update the order")
+	}
```
danwt, line 92: owner v submitter vs recipient terminology
pick one and be consistent

x/eibc/keeper/msg_server.go
```diff
@@ -66,3 +69,66 @@ func (m msgServer) FulfillOrder(goCtx context.Context, msg *types.MsgFulfillOrde
...
+	newPrice := transferTotal.Sub(newFeeInt)
+	if newPrice.IsNegative() {
+		return nil, types.ErrFeeTooHigh
+	}
+	bridgingFee := m.dack.BridgingFeeFromAmt(ctx, transferTotal)
+	newPrice = newPrice.Sub(bridgingFee)
+
+	denom := demandOrder.Price[0].Denom
+	demandOrder.Fee = sdk.NewCoins(sdk.NewCoin(denom, newFeeInt))
+	demandOrder.Price = sdk.NewCoins(sdk.NewCoin(denom, newPrice))
```
danwt, line 116: so much duplicated computation
please dedup/dry out with the other similar computations on the order creation etc

x/eibc/keeper/msg_server.go
```diff
@@ -66,3 +69,66 @@ func (m msgServer) FulfillOrder(goCtx context.Context, msg *types.MsgFulfillOrde
...
+		return nil, err
+	}
+
+	// calculate the new price: transferTotal - newFee - bridgingFee
+	newFeeInt, _ := sdk.NewIntFromString(msg.NewFee)
+	transferTotal, _ := sdk.NewIntFromString(data.Amount)
+	newPrice := transferTotal.Sub(newFeeInt)
+	if newPrice.IsNegative() {
+		return nil, types.ErrFeeTooHigh
+	}
```
danwt, line 110: isn't this in the wrong place

x/eibc/keeper/msg_server.go
```diff
@@ -66,3 +69,66 @@ func (m msgServer) FulfillOrder(goCtx context.Context, msg *types.MsgFulfillOrde
...
+	bridgingFee := m.dack.BridgingFeeFromAmt(ctx, transferTotal)
+	newPrice = newPrice.Sub(bridgingFee)
```
danwt, line 112: can go to zero/negative?

x/eibc/types/demand_order.go
```diff
@@ -32,6 +32,25 @@ func NewDemandOrder(rollappPacket commontypes.RollappPacket, price, fee math.Int
 }
 
 func (m *DemandOrder) ValidateBasic() error {
+	if len(m.Price) > 1 || len(m.Fee) > 1 {
+		return ErrMultipleDenoms
+	}
```
danwt, line 37: I don't understand it
Please comment

danwt, line 37: Why do we even have a validate basic function if the data always comes from inside the system?
it should be correct by construction

danwt (approved):
The calculations on the update function look error prone

Please dry out

I haven't though through the details but why can't there just be one pure function to handle all of the calculations needed for the whole eibc

Also please pick another word than 'active', I made some suggestions

x/eibc/types/demand_order.go
mtsitrin: will rename
It's not that wierd, as we expect to interact with only "Outstanding" orders.
maybe `ValidateLiveOrder` instead of `IsActive` is more suitable

x/eibc/keeper/msg_server.go
mtsitrin: yeah seems redundant

x/eibc/keeper/msg_server.go
mtsitrin: price cannot. fee can

x/eibc/types/demand_order.go
mtsitrin: the `DemandOrder` struct have `sdk.Coins` field
the validation checks it's not contains multiple coins

x/eibc/keeper/demand_order.go
```diff
@@ -78,30 +78,23 @@ func (k *Keeper) CreateDemandOrderOnRecv(ctx sdk.Context, fungibleTokenPacketDat
...
 	demandOrderRecipient := fungibleTokenPacketData.Receiver // who we tried to send to
 
-	order := eibctypes.NewDemandOrder(*rollappPacket, demandOrderPrice, fee, demandOrderDenom, demandOrderRecipient)
+	order := types.NewDemandOrder(*rollappPacket, demandOrderPrice, fee, demandOrderDenom, demandOrderRecipient)
 	return order, nil
 }
 
 // CreateDemandOrderOnAck creates a demand order for a timeout or errack packet.
 // The fee multiplier is read from params and used to calculate the fee.
 func (k Keeper) CreateDemandOrderOnAck(ctx sdk.Context, fungibleTokenPacketData transfertypes.FungibleTokenPacketData,
```
omritoptix, line 95: function name is misleading as in fact it's `OnErrAckOrTimeout`

x/eibc/types/demand_order.go
mtsitrin: it shouold be correct, the enforcement done on the `Validate` instead of the `NewDemandOrder`
it probably should be used in case of loading orders on genesis as well

omritoptix (approved):

x/eibc/keeper/msg_server.go
```diff
@@ -23,29 +26,25 @@ var _ types.MsgServer = msgServer{}
...
-	}
-	// Check for blocked address
-	if m.bk.BlockedAddr(demandOrder.GetRecipientBech32Address()) {
-		return nil, types.ErrBlockedAddress
+	// Check that the fulfiller expected fee is equal to the demand order fee
+	for _, coin := range demandOrder.Fee {
+		expectedFee, _ := sdk.NewIntFromString(msg.ExpectedFee)
+		if !coin.Amount.Equal(expectedFee) {
+			return nil, types.ErrExpectedFeeNotMet
+		}
```
danwt, line 45: why is demandOrder.Fee `Coins` but expectedFee is `Coin`?

x/eibc/types/fees.go
```diff
@@ -0,0 +1,14 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// calculate the new price: transferTotal - fee - bridgingFee
+func CalcPriceWithBridgingFee(amt sdk.Int, feeInt sdk.Int, bridgingFeeMultiplier sdk.Dec) (sdk.Int, error) {
+	bridgingFee := bridgingFeeMultiplier.MulInt(amt).TruncateInt()
+	price := amt.Sub(feeInt).Sub(bridgingFee)
```
danwt, line 8: do we want to protect against bridge fee being zero?

danwt (commented):
I'm a bit confused why there is Coins on the DemandOrder struct but we treat it as Coin always

Is it just a protobuf limitation?

x/eibc/types/fees.go
mtsitrin: it can happen with small amt transfers / low fee percentage

omritoptix (approved):

x/eibc/types/fees.go
danwt: is that open to abuse? People can avoid the fee by sending small transfers?
Or maybe it's unrealistic because it would take so long

Issue #916, opened 2024-06-05
add UT for `createDemandOrderFromIBCPacket`


mtsitrin:
need to add UT for the delayedAck <-> eIBC integration

PR #918, opened 2024-06-10
fix(eIBC): bridging_fee taken from original recipient and not from fufliller 

### LINKED ISSUE #917: Change ibc bridge fee to be taken from original recipient and not from fufliller when using eibc
In the case of eibc, furrently we're chraging the bridge fee from the fulfiller.
this creates multiple issues:
1. incosistency with the `fee` he should get (as it's not really the `fee` cause we deduce the `bridging fee`)
2. incosistency from the sender side in term of how much he is willing to pay to the fulfiller (because the eibc fee is deducted with bridging fee)
3. there is an implicit assumption now that the eibc = `bridging fee + tip` which is confusing 

Ideally, eibc should come as an extra on top of the bridging fee which should always be charged from the sender size, regardless of the eibc amount.

mtsitrin:
This PR moves the "bridging fee" tax from the fulfiller to the original destination address of the transfer.

So when a demand order created:
```
* Price = transfer amt - fee **- bridging fee**
* Fee  = as set in memo
```
instead of currently:
```
* Price = transfer amt - fee
* Fee  = as set in memo - bridging fee
```

So the fulfiller sends the `Price` on fulfillment, and gets
`Amt **- bridging fee**` (==`fee`)

Closes #917 

## Summary by CodeRabbit

- **New Features**
  - Introduced `EIBCDemandOrderHandler` to manage EIBC packets by creating demand orders based on packet data.
  - Added `LogString` method for `RollappPacket` to generate formatted strings of packet details.
  - Added `BlockedAddr` method to check for blocked addresses.

- **Refactor**
  - Renamed various methods and fields for clarity and consistency.
  - Updated package name from `delayedack` to `keeper`.

- **Style**
  - Improved comments for better clarity and consistency.


x/eibc/keeper/demand_order.go
```diff
@@ -86,19 +94,22 @@ func (im IBCMiddleware) createDemandOrderFromIBCPacket(ctx sdk.Context, fungible
...
 	// Get the fee from the memo
 	fee, _ := eibcMetaData.FeeInt() // guaranteed ok by above validation
-
 	if amt.LT(fee) {
 		return nil, fmt.Errorf("fee cannot be larger than amount: fee: %s: amt :%s", fee, fungibleTokenPacketData.Amount)
 	}
 
+	// Get the bridging fee from the amount
+	bridgingFee := k.dack.BridgingFeeFromAmt(ctx, amt)
+	demandOrderPrice := amt.Sub(fee).Sub(bridgingFee)
```
zale144, line 111: isn't it possible that the `demandOrderPrice` can be zero or less? I think the check on line 105: `if amt.LT(fee)` should include the `bridgingFee` as well

zale144: also, it would be nice to cover this with tests

zale144 (commented):
Nice work! left a couple of comments

x/eibc/keeper/demand_order.go
mtsitrin: added check.
UT will be handled in https://github.com/dymensionxyz/dymension/issues/916

x/bridging_fee/ibc_middleware.go
```diff
@@ -108,6 +109,9 @@ func (im *BridgingFeeMiddleware) OnRecvPacket(ctx sdk.Context, packet channeltyp
 				EventTypeBridgingFee,
 				sdk.NewAttribute(AttributeKeyFee, fee.String()),
 				sdk.NewAttribute(sdk.AttributeKeySender, transferPacketData.Sender),
+				sdk.NewAttribute(transfertypes.AttributeKeyReceiver, transferPacketData.Receiver),
```
omritoptix, line 112: AFAIU ,the receiver of this packet here is the fulfiller, which I think is misleading to put as he actually didn't pay the bridging fee.

I'd expect the receiver here to be the actual receiver who paid the fee (you can argue that the fee is paid by the sender but that's for the sender/receiver to decide) - which is the original recipient.

x/eibc/keeper/demand_order.go
```diff
@@ -0,0 +1,186 @@
...
+	order is created with (price,fee)
+	the order creator immediately receives price on fulfillment from fulfiller
+	when the ack/timeout finalizes, the fulfiller receives the transfer_amount
+
+	therefore:
+	fulfiller balance += (transfer_amount - (transfer_amount-fee))
+	equivalent to += fee_multiplier*transfer_amount
+	demander balance += (transfer_amount - fee)
+	equivalent to += (1-fee_multiplier)*transfer_amount
+*/
```
danwt, line 146: invalid docstring format

x/eibc/keeper/demand_order.go
```diff
@@ -0,0 +1,186 @@
...
+func (k Keeper) PrepareProtocolDemandOrder(ctx sdk.Context, fungibleTokenPacketData transfertypes.FungibleTokenPacketData,
+	rollappPacket *commontypes.RollappPacket,
+) (*eibctypes.DemandOrder, error) {
+	// Validate the fungible token packet data as we're going to use it to create the demand order
+	if err := fungibleTokenPacketData.ValidateBasic(); err != nil {
+		return nil, err
```
danwt, line 152: its already validated right?

x/eibc/keeper/demand_order.go
```diff
@@ -0,0 +1,186 @@
...
+
+	trace := transfertypes.ParseDenomTrace(fungibleTokenPacketData.Denom)
+	demandOrderDenom := trace.IBCDenom()
+	demandOrderRecipient := fungibleTokenPacketData.Sender // and who tried to send it (refund because it failed)
+
+	order := eibctypes.NewDemandOrder(*rollappPacket, demandOrderPrice, fee, demandOrderDenom, demandOrderRecipient)
+	if err := order.Validate(); err != nil {
+		return nil, fmt.Errorf("validate eibc data: %w", err)
+	}
+	return order, nil
```
danwt, line 185: duplicated, could be dried out

danwt (changes requested):
Please rename and give a short docstring to PrepareProtocolDemandOrder
And dry out the fungibleTokenPacketData vaildation and blocked addr, and 
`eibctypes.NewDemandOrder` bits

other than that nice work

danwt:
I would say though it's getting very hard and confusing to follow for me

We should refactor at some point to just have 1 very clear calculation in the code somehwere

```
	// fulfiller balance += transferAmt - bridgeFee - (transferAmt - bridgeFee - fee)
	// receiver balance += transferAmt - bridgeFee - fee
	// bridge balance += fee
	```

x/eibc/keeper/demand_order.go
mtsitrin: dried out some
but the creation I prefer to do in the inner func, otherwise I need to return all args (`demandOrderPrice, fee, demandOrderDenom, demandOrderRecipient`)

omritoptix (approved):

danwt (approved):
gj

PR #950, opened 2024-06-17
feat(eibc): add extra info to demand order

### LINKED ISSUE #873: Add extra information to eibc order events
The eibc demand orders could use some extra information, such as the fulfiller address, block height when it was fulfilled and so on...
The reasoning is that these will be picked up by the indexer and made available for querying, which will come useful when tracking fulfilled demand orders, in particular for the funds tracking of the eibc-client.

zale144:
## Description

----

Closes #873 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:

## Summary by CodeRabbit

- **New Features**
  - Added a new field `FulfillerAddress` to `DemandOrder`.
  - `FulfillOrder` function now sets the `FulfillerAddress` field.
  - Events related to `DemandOrder` now include the `FulfillerAddress` attribute if present.


proto/dymension/eibc/demand_order.proto
```diff
@@ -30,4 +30,6 @@ message DemandOrder {
     dymensionxyz.dymension.common.Status tracking_packet_status = 8;
     string rollapp_id = 9;
     common.RollappPacket.Type type = 10;
+  // fulfiller_address is the bech32-encoded address of the account which the message was sent from.
```
danwt, line 33: what message?

x/eibc/keeper/keeper.go
```diff
@@ -108,6 +108,7 @@ func (k *Keeper) UpdateDemandOrderWithStatus(ctx sdk.Context, demandOrder *types
 // FulfillOrder should be called only at most once per order.
 func (k Keeper) FulfillOrder(ctx sdk.Context, order *types.DemandOrder, fulfillerAddress sdk.AccAddress) error {
 	order.IsFulfilled = true
+	order.FulfillerAddress = fulfillerAddress.String()
```
danwt, line 111: seems to me we could deprecate isFulfilled and replace it with a method that does FulfillerAdddress!=""?

danwt (commented):
https://github.com/dymensionxyz/dymension/issues/873#issuecomment-2173660818

proto/dymension/eibc/demand_order.proto
```diff
@@ -26,8 +26,9 @@ message DemandOrder {
       (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coins"
     ];
     string recipient = 5;
-    bool is_fulfilled = 6;
     dymensionxyz.dymension.common.Status tracking_packet_status = 8;
```
danwt, line 29: in general please dont delete proto fields
need to reserve 6

x/eibc/client/cli/query_command_orders.go
```diff
@@ -68,11 +93,62 @@ func CmdListDemandOrdersByStatus() *cobra.Command {
...
+  rollapp_id:	%s
+  status:	%s
+  packet_key:	%s
+  fulfiller:	%s
+`,
+					o.Id, o.Recipient, parseAndFormat(o.Price.String()), parseAndFormat(o.Fee.String()), o.RollappId,
+					o.TrackingPacketStatus.String(), strings.TrimSpace(o.TrackingPacketKey), o.FulfillerAddress)
+			}
+
+			fmt.Printf("\ncount: %d; ts: %s\n", count, time.Now().String())
```
danwt, line 120: It's still possible to use `-o json` right? to get json output?

danwt: also: len(res.DEmandOrders) instead of the count?

x/eibc/client/cli/query_command_orders.go
```diff
@@ -68,11 +93,62 @@ func CmdListDemandOrdersByStatus() *cobra.Command {
...
+			return nil
 		},
 	}
 
+	cmd.Flags().StringP("rollapp", "r", "", "Rollapp ID")
+	cmd.Flags().StringP("recipient", "c", "", "Recipient address")
+	cmd.Flags().StringP("type", "t", "", "Packet type")
+	cmd.Flags().StringP("denom", "d", "", "Denom")
+	cmd.Flags().StringP("fulfilled", "f", "", "Filter by fulfillment status")
+	cmd.Flags().StringP("fulfiller", "a", "", "Filter by fulfiller address")
```
danwt, line 131: I feel like the shorthands are pretty non obvious for recipient, fulfiller, fulfilled

x/eibc/client/cli/query_command_orders.go
```diff
@@ -68,11 +93,62 @@ func CmdListDemandOrdersByStatus() *cobra.Command {
...
+		return coin.String()
+	}
+	return formatAmount(coin.Amount.String()) + coin.Denom
+}
+
+func formatAmount(numStr string) string {
+	if len(numStr) <= 18 {
+		return "0," + strings.Repeat("0", 18-len(numStr)) + numStr
+	}
+	return numStr[:len(numStr)-18] + "," + numStr[len(numStr)-18:]
```
danwt, line 153: what is this doing?

x/eibc/client/cli/query_command_orders.go
```diff
@@ -68,11 +93,62 @@ func CmdListDemandOrdersByStatus() *cobra.Command {
...
+func parseAndFormat(amountStr string) string {
+	coin, err := sdk.ParseCoinNormalized(amountStr)
+	if err != nil {
+		panic(err)
+	}
+	if coin.Denom != "adym" {
+		return coin.String()
+	}
+	return formatAmount(coin.Amount.String()) + coin.Denom
+}
```
danwt, line 147: just a tip
another way would be like

```
type Adym struct {
	sdk.Coin
}

func (a Adym) String() string {
	if a.Denom != "adym"{
		return a.Coin.String()
	}
	coin, err := sdk.ParseCoinNormalized(amountStr)
	if err != nil {
		panic(err)
	}
	if len(numStr) <= 18 {
		return "0," + strings.Repeat("0", 18-len(numStr)) + numStr
	}
	return numStr[:len(numStr)-18] + "," + numStr[len(numStr)-18:]
}
```

fyi

You dont have to use it

danwt (changes requested):

x/eibc/client/cli/query_command_orders.go
zale144: what would you propose then? users are free to not use them

x/eibc/client/cli/query_command_orders.go
zale144: the idea was to handle adym and separate the decimal part. This was originally for my own convenience, but I am not sure if everyone will want this

x/eibc/client/cli/query_command_orders.go
zale144: sure

x/eibc/client/cli/query_command_orders.go
zale144: yes, `-o json` works still

mtsitrin (approved):

x/eibc/client/cli/query_command_orders.go
danwt: I guess it's fine but maybe just dont' even given options for short ones?
Because we might add another arg later which is a better fit for a short name, but it's already taken. That would lead to either a confusing backwards compatible api, or a non backwards compatible one

x/eibc/client/cli/query_command_orders.go
danwt: ok but it should have a doc string so readers dont have to reverse engineer it

danwt: anyways can you check if there is a simpler way or existing solution for this?
this kind of manual code looks sus
https://stackoverflow.com/a/51546906

danwt: and it seems like a generally good fit for a 'denom utility package' in https://github.com/dymensionxyz/sdk-utils

danwt (changes requested):

danwt (approved):

mtsitrin (approved):

PR #966, opened 2024-06-21
fix(delayedack): handle finalization errors per rollapp

### LINKED ISSUE #670: Finalization queue logic should be done per rollapp
Currently if one rollapp finalization fail we fail the entire rollapps finalization.
We should fail it on a per rollapp basis.

zale144:
## Description

----

Closes #670 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,62 +2,81 @@ package keeper
...
+		}
+		// remove the blockHeightToFinalizationQueue if all the rollapps are finalized
+		if len(rollappsToRemoveFromQueue) == len(blockHeightToFinalizationQueue.FinalizationQueue) {
+			k.RemoveBlockHeightToFinalizationQueue(ctx, blockHeightToFinalizationQueue.CreationHeight)
+		} else {
+			// remove only the rollapps that were successfully finalized
+			blockHeightToFinalizationQueue.FinalizationQueue = slices.DeleteFunc(blockHeightToFinalizationQueue.FinalizationQueue,
+				func(si types.StateInfoIndex) bool {
+					return slices.Contains(rollappsToRemoveFromQueue, si.RollappId)
+				})
```
danwt, line 51: this is an n^2 algorithm right
I think it's fine now because we don't have many rollapps
but IMO you need to add a warning/todo here to fix it

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,62 +2,81 @@ package keeper
...
-			keeperHooks := k.GetHooks()
-			err := keeperHooks.AfterStateFinalized(ctx, stateInfoIndex.RollappId, &stateInfo)
+			err := osmoutils.ApplyFuncIfNoError(ctx,
+				func(ctx sdk.Context) error {
+					return k.finalizePendingState(ctx, stateInfoIndex)
+				})
 			if err != nil {
-				return fmt.Errorf("after state finalized: %w", err)
+				k.Logger(ctx).Error("failed to finalize state", "rollapp", stateInfoIndex.RollappId, "index", stateInfoIndex.Index, "error", err.Error())
+				continue
```
danwt, line 39: Can you add a comment, either here, or maybe as the function docstring, which explains the high level philosophy here about how errors are handled? Like the full lifecycle.

For example, what happens if the rollapp doesnt finalize over and over?
Needs some explanation and guidance, otherwise I'm reading the code and it just seems super error prone and I'm not sure what the assumptions and guarantees are

danwt: Basically, like, even though the code may be correct, it's not obvious that it is. There is some product assumption here

x/rollapp/keeper/block_height_to_finalization_queue_test.go
```diff
@@ -129,6 +304,16 @@ func createNBlockHeightToFinalizationQueue(keeper *keeper.Keeper, ctx sdk.Contex
...
 
+func countFinalized(response abci.ResponseEndBlock) int {
+	count := 0
+	for _, event := range response.Events {
+		if event.Type == types.EventTypeStatusChange {
+			count++
+		}
+	}
+	return count
+}
```
danwt, line 315: could also use len(slices.DeleteFunc(..))

danwt (changes requested):
Looks correct to me but I think it could do with some better documentation as this is a very critical part of the code

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,62 +2,84 @@ package keeper
...
+			err := osmoutils.ApplyFuncIfNoError(ctx,
+				func(ctx sdk.Context) error {
+					return k.finalizePendingState(ctx, stateInfoIndex)
+				})
 			if err != nil {
-				return fmt.Errorf("after state finalized: %w", err)
+				// if a rollapp's state fails to finalize, move on (state index will stay in the queue)
+				// TODO: think about (non)recoverable errors and how to handle them accordingly
+				k.Logger(ctx).Error("failed to finalize state", "rollapp", stateInfoIndex.RollappId, "index", stateInfoIndex.Index, "error", err.Error())
+				continue
```
mtsitrin, line 42: can it cause invariant break?
if some states of a rollapp are finalized but some don't

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,62 +2,84 @@ package keeper
...
 			}
-
-			// emit event
-			ctx.EventManager().EmitEvent(
-				sdk.NewEvent(types.EventTypeStatusChange,
-					stateInfo.GetEvents()...,
-				),
-			)
+			rollappsToRemoveFromQueue = append(rollappsToRemoveFromQueue, stateInfoIndex.RollappId)
```
mtsitrin, line 44: u assume single state per rollapp in the queue?

x/rollapp/keeper/keeper.go
```diff
@@ -21,22 +21,36 @@ type Keeper struct {
 
 	ibcClientKeeper types.IBCClientKeeper
 	channelKeeper   types.ChannelKeeper
+
+	finalizePending func(ctx sdk.Context, stateInfoIndex types.StateInfoIndex) error
```
mtsitrin, line 25: why we need this as an interface?

mtsitrin (approved):

x/rollapp/keeper/keeper.go
zale144: To be able to mock it easily, for returning errors on certain rollapp state indexes in tests

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,65 +2,114 @@ package keeper
...
+	})
+	if err != nil {
+		// TODO: think about (non)recoverable errors and how to handle them accordingly
+		k.Logger(ctx).Error(
+			"failed to finalize state",
+			"rollapp", stateInfoIndex.RollappId,
+			"index", stateInfoIndex.Index,
+			"error", err.Error(),
+		)
+		return fmt.Errorf("finalize state: %w", err)
```
danwt, line 65: you shouldn't really log and return an error, generally

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,65 +2,114 @@ package keeper
...
+}
+
+func (k Keeper) updateQueueForHeight(ctx sdk.Context, blockHeightToFinalizationQueue types.BlockHeightToFinalizationQueue, failedRollapps map[string]uint64) {
+	// remove the blockHeightToFinalizationQueue if all the rollapps' states are finalized
+	if len(failedRollapps) == 0 {
+		k.RemoveBlockHeightToFinalizationQueue(ctx, blockHeightToFinalizationQueue.CreationHeight)
+		return
+	}
+	// remove from the queue only the rollapps that were successfully finalized at all indices.
+	// for failed rollapps at current height, all state indices including and following the failed index will stay in the queue
```
danwt, line 103:
```suggestion
	// remove from the queue the rollapps that were successfully finalized at all indices.
	// for failed rollapps at current height, all state indices including and following the failed index will stay in the queue
```

danwt (changes requested):

x/rollapp/keeper/block_height_to_finalization_queue.go
zale144: Is this really a blocker?

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,65 +2,114 @@ package keeper
...
-					stateInfo.GetEvents()...,
-				),
-			)
+		k.finalizeQueueForHeight(ctx, blockHeightToFinalizationQueue, failedRollapps)
+	}
+}
+
+func (k Keeper) finalizeQueueForHeight(ctx sdk.Context, queue types.BlockHeightToFinalizationQueue, failedRollapps map[string]uint64) {
+	// finalize pending states
+	for _, stateInfoIndex := range queue.FinalizationQueue {
```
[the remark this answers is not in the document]
zale144: the iteration is the same as it was, the queue is retrieved the same way as previously

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,65 +2,114 @@ package keeper
...
+	if len(failedRollapps) == 0 {
+		k.RemoveBlockHeightToFinalizationQueue(ctx, blockHeightToFinalizationQueue.CreationHeight)
+		return
+	}
+	// remove from the queue only the rollapps that were successfully finalized at all indices.
+	// for failed rollapps at current height, all state indices including and following the failed index will stay in the queue
+	blockHeightToFinalizationQueue.FinalizationQueue = slices.DeleteFunc(blockHeightToFinalizationQueue.FinalizationQueue,
+		func(si types.StateInfoIndex) bool {
+			idx, failed := failedRollapps[si.RollappId]
+			return !failed || si.Index < idx
```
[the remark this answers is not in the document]
zale144: it means, while iterating the queue for deleting the successfully finalized states, we remove them if
- it was not found in the `failedRollapps` map
- if it was found, the indexes to be deleted should only be the ones up to the point of the index of the first failed state change

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -2,65 +2,114 @@ package keeper
...
-				return fmt.Errorf("after state finalized: %w", err)
-			}
-
-			// emit event
-			ctx.EventManager().EmitEvent(
-				sdk.NewEvent(types.EventTypeStatusChange,
-					stateInfo.GetEvents()...,
-				),
-			)
+		k.finalizeQueueForHeight(ctx, blockHeightToFinalizationQueue, failedRollapps)
```
omritoptix, line 32: seems to me like the you can have a state index of e.g rollapp abc being finalized after  previous state index of rollapp abc failed in previous queue. e.g :

1. queue1 = Queue{height: 10, stateIndexes: [{index:1, rollapp: abc}]}
2. queue2 = Queue{height: 11, stateIndexes: [{index:2, rollapp: abc}]} 

In that case, if e.g queue1 fails, it will not finalize the state index, but it seems like queue2 can be independently finalized (As we're trying to finalize each rollapp state index queue without consideration of previous rollapp queue results) vs failing all subsequent state index for that rollapp in the following queues.

_not sure if there is some other place in the code which will fail index2, due to index1 not being finalized, but I don't see it explicitly in this code._

Also I'd suggest a simple validation of only finalizing a state index, if the previous is also finalized (which can avoid all this cases) as currently the finalize function is the following:

```
func (s *StateInfo) Finalize() {
	s.Status = common.Status_FINALIZED
}
```

x/rollapp/keeper/block_height_to_finalization_queue.go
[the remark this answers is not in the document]
omritoptix: ok. still not sure what's the answer though.

x/rollapp/keeper/block_height_to_finalization_queue.go
zale144: [here ](https://github.com/dymensionxyz/dymension/blob/b159bbd3c001b297f5b8f7d5ba6bc02f8cb308be/x/rollapp/keeper/block_height_to_finalization_queue.go#L39) you can see that if the rollapp had previously failed, its state finalization will be skipped

[here](https://github.com/dymensionxyz/dymension/blob/b159bbd3c001b297f5b8f7d5ba6bc02f8cb308be/x/rollapp/keeper/block_height_to_finalization_queue.go#L46) the failure is recorded and will be kept for the duration of the current `EndBlocker` execution. 

this will result in the failed state changes to not be deleted from the queue and iterated over next iteration of `EndBlocker`.

x/rollapp/keeper/block_height_to_finalization_queue.go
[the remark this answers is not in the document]
omritoptix: > if it was found, the indexes to be deleted should only be the ones up to the point of the index of the first failed state change

Ok. I think it should probably be clearer in the func description as it wasn't that clear for me:
```
for failed rollapps at current height, all state indices including and following the failed index will stay in the queue
```

also in the description of the `failedRollapps` is missing with this detail:

```
// cache the rollapps that failed to finalize at current EndBlocker execution
	failedRollapps := make(map[string]uint64)
```

In general lets try to be a bit more explicit.

x/rollapp/keeper/block_height_to_finalization_queue.go
omritoptix: right. missed that. my bad.

x/rollapp/keeper/block_height_to_finalization_queue.go
danwt: I mean it's wrong and makes it harder to understand tbh

danwt: you can see it confused omri as well
https://github.com/dymensionxyz/dymension/pull/966#discussion_r1680629576

omritoptix (dismissed):

mtsitrin (dismissed):

danwt (approved):
great work

mtsitrin (approved):

omritoptix (approved):

PR #1006, opened 2024-07-23
feat(sequencers): sequencer rotation support

### LINKED ISSUE #993: sequencer rotation support
- [x] manage activeSequencer and nextSequencer for a rollapp
- [x] manage notice_period for unbonding sequencers
- [x] emit rotate_sequencer event
- [x] expect and handle “lastBatch”
- [x] complete the sequencer rotation
- [x] #1025

phase 2:
* enforce state_update_timeout (#927  )
* enforce final_state_update_timeout (#1085 )

mtsitrin:
Closes #993 

will be done separately:

- [x] https://github.com/dymensionxyz/dymension/issues/1025
- [x] #1026

## Summary by CodeRabbit

- **New Features**
	- Enhanced functionality for querying current and next proposers of rollapps.
	- Introduced parameters for unbonding time and notice period for sequencers in genesis configuration.
	- Added new methods for managing sequencer rotations and notice periods.

- **Bug Fixes**
	- Improved error handling and clarity for sequencer registration and unbonding processes.

- **Documentation**
	- Enhanced comments across various files for better understanding of code logic and structure.

- **Style**
	- Refactored code for improved readability and consistency in method naming and logic flow.

- **Tests**
	- Expanded test coverage for sequencer management and proposer rotations.
	- Refactored tests to improve clarity and maintainability.


x/sequencer/keeper/hook_listener.go
```diff
@@ -58,12 +70,13 @@ func (hook rollappHook) FraudSubmitted(ctx sdk.Context, rollappID string, height
 	// unbond all other bonded sequencers
 	sequencers := hook.k.GetSequencersByRollappByStatus(ctx, rollappID, types.Bonded)
 	for _, sequencer := range sequencers {
-		err := hook.k.forceUnbondSequencer(ctx, sequencer.SequencerAddress)
-		if err != nil {
-			return err
-		}
+		hook.k.startUnbondingPeriodForSequencer(ctx, &sequencer)
```
mtsitrin, line 73: on fraud, i've changed the bonded seqs to start `unbonding` instead of fully unbonded

x/sequencer/keeper/sequencer.go
```diff
@@ -171,3 +135,128 @@ func (k Keeper) removeUnbondingSequencer(ctx sdk.Context, sequencer types.Sequen
 	unbondingQueueKey := types.UnbondingSequencerKey(sequencer.SequencerAddress, sequencer.UnbondTime)
 	store.Delete(unbondingQueueKey)
 }
+
+/* -------------------------------------------------------------------------- */
+/*                                notice period                               */
+/* -------------------------------------------------------------------------- */
+// get finished notice period sequencers
+func (k Keeper) GetMatureNoticePeriodSequencers(ctx sdk.Context, endTime time.Time) (list []types.Sequencer) {
```
danwt, line 143: wrong docsring

x/sequencer/keeper/sequencer.go
```diff
@@ -171,3 +135,128 @@ func (k Keeper) removeUnbondingSequencer(ctx sdk.Context, sequencer types.Sequen
...
+// remove sequencer from notice period queue
+func (k Keeper) removeNoticePeriodSequencer(ctx sdk.Context, sequencer types.Sequencer) {
+	store := ctx.KVStore(k.storeKey)
+	noticePeriodKey := types.NoticePeriodSequencerKey(sequencer.SequencerAddress, sequencer.UnbondTime)
+	store.Delete(noticePeriodKey)
+}
+
+/* ------------------------- proposer/next proposer ------------------------- */
+// GtAllProposers returns all proposers for all rollapps
+func (k Keeper) GetAllProposers(ctx sdk.Context) (list []types.Sequencer) {
```
danwt, line 176: wrong docstring

x/sequencer/keeper/sequencer.go
```diff
@@ -171,3 +135,128 @@ func (k Keeper) removeUnbondingSequencer(ctx sdk.Context, sequencer types.Sequen
...
+		var val types.Sequencer
+		k.cdc.MustUnmarshal(iterator.Value(), &val)
+		list = append(list, val)
+	}
+
+	return
+}
+
+// set sequencer to notice period
+func (k Keeper) SetNoticePeriodQueue(ctx sdk.Context, sequencer types.Sequencer) {
```
danwt, line 159: wrong docstring

x/sequencer/keeper/sequencer.go
```diff
@@ -171,3 +135,128 @@ func (k Keeper) removeUnbondingSequencer(ctx sdk.Context, sequencer types.Sequen
...
+	store := ctx.KVStore(k.storeKey)
+	noticePeriodKey := types.NoticePeriodSequencerKey(sequencer.SequencerAddress, sequencer.UnbondTime)
+	store.Delete(noticePeriodKey)
+}
+
+/* ------------------------- proposer/next proposer ------------------------- */
+// GtAllProposers returns all proposers for all rollapps
+func (k Keeper) GetAllProposers(ctx sdk.Context) (list []types.Sequencer) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.ProposerByRollappKey(""))
+	iterator := sdk.KVStorePrefixIterator(store, []byte{})
```
danwt, line 178: it's more readable to use ProposerKeyPrefix as the key here

x/sequencer/keeper/unbond.go
```diff
@@ -9,6 +9,29 @@ import (
 	"github.com/osmosis-labs/osmosis/v15/osmoutils"
 )
 
+// startUnbondingPeriodForSequencer sets the sequencer to unbonding status
+// can be called after notice period or directly if notice period is not required
+// caller is responsible for updating the proposer for the rollapp if neeeded
+func (k Keeper) startUnbondingPeriodForSequencer(ctx sdk.Context, seq *types.Sequencer) time.Time {
+	completionTime := ctx.BlockHeader().Time.Add(k.UnbondingTime(ctx))
```
danwt, line 16: BlockTime()

x/sequencer/keeper/sequencer.go
```diff
@@ -157,7 +121,7 @@ func (k Keeper) GetMatureUnbondingSequencers(ctx sdk.Context, endTime time.Time)
 	return
 }
 
-func (k Keeper) setUnbondingSequencerQueue(ctx sdk.Context, sequencer types.Sequencer) {
+func (k Keeper) SetUnbondingSequencerQueue(ctx sdk.Context, sequencer types.Sequencer) {
 	store := ctx.KVStore(k.storeKey)
 	b := k.cdc.MustMarshal(&sequencer)
```
danwt, line 127: weird function name because it adds the sequencer to the queue, not set the queue, as such

x/sequencer/keeper/unbond_test.go
```diff
@@ -15,16 +15,19 @@ func (suite *SequencerTestSuite) TestUnbondingMultiple() {
...
-	numOfSequencers := 5
+	numOfSequencers := 4
 	numOfSequencers2 := 3
 	unbodingSeq := 2
 
 	seqAddr1 := make([]string, numOfSequencers)
 	seqAddr2 := make([]string, numOfSequencers2)
 
-	// create 5 sequencers for rollapp1
+	// create 4 sequencers for rollapp1
```
danwt, line 30:
```suggestion
	// create sequencers for rollapp1
```

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+	"sort"
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dymensionxyz/dymension/v3/x/sequencer/types"
+)
+
+func (k Keeper) startNoticePeriodForSequencer(ctx sdk.Context, seq *types.Sequencer) time.Time {
+	completionTime := ctx.BlockHeader().Time.Add(k.NoticePeriod(ctx))
+	seq.UnbondTime = completionTime
```
danwt, line 13: Unbond time seems overloaded because this is not unbond time but actually the time of finishing the notice period

x/sequencer/types/sequencer.go
```diff
@@ -8,12 +8,22 @@ import (
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 )
 
+// validateBasic
+func (seq Sequencer) ValidateBasic() error {
```
danwt, line 12: docstring

x/sequencer/types/sequencer.go
```diff
@@ -8,12 +8,22 @@ import (
...
+}
+
 func (seq Sequencer) IsBonded() bool {
 	return seq.Status == Bonded
 }
 
-func (seq Sequencer) IsProposer() bool {
-	return seq.Proposer
+// is notice period in progress
+func (seq Sequencer) IsNoticePeriodInProgress() bool {
```
danwt, line 25: dosctring

danwt (changes requested):
Looks good, got a few comments


questions
    1. is change unbonding_height -> unbond_request_height migration safe?
    2. the hook is before, could it be cleaner after?
    3. nextProposer, what if he is not available by the time the first sequencer finishes unbonding (he can finish notice?). Phase 2 I guess?
    4. unbondTime field on sequencer is overloaded
    5. do you automatically start unbonding period when note period finishes?
    6. you call startRotation even when nextProposer finishes notice period
    7. nextProposer notice period needs to be extended to maximum duration when he becomes active proposer
    8. when is UnbondRequestHeight reset to 0?

proto/dymensionxyz/dymension/sequencer/query.proto
```diff
@@ -85,4 +91,22 @@ message QueryGetSequencersByRollappByStatusResponse {
...
-// this line is used by starport scaffolding # 3
+// Request type for the GetProposerByRollapp RPC method.
+message QueryGetProposerByRollappRequest {
+    string rollappId = 1;
+}
+
+// Response type for the GetProposerByRollapp RPC method.
+message QueryGetProposerByRollappResponse {
+    string proposer = 1;
+}
```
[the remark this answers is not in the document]
mtsitrin: yes. renamed

proto/dymensionxyz/dymension/sequencer/query.proto
```diff
@@ -85,4 +91,22 @@ message QueryGetSequencersByRollappByStatusResponse {
...
+
+// Request type for the GetNextProposerByRollapp RPC method.
+message QueryGetNextProposerByRollappRequest {
+    string rollappId = 1;
+}
+
+// Response type for the GetNextProposerByRollapp RPC method.
+message QueryGetNextProposerByRollappResponse {
+    string nextProposer = 1;
+}
```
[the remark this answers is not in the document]
mtsitrin: renamed

utils/keys.go
```diff
@@ -0,0 +1,19 @@
...
+	timeBz := sdk.FormatTimeBytes(endTime)
+	prefixL := len(queueKey)
+	bz := make([]byte, prefixL+len(timeBz))
+
+	// copy the prefix
+	copy(bz[:prefixL], queueKey)
+	// copy the encoded time bytes
+	copy(bz[prefixL:prefixL+len(timeBz)], timeBz)
+	return bz
+}
```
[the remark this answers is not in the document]
mtsitrin: i'll open a separate issue for it.
BTW, `x/streamer` has similar `getTimeKey` but with bit different implementation.
It adds the `time bytes length` to the key as well.
wdyt about it

x/sequencer/keeper/hook_listener.go
```diff
@@ -37,9 +37,21 @@ func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rolla
...
+
+	if lastStateOfSequencer {
+		if !hook.k.IsRotating(ctx, rollappId) {
+			return types.ErrInvalidRequest
+		}
+		// TODO: the hub should probably validate the lastBlock in the lastBatch,
+		// to make sure the sequencer is passing the correct nextSequencer on the L2
+
+		hook.k.RotateProposer(ctx, rollappId)
+	}
```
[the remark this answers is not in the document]
mtsitrin: not sure what will be broken. if the state update fails, the rotation won't complete

x/sequencer/keeper/msg_server_create_sequencer.go
```diff
@@ -85,9 +84,10 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
...
 		return nil, types.ErrMaxSequencersLimit
 	}
-	// this is the first sequencer, make it a PROPOSER
-	if len(bondedSequencers) == 0 {
-		sequencer.Proposer = true
+	// if this is the first sequencer, make it a PROPOSER
+	proposer := len(bondedSequencers) == 0
+	if proposer {
+		k.SetProposer(ctx, sequencer.RollappId, sequencer.SequencerAddress)
 	}
```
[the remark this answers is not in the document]
mtsitrin: this logic will be added later on, and then this snippet will be removed/
added TODO

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+
+	k.UpdateSequencer(ctx, *seq) // only bonded sequencers can have notice period
+	k.SetNoticePeriodQueue(ctx, *seq)
+
+	nextSeq := k.ExpectedNextProposer(ctx, seq.RollappId)
+	if nextSeq.SequencerAddress == "" {
+		k.Logger(ctx).Info("rollapp will be left with no proposer after notice period", "rollappId", seq.RollappId, "sequencer", seq.SequencerAddress)
+	}
```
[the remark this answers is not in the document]
mtsitrin: changed it to debug log.

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+	for _, seq := range seqs {
+		k.StartRotation(ctx, seq.RollappId)
+		k.removeNoticePeriodSequencer(ctx, seq)
+	}
+}
+
+// IsRotating returns true if the rollapp is currently rotating proposers
+func (k Keeper) IsRotating(ctx sdk.Context, rollappId string) bool {
+	return k.HasNextProposer(ctx, rollappId)
+}
```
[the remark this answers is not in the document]
mtsitrin: empty `nextProposer` will return `k.HasNextProposer() = true`
not sure what u found buggy. pls elaborate

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+
+// IsNoticePeriodRequired returns true if the sequencer requires a notice period before unbonding
+func (k Keeper) IsNoticePeriodRequired(ctx sdk.Context, seq types.Sequencer) bool {
+	return k.IsProposer(ctx, seq.RollappId, seq.SequencerAddress) || k.IsNextProposer(ctx, seq.RollappId, seq.SequencerAddress)
+}
```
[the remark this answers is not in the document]
mtsitrin: it does make sense,
once nextProposer is set on the hub, we need to assume it started to produce blocks on the L2 (even if the rotation didn't complete on the hub).
So it needs notice period as well.
https://www.notion.so/dymension/ADR-x-Sequencer-Rotation-014f9f8c781543b5b99f4f2571969c3d?pvs=4#7ec654a71e3c400b8519f24e98c63a49

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// filter out proposer and nextProposer
+	active, _ := k.GetProposer(ctx, rollappId)
+	next, _ := k.GetNextProposer(ctx, rollappId)
+	for _, s := range seqs {
+		if s.SequencerAddress == active.SequencerAddress || s.SequencerAddress == next.SequencerAddress {
+			continue
+		}
+		return s
+	}
```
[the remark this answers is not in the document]
mtsitrin: it was to avoid a case where
* seqA finishes notice period,
* nextSeq = seqB
and then seqB starts it's notice period (before rotation completed)

the idea was that if nextSeq is set, it means the rotation is in progress.
calling `ExpectedNextProposer` should give the **next** proposer, after the current rotation in progress

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+	proposer, ok := k.GetProposer(ctx, rollappId)
+	if ok {
+		k.startUnbondingPeriodForSequencer(ctx, &proposer)
+	}
+
+	nextProposer, _ := k.GetNextProposer(ctx, rollappId) // nextProposer is guaranteed to exist by caller
+	k.RemoveNextProposer(ctx, rollappId)
+
+	if nextProposer.SequencerAddress == "" {
+		k.Logger(ctx).Info("rollapp left with no proposer", "rollappId", rollappId)
```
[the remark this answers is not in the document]
mtsitrin: nil is where `nextProposer` is not set at all (no rotation in progress)
empty is where the nextProposer is nobody, and the rollapp will halt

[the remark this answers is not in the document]
mtsitrin: added godoc

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// MatureSequencersWithNoticePeriod moves all the sequencers that have finished their notice period
+func (k Keeper) MatureSequencersWithNoticePeriod(ctx sdk.Context, currTime time.Time) {
+	seqs := k.GetMatureNoticePeriodSequencers(ctx, currTime)
+	for _, seq := range seqs {
+		k.StartRotation(ctx, seq.RollappId)
+		k.removeNoticePeriodSequencer(ctx, seq)
+	}
+}
```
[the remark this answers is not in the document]
mtsitrin: the nextProposer will be set to the next sequencer in line (after the current proposer and current NextProposer)

x/sequencer/keeper/sequencer.go
mtsitrin: wanted to add the `KeySeparator` to the prefix as well

x/sequencer/keeper/rotation.go
mtsitrin: u right but I didn't want to add new field.
If seq bonded: UnbondTime is notice period completion time
if seq unbonding: UnbondTime is unbonding completion time
want to suggest better name?

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
danwt: that's super confusing IMO why do we need to differentiate the case
we can just have
nil or absent : rotation not in progress or it is and chain will halt
present: rotation in progress and chain will not halt

no need for zero value case

x/sequencer/keeper/rotation.go
danwt: I'd 100% just add a new field tbh
if you really want one then 'transitionTime' or just 'noticeOrUnbondTime'

x/sequencer/keeper/unbond_test.go
danwt: resolved?

x/sequencer/keeper/unbond.go
danwt: resolved?

x/sequencer/keeper/sequencer.go
danwt: ok ok, we can use this pattern

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
danwt: ok

x/sequencer/keeper/hook_listener.go
```diff
@@ -37,9 +37,21 @@ func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rolla
 		return types.ErrInvalidSequencerStatus
 	}
 
-	if !sequencer.Proposer {
+	seq, _ := hook.k.GetProposer(ctx, rollappId)
```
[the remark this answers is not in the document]
mtsitrin: there might be cases where proposer is not set, and i don't want to panic
added err check

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+
+// IsNoticePeriodRequired returns true if the sequencer requires a notice period before unbonding
+func (k Keeper) IsNoticePeriodRequired(ctx sdk.Context, seq types.Sequencer) bool {
+	return k.IsProposer(ctx, seq.RollappId, seq.SequencerAddress) || k.IsNextProposer(ctx, seq.RollappId, seq.SequencerAddress)
+}
+
+// ExpectedNextProposer returns the next proposer for a rollapp
+func (k Keeper) ExpectedNextProposer(ctx sdk.Context, rollappId string) types.Sequencer {
+	seqs := k.GetSequencersByRollappByStatus(ctx, rollappId, types.Bonded)
+	if len(seqs) == 0 {
```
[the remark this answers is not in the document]
mtsitrin: prefer it this way, not changing it now, it's commonly used

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
mtsitrin: need to seperate `rotation not in progress or it is and chain will halt`

mtsitrin:
> 1. is change unbonding_height -> unbond_request_height migration safe?

yes

> 2. the hook is before, could it be cleaner after?

didn't want to refactor existing logic if unnecessary

> 3. nextProposer, what if he is not available by the time the first sequencer finishes unbonding (he can finish notice?). Phase 2 I guess?

on the end of notice period, the nextProposer is set.
if no nextProposer available, the rollapp will remain "proposerless"
on beginBlock we check if there's new bonded sequencer, if so set it as the proposer.

> 4. unbondTime field on sequencer is overloaded

right. I'll rename or add another one

> 5. do you automatically start unbonding period when note period finishes?

no, it starts on the RotationComplete, when last state update received


> 6. you call startRotation even when nextProposer finishes notice period

can't happen as `notice_period` > `max_time_for_last_state_update`.
so even if nextProposer started his notice period, he expected to become `proposer` before notice period ends
I'll add comments and checks


> 7. nextProposer notice period needs to be extended to maximum duration when he becomes active proposer

not sure why.

> 8. when is UnbondRequestHeight, unbondTime reset to 0?

`unbondTime` sets to the transition time (notice period complete / unbonding complete)

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,122 @@
...
+		return types.Sequencer{}
+	}
+
+	// take the next bonded sequencer to be the proposer. sorted by bond
+	sort.SliceStable(seqs, func(i, j int) bool {
+		return seqs[i].Tokens.IsAllGT(seqs[j].Tokens)
+	})
```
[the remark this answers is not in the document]
mtsitrin: yes,bond is single coin

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
mtsitrin: fixed. maturing notice period is only for the proposer

x/sequencer/keeper/hook_listener.go
```diff
@@ -37,9 +37,24 @@ func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rolla
 		return types.ErrInvalidSequencerStatus
```
omritoptix, line 37: isnt this check redundant if we check later if the seq is the proposer?

x/sequencer/keeper/hook_listener.go
```diff
@@ -37,9 +37,24 @@ func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rolla
...
 	}
 
-	if !sequencer.Proposer {
+	seq, ok := hook.k.GetProposer(ctx, rollappId)
+	if !ok {
+		return types.ErrNoProposer
+	}
+	if sequencer.SequencerAddress != seq.SequencerAddress {
```
omritoptix, line 44: should we change the err to `ErrNotProposer` to be aligned with conventions?

x/sequencer/keeper/hook_listener.go
```diff
@@ -20,7 +20,7 @@ func (k Keeper) RollappHooks() rollapptypes.RollappHooks {
 	}
 }
 
-func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rollappId string) error {
+func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr, rollappId string, lastStateOfSequencer bool) error {
```
omritoptix, line 23: can we have a godoc on this function?

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,136 @@
...
+	seqs := k.GetMatureNoticePeriodSequencers(ctx, currTime)
+	for _, seq := range seqs {
+		if k.IsProposer(ctx, seq.RollappId, seq.SequencerAddress) {
+			k.StartRotation(ctx, seq.RollappId)
+			k.removeNoticePeriodSequencer(ctx, seq)
+		}
+	}
+}
+
+// IsRotating returns true if the rollapp is currently rotating proposers
```
omritoptix, line 49: can u expand godoc on what does rotating proposers mean?

x/sequencer/keeper/sequencer.go
```diff
@@ -19,14 +18,13 @@ func (k Keeper) SetSequencer(ctx sdk.Context, sequencer types.Sequencer) {
...
-	// To support InitGenesis scenario
-	if sequencer.Status == types.Unbonding {
-		k.setUnbondingSequencerQueue(ctx, sequencer)
-	}
+func (k Keeper) UpdateSequencer(ctx sdk.Context, sequencer types.Sequencer) {
+	k.UpdateSequencerWithStateChange(ctx, sequencer, sequencer.Status)
 }
 
-func (k Keeper) UpdateSequencer(ctx sdk.Context, sequencer types.Sequencer, oldStatus types.OperatingStatus) {
+func (k Keeper) UpdateSequencerWithStateChange(ctx sdk.Context, sequencer types.Sequencer, oldStatus types.OperatingStatus) {
```
omritoptix, line 27: `UpdateSequencerWithStateChange` - does `state` here means `status` ? if so i think better change it.
Also if I think it's cleaner to have the function interface having oldStatus and newSTatus and not assume caller will change it on the object itself (cause otherwise it's not necessarily only updating the status but possibly other fields)

x/sequencer/keeper/hook_listener.go
omritoptix: why is that? I think it actually doesn't make sense cause most likely one of those sequencers will be the one starting the forked rollapp (we can look at it as a non-graceful rotation) and in essence we lock them for 21 days.

x/sequencer/keeper/hook_listener.go
```diff
@@ -58,12 +73,13 @@ func (hook rollappHook) FraudSubmitted(ctx sdk.Context, rollappID string, height
 	// unbond all other bonded sequencers
 	sequencers := hook.k.GetSequencersByRollappByStatus(ctx, rollappID, types.Bonded)
```
omritoptix, line 74: what if there are sequencers which are in unbonding status? Look at my next comment but I think we should simply unbond all of the sequencers regardless of their status (and slash proposer obviously)

x/sequencer/keeper/hook_listener.go
mtsitrin: I've changed it as:
* Currently, seqeuncer is coupled with rollapp. even after unbonded.
to sequence a new rollapp, one needs a different sequencer address
* I thought there might be different scenarios where u might need to slash the **non-proposer** (e.g the next proposer)
thought the safe side will be to enforce unbonding period

x/sequencer/keeper/rotation.go
omritoptix, line 49:
```suggestion
// IsRotating returns true if the rollapp is currently in the process of rotation. A process of rotation
// is defined by the time between when the proposer notice period is over till the proposer sends his last batch (assuming there was a next proposer in queue). 
```

x/sequencer/keeper/msg_server_unbond.go
```diff
@@ -14,54 +13,42 @@ import (
...
-			sdk.NewAttribute(types.AttributeKeyCompletionTime, completionTime.String()),
-		),
-	)
-
-	if wasProposer {
-		k.RotateProposer(ctx, seq.RollappId)
+	if seq.UnbondRequestHeight != 0 {
+		return nil, errorsmod.Wrapf(
+			types.ErrInvalidSequencerStatus,
+			"sequencer has already requested to unbond",
```
omritoptix, line 32: isnt the `!seq.IsBonded()` already covers this case?

x/sequencer/keeper/msg_server_unbond.go
```diff
@@ -14,54 +13,42 @@ import (
...
+		)
+	}
+	seq.UnbondRequestHeight = ctx.BlockHeight()
+
+	// check if sequencer required for a notice period before unbonding
```
omritoptix, line 37:
```suggestion
	// if the sequencer is the proposer, he will be required for a notice period
```

omritoptix, line 37:
```suggestion
	// check if the sequencer is required for a notice period before unbonding
```

x/sequencer/keeper/hook_listener.go
mtsitrin: removed

utils/keys.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+import (
+	"time"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
```
omritoptix, line 8:
```suggestion
// EncodeTimeToKey combines a given byte slice (queueKey) with an encoded representation of a specified time (endTime).
// The resulting byte slice can be used for lexicographical sorting, as the encoded time is in big-endian order.
//
// The function ensures that different times will be sorted correctly when iterated using big-endian order, 
// as the time encoding preserves the natural chronological order.
//
// Example:
//  queueKey := []byte("exampleKey")
//  endTime1 := time.Date(2023, time.October, 1, 12, 0, 0, 0, time.UTC)
//  endTime2 := time.Date(2023, time.October, 2, 12, 0, 0, 0, time.UTC)
//  encodedKey1 := EncodeTimeToKey(queueKey, endTime1)
//  encodedKey2 := EncodeTimeToKey(queueKey, endTime2)
//  fmt.Printf("%x\n", encodedKey1) // Output will be the hexadecimal representation of the combined byte slice for endTime1
//  fmt.Printf("%x\n", encodedKey2) // Output will be the hexadecimal representation of the combined byte slice for endTime2
//  fmt.Println(string(encodedKey1) < string(encodedKey2)) // This will print 'true' as endTime1 is before endTime2
```

proto/dymensionxyz/dymension/sequencer/sequencer.proto
```diff
@@ -16,32 +16,42 @@ import "cosmos/msg/v1/msg.proto";
...
+  // unbond_request_height stores the height at which this sequencer has
+  // requested to unbond.
+  int64 unbond_request_height = 9;
+  // unbond_time defines, if unbonding, the min time for the sequencer to
+  // complete unbonding.
+  google.protobuf.Timestamp unbond_time = 10
+      [ (gogoproto.nullable) = false, (gogoproto.stdtime) = true ];
+
+  // notice_period_time defines the time when the sequencer will finish it's notice period if started
+  google.protobuf.Timestamp notice_period_time = 11
```
omritoptix, line 55: not sure I understand why is this an attribute of the sequencer vs being calculated when needed?

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,136 @@
...
+		return seq
+	}
+
+	seqs := k.GetSequencersByRollappByStatus(ctx, rollappId, types.Bonded)
+	if len(seqs) == 0 {
+		return types.Sequencer{}
+	}
+
+	// take the next bonded sequencer to be the proposer. sorted by bond
+	sort.SliceStable(seqs, func(i, j int) bool {
```
omritoptix, line 74: I think there is an edge case here where you may have a sequencer that is in "reduce bond" period, but it's still not reflected in his Tokens amount. 
In that case imo you want to already include the potential reduct in the highest bond calculation. 
more details here: https://www.notion.so/dymension/ADR-x-Sequencer-Bond-Management-b0a7061435fb40a28bd3cf64b899e968?pvs=4#8f0733cf6d2c427c87c07aeb522a7dff

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,136 @@
...
+
+// StartRotation sets the nextSequencer for the rollapp.
+// This function will not clear the current proposer
+// This function called when the sequencer has finished its notice period
+func (k Keeper) StartRotation(ctx sdk.Context, rollappId string) {
+	// next proposer can be empty if there are no bonded sequencers available
+	nextProposer := k.ExpectedNextProposer(ctx, rollappId)
+	k.SetNextProposer(ctx, rollappId, nextProposer.SequencerAddress)
+
+	k.Logger(ctx).Info("rotation started", "rollappId", rollappId, "nextProposer", nextProposer.SequencerAddress)
```
omritoptix, line 97: it seems a bit misleading as if there is no next proposer, you still log this message and report rotation (though there is no rotation)

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,136 @@
...
+	return types.Sequencer{}
+}
+
+// StartRotation sets the nextSequencer for the rollapp.
+// This function will not clear the current proposer
+// This function called when the sequencer has finished its notice period
+func (k Keeper) StartRotation(ctx sdk.Context, rollappId string) {
+	// next proposer can be empty if there are no bonded sequencers available
+	nextProposer := k.ExpectedNextProposer(ctx, rollappId)
+	k.SetNextProposer(ctx, rollappId, nextProposer.SequencerAddress)
```
omritoptix, line 95: not sure why we need to set the next proposer to empty address if there is no sequencer vs leaving the key empty?

omritoptix (commented):
few minor comments.
overall looks good!

x/sequencer/keeper/msg_server_unbond.go
mtsitrin: no, it might be in notice period

x/sequencer/keeper/rotation.go
mtsitrin: it's differentiate between rotation in progress with empty next proposer, and no rotation in progress

x/sequencer/keeper/rotation.go
mtsitrin: there is rotation. to empty next proposer.
not sure we need to add additional terminology.
we can call it "handover" instead of rotation

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,136 @@
...
+func (k Keeper) startNoticePeriodForSequencer(ctx sdk.Context, seq *types.Sequencer) time.Time {
+	completionTime := ctx.BlockTime().Add(k.NoticePeriod(ctx))
+	seq.NoticePeriodTime = completionTime
+
+	k.UpdateSequencer(ctx, *seq)
+	k.AddSequencerToNoticePeriodQueue(ctx, *seq)
+
+	nextSeq := k.ExpectedNextProposer(ctx, seq.RollappId)
+	if nextSeq.SequencerAddress == "" {
+		k.Logger(ctx).Debug("rollapp will be left with no proposer after notice period", "rollappId", seq.RollappId, "sequencer", seq.SequencerAddress)
```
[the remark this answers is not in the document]
mtsitrin: removed

x/sequencer/keeper/hook_listener.go
mtsitrin: reverted

x/sequencer/keeper/rotation.go
mtsitrin: should be adjusted in the bond management PR.
out of scope for this PR

x/sequencer/keeper/slashing.go
```diff
@@ -31,20 +31,34 @@ func (k Keeper) Slashing(ctx sdk.Context, seqAddr string) error {
...
 	}
+	// remove from notice period queue if needed
+	if seq.IsNoticePeriodInProgress() {
+		k.removeNoticePeriodSequencer(ctx, seq)
+	}
+	// if the slashed sequencer is the proposer, remove it
+	// the caller should rotate the proposer
+	if k.isProposer(ctx, seq.RollappId, seqAddr) {
+		k.removeProposer(ctx, seq.RollappId)
+	}
```
danwt, line 48: if the caller needs to rotate, why cant the caller also remove?

x/sequencer/keeper/slashing.go
```diff
@@ -31,20 +31,34 @@ func (k Keeper) Slashing(ctx sdk.Context, seqAddr string) error {
...
+
+	// if we slash the next proposer, we're in the middle of rotation
+	// instead of removing the next proposer, we set it to empty
+	if k.isNextProposer(ctx, seq.RollappId, seqAddr) {
+		k.setNextProposer(ctx, seq.RollappId, "")
+	}
```
danwt, line 54: Slashing is now very coupled to the proposer rotation
What is supposed to happen in case of partial slash?

x/sequencer/keeper/slashing.go
```diff
@@ -31,20 +31,34 @@ func (k Keeper) Slashing(ctx sdk.Context, seqAddr string) error {
...
 
 	// set the status to unbonded
 	seq.Status = types.Unbonded
 	seq.Jailed = true
-	seq.Proposer = false
-	seq.UnbondingHeight = ctx.BlockHeight()
-	seq.UnbondTime = ctx.BlockHeader().Time
+
+	seq.UnbondRequestHeight = ctx.BlockHeight()
```
danwt, line 60: why are you overwriting this?

x/sequencer/keeper/slashing.go
```diff
@@ -31,20 +31,34 @@ func (k Keeper) Slashing(ctx sdk.Context, seqAddr string) error {
...
+	seq.UnbondTime = ctx.BlockTime()
```
danwt, line 61: shouldn't slashing pass off to an unbond function
Not sure why all this logic is coupled so tightly to slashing

x/sequencer/keeper/unbond.go
```diff
@@ -9,13 +9,36 @@ import (
 	"github.com/osmosis-labs/osmosis/v15/osmoutils"
 )
 
+// startUnbondingPeriodForSequencer sets the sequencer to unbonding status
+// can be called after notice period or directly if notice period is not required
+// caller is responsible for updating the proposer for the rollapp if neeeded
```
danwt, line 14:
```suggestion
// caller is responsible for updating the proposer for the rollapp if needed
```

x/sequencer/keeper/hook_listener.go
```diff
@@ -55,15 +74,20 @@ func (hook rollappHook) FraudSubmitted(ctx sdk.Context, rollappID string, height
...
+	for _, sequencer := range append(bonded, unbonding...) {
+		err := hook.k.unbondSequencer(ctx, sequencer.SequencerAddress)
 		if err != nil {
 			return err
 		}
 	}
 
+	// clear the proposer and next proposer
+	hook.k.removeProposer(ctx, rollappID)
+	hook.k.removeNextProposer(ctx, rollappID)
```
danwt, line 89: duplicated in slashing

x/sequencer/types/errors.go
```diff
@@ -28,4 +28,6 @@ var (
 	ErrUnknownRequest           = errorsmod.Register(ModuleName, 1017, "unknown request")
 	ErrInvalidRequest           = errorsmod.Register(ModuleName, 1018, "invalid request")
 	ErrSequencerJailed          = errorsmod.Register(ModuleName, 1019, "sequencer is jailed")
+	ErrRotationInProgress       = errorsmod.Register(ModuleName, 1020, "sequencer rotation in progress")
+	ErrNoProposer               = errorsmod.Register(ModuleName, 1021, "no proposer")
```
danwt, line 32: should wrap gerrc.ErrNotFound instead of registering a new one

x/sequencer/types/errors.go
```diff
@@ -28,4 +28,6 @@ var (
 	ErrUnknownRequest           = errorsmod.Register(ModuleName, 1017, "unknown request")
 	ErrInvalidRequest           = errorsmod.Register(ModuleName, 1018, "invalid request")
 	ErrSequencerJailed          = errorsmod.Register(ModuleName, 1019, "sequencer is jailed")
+	ErrRotationInProgress       = errorsmod.Register(ModuleName, 1020, "sequencer rotation in progress")
```
danwt, line 31: shoudl wrap gerrc.ErrFailedPrecondition instead of registering a new one

x/sequencer/keeper/unbond.go
```diff
@@ -66,52 +88,9 @@ func (k Keeper) forceUnbondSequencer(ctx sdk.Context, seqAddr string) error {
...
-	}
-
-	if seq.Status != types.Unbonding {
-		return errorsmod.Wrapf(
-			types.ErrInvalidSequencerStatus,
-			"sequencer status is not unbonding: got %s",
-			seq.Status.String(),
-		)
+	if seq.IsNoticePeriodInProgress() {
+		k.removeNoticePeriodSequencer(ctx, seq)
```
danwt, line 92: I don't get why this is here because they should never be able to unbond gracefully if notice period is not finished

if this is just for slash unbond, then it should be in caller

danwt (changes requested):
- In general RotateProposer does not read well in case there is actually no nextProposer (halt case)
    - I guess there is no technical reason to have a separate notice/completion queue?
    - I still dont' get the empty types.Sequencer{} for nextProposer
    - the slashing function couples very tightly with set/remove current/next proposer so how will it work when slashing can be partial? (not all tokens)?
    - what if slash occurs on unbonding sequencer?
    - why are concrete completion times written to queue? why not queue based off the start time and query using the parameter?
    - should have invariants related to the queues IMO
    - what happens if nextProposer notice period elapses?

danwt:
```
Instead of having 

IsRotating(){
    return nextProposer!=nil
}

IsRotating(){
    return currentProposer().NoticePeriodTime < now
}
```

?

Then we don't need nil/empty type in nextPRoposer?

x/sequencer/types/params.go
```diff
@@ -47,19 +52,20 @@ func DefaultParams() Params {
...
 func (p *Params) ParamSetPairs() paramtypes.ParamSetPairs {
 	return paramtypes.ParamSetPairs{
 		paramtypes.NewParamSetPair(KeyMinBond, &p.MinBond, validateMinBond),
-		paramtypes.NewParamSetPair(KeyUnbondingTime, &p.UnbondingTime, validateUnbondingTime),
+		paramtypes.NewParamSetPair(KeyUnbondingTime, &p.UnbondingTime, validateTime),
+		paramtypes.NewParamSetPair(KeyNoticePeriod, &p.NoticePeriod, validateTime),
 	}
 }
 
-func validateUnbondingTime(i interface{}) error {
```
omritoptix, line 62: we should [validate unbonding time > grace period time](https://github.com/dymensionxyz/dymension/issues/613). Also why are we using time here vs blocks (like in the dispute period case?)
seems like it doesn't account for hub going down scenarios. 

in that case, if you have grace period in blocks, and unbonding time in time, it's not correlated.

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,128 @@
...
+
+// MatureSequencersWithNoticePeriod start rotation flow for all sequencers that have finished their notice period
+// The next proposer is set to the next bonded sequencer
+// The hub will expect a "last state update" from the sequencer to start unbonding
+func (k Keeper) MatureSequencersWithNoticePeriod(ctx sdk.Context, currTime time.Time) {
+	seqs := k.GetMatureNoticePeriodSequencers(ctx, currTime)
+	for _, seq := range seqs {
+		if k.isProposer(ctx, seq.RollappId, seq.SequencerAddress) {
+			k.startRotation(ctx, seq.RollappId)
+			k.removeNoticePeriodSequencer(ctx, seq)
```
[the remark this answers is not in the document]
mtsitrin: only proposer can have notice period

x/sequencer/keeper/hook_listener.go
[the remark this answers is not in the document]
mtsitrin: cleaner

mtsitrin: slashing is handling a specific sequencer
while here is at the rollapp level

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,128 @@
...
+	// if nextProposer is set, were in the middle of rotation
+	seq, ok := k.GetNextProposer(ctx, rollappId)
+	if ok {
+		return seq
+	}
+
+	seqs := k.GetSequencersByRollappByStatus(ctx, rollappId, types.Bonded)
+	if len(seqs) == 0 {
+		return types.Sequencer{}
+	}
```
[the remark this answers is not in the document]
mtsitrin: early return

x/sequencer/keeper/msg_server_unbond.go
```diff
@@ -14,54 +13,42 @@ import (
...
 	}
 
 	if !seq.IsBonded() {
-		return time.Time{}, errorsmod.Wrapf(
+		return nil, errorsmod.Wrapf(
 			types.ErrInvalidSequencerStatus,
 			"sequencer status is not bonded: got %s",
 			seq.Status.String(),
 		)
 	}
```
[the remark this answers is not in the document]
mtsitrin: not sure what confusing.
u can't submit unbond request if u not bonded

x/sequencer/keeper/msg_server_unbond.go
mtsitrin: again, not sure what confusing

x/sequencer/keeper/unbond.go
mtsitrin: it's the actual unbond method. it should leave the store clean and insync

x/sequencer/keeper/slashing.go
mtsitrin: because having defined proposer while it's unbonded, is invariant breaking.
when unbonding a sequencer, it can't be left as a proposer.
whether rotation required, is handled by the caller

x/sequencer/keeper/slashing.go
mtsitrin: i'll move the code to unbond handler
than the slashing logic should decide whether unbond required

x/sequencer/keeper/hook_listener.go
```diff
@@ -55,15 +74,20 @@ func (hook rollappHook) FraudSubmitted(ctx sdk.Context, rollappID string, height
 		return err
 	}
 
-	// unbond all other bonded sequencers
-	sequencers := hook.k.GetSequencersByRollappByStatus(ctx, rollappID, types.Bonded)
-	for _, sequencer := range sequencers {
-		err := hook.k.forceUnbondSequencer(ctx, sequencer.SequencerAddress)
+	// unbond all other sequencers
+	bonded := hook.k.GetSequencersByRollappByStatus(ctx, rollappID, types.Bonded)
+	unbonding := hook.k.GetSequencersByRollappByStatus(ctx, rollappID, types.Unbonding)
```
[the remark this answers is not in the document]
mtsitrin: prefer to avoid it for now

mtsitrin:
> ```
> Instead of having 
> 
> IsRotating(){
>     return nextProposer!=nil
> }
> 
> IsRotating(){
>     return currentProposer().NoticePeriodTime < now
> }
> ```
> 
> ?
> 
> Then we don't need nil/empty type in nextPRoposer?

I don't like it. It's not hermetic enough.
I'm leaving it as `Rotation is started == NextProposer is defined` as this is undisputed truth.

danwt:
>I don't like it. It's not hermetic enough.
I'm leaving it as Rotation is started == NextProposer is defined as this is undisputed truth.

It's just very unintuitive, because there **is no next proposer** if it's empty type, but the code reads like there is.

x/sequencer/keeper/msg_server_unbond.go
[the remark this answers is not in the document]
danwt: >u can't submit unbond request if u not bonded
why?

It can just be no op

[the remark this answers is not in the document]
danwt: Like you can delete a key from map if the key is not there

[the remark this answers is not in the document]
mtsitrin: but u want to return an error for this tx, why would u want it to pass `checkTx`?

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
mtsitrin: the protocol needs to know whether rotation in progress.
the most reliable and straight forward is to check wether `NextProposer` is set (regardless of its value)

mtsitrin:
> - what happens if nextProposer notice period elapses?
it will wait for current proposer to complete rotation. nextProposer will be the proposer.
than his notice period will be handled, and he'll start rotating as well

x/sequencer/keeper/sequencer.go
```diff
@@ -19,14 +18,15 @@ func (k Keeper) SetSequencer(ctx sdk.Context, sequencer types.Sequencer) {
...
 }
 
-func (k Keeper) UpdateSequencer(ctx sdk.Context, sequencer types.Sequencer, oldStatus types.OperatingStatus) {
+// UpdateSequencer updates the state of a sequencer in the keeper.
+// Parameters:
+//   - ctx: The context object that provides access to the store and other information.
+//   - sequencer: The sequencer object to be updated.
+//   - oldStatus: An optional parameter representing the old status of the sequencer.
+//     Needs to be provided if the status of the sequencer has changed (e.g from Bonded to Unbonding).
+func (k Keeper) UpdateSequencer(ctx sdk.Context, sequencer types.Sequencer, oldStatus ...types.OperatingStatus) {
```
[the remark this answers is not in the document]
mtsitrin: I know. it's variadic argument to allow calling `UpdateSequencer` w/o providing old status

omritoptix (approved):

x/sequencer/keeper/unbond.go
```diff
@@ -26,158 +49,130 @@ func (k Keeper) UnbondAllMatureSequencers(ctx sdk.Context, currTime time.Time) {
...
+	} else {
+		// remove from notice period queue if needed
+		if seq.IsNoticePeriodInProgress() {
+			k.removeNoticePeriodSequencer(ctx, seq)
 		}
 
-		err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, seqAcc, seqTokens)
-		if err != nil {
-			return err
+		// if we unbond the the proposer, remove it
```
danwt, line 135: the the

x/sequencer/keeper/unbond.go
```diff
@@ -26,158 +49,130 @@ func (k Keeper) UnbondAllMatureSequencers(ctx sdk.Context, currTime time.Time) {
...
-	// set the status to unbonded and remove from the unbonding queue
-	seq.Status = types.Unbonded
-	seq.Tokens = sdk.Coins{}
+	if jail {
+		seq.Jailed = true
+	}
+	// set the unbonding height and time, if not already set
+	if seq.UnbondRequestHeight == 0 {
+		seq.UnbondRequestHeight = ctx.BlockHeight()
+	}
```
danwt, line 161: this is overload of bond request height, it's kind of a fake meaning
why need to set it? should explain

x/sequencer/keeper/unbond.go
```diff
@@ -26,158 +49,130 @@ func (k Keeper) UnbondAllMatureSequencers(ctx sdk.Context, currTime time.Time) {
...
+	if seq.UnbondTime.IsZero() {
+		seq.UnbondTime = ctx.BlockTime()
+	}
```
danwt, line 164: it's not clear why to set it here

>   // unbond_time defines, if unbonding, the min time for the sequencer to complete unbonding

the docstring implies only used to facilitate queue

So why set it here? need to explain in docstring

x/sequencer/keeper/hook_listener.go
```diff
@@ -32,32 +41,39 @@ func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rolla
 		return types.ErrSequencerRollappMismatch
 	}
 
-	// check to see if the sequencer is active and can make the update
-	if sequencer.Status != types.Bonded {
-		return types.ErrInvalidSequencerStatus
+	proposer, ok := hook.k.GetProposer(ctx, rollappId)
+	if !ok {
+		return errors.Join(gerrc.ErrNotFound, types.ErrNoProposer)
 	}
```
danwt, line 47: why join error
no proposer already wraps not found

x/sequencer/keeper/hook_listener.go
```diff
@@ -20,7 +23,13 @@ func (k Keeper) RollappHooks() rollapptypes.RollappHooks {
...
-func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rollappId string) error {
+// BeforeUpdateState checks various conditions before updating the state.
+// It verifies if the sequencer has been registered, if the rollappId matches the one of the sequencer,
+// if there is a proposer for the given rollappId, and if the sequencer is the active one.
+// If the lastStateUpdateBySequencer flag is true, it also checks if the rollappId is rotating and
+// performs a rotation of the proposer.
+// Returns an error if any of the checks fail, otherwise returns nil.
+func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr, rollappId string, lastStateUpdateBySequencer bool) error {
 	// check to see if the sequencer has been registered before
 	sequencer, found := hook.k.GetSequencer(ctx, seqAddr)
```
danwt, line 34: why get the sequencer?
no need
just get the proposer for the rollapp and check the addr matches the msg addr

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,133 @@
...
+			sdk.NewAttribute(types.AttributeKeyRollappId, rollappId),
+			sdk.NewAttribute(types.AttributeKeyNextProposer, nextProposer.Address),
+		),
+	)
+}
+
+// RotateProposer completes the sequencer rotation flow.
+// It's called when a last state update is received from the active, rotating sequencer.
+// it will start unbonding the current proposer, and set new proposer from the bonded sequencers
+func (k Keeper) RotateProposer(ctx sdk.Context, rollappId string) {
```
danwt, line 108: can you call it FinishRotation or completionRotation

to be clearer with startRotation func above, and concept of 'isRotating'

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,133 @@
...
+	nextProposerAddr, ok := k.GetNextProposerAddr(ctx, rollappId)
+	if !ok { // nextProposer is guaranteed to be set by caller
+		k.Logger(ctx).Error("next proposer not set. rotation didn't completed", "rollappId", rollappId)
+		return
+	}
```
danwt, line 113: panic

danwt: Must function

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// start unbonding the current proposer
+	proposer, ok := k.GetProposer(ctx, rollappId)
+	if ok {
+		k.startUnbondingPeriodForSequencer(ctx, &proposer)
+	}
```
danwt, line 119: can move this block above k.GetNextProposerAddr block

x/sequencer/keeper/sequencer.go
```diff
@@ -172,3 +146,127 @@ func (k Keeper) removeUnbondingSequencer(ctx sdk.Context, sequencer types.Sequen
...
+	b := store.Get(types.NextProposerByRollappKey(rollappId))
+	if b == nil {
+		return "", false
+	}
+
+	address := string(b)
+	if address == NO_SEQUENCER_AVAILABLE {
+		return "", true
+	}
+	return address, true
```
danwt, line 254: what is the point of this

x/sequencer/keeper/msg_server_unbond.go
danwt, line 32: not an accurate message since UnbondRequestHeight is overloaded and set elsewhere too

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
danwt: please dont do that
that is very micro optimisation
completely useless and just adds code to read

x/sequencer/keeper/sequencer.go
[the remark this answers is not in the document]
danwt: should be pointer

x/sequencer/keeper/rotation.go
```diff
@@ -0,0 +1,133 @@
...
+	return completionTime
+}
+
+// MatureSequencersWithNoticePeriod start rotation flow for all sequencers that have finished their notice period
+// The next proposer is set to the next bonded sequencer
+// The hub will expect a "last state update" from the sequencer to start unbonding
+func (k Keeper) MatureSequencersWithNoticePeriod(ctx sdk.Context, currTime time.Time) {
+	seqs := k.GetMatureNoticePeriodSequencers(ctx, currTime)
+	for _, seq := range seqs {
+		if k.isProposer(ctx, seq.RollappId, seq.Address) {
```
danwt, line 36: can you document this wrt. non proposer who is in notice period

as I understand, non proposer in notice period must be next proposer
so when rotation completes he become proposer and will immediately finish notice period here

So notice period not working as intended, as a notice from the time they become proposer..

x/sequencer/keeper/rotation.go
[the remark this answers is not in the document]
danwt: I opened another thread

x/sequencer/keeper/unbond.go
```diff
@@ -26,158 +49,130 @@ func (k Keeper) UnbondAllMatureSequencers(ctx sdk.Context, currTime time.Time) {
...
 		return types.ErrUnknownSequencer
 	}
 
-	if seq.Status != types.Unbonding {
+	if seq.Status == types.Unbonded {
 		return errorsmod.Wrapf(
 			types.ErrInvalidSequencerStatus,
-			"sequencer status is not unbonding: got %s",
-			seq.Status.String(),
+			"sequencer status is already unbonded",
```
danwt, line 113: need to document when this can happen
if it can't happen, needs to be a panic
in any case needs docs
we can just have all this defensive code it becomes impossible to understand anything

danwt (changes requested):
Still reviewing

I unresolved 2 or 3 comments from last rounds as well

x/rollapp/keeper/msg_server_update_state.go
```diff
@@ -83,6 +81,15 @@ func (k msgServer) UpdateState(goCtx context.Context, msg *types.MsgUpdateState)
 		FinalizationQueue: newFinalizationQueue,
 	})
 
+	// check if sequencer rotation in progress
+	nextAddr, found := k.sequencerKeeper.GetNextProposerAddr(ctx, msg.RollappId)
+	res := &types.MsgUpdateStateResponse{
+		NextProposerAddr:   nextAddr,
+		RotationInProgress: found,
+	}
```
danwt, line 89: would document here or add ADR faq section/linkt to why we do it like this instead of event

x/sequencer/keeper/grpc_query_sequencers_by_rollapp.go
```diff
@@ -46,3 +45,35 @@ func (k Keeper) SequencersByRollappByStatus(c context.Context, req *types.QueryG
...
+func (k Keeper) GetProposerByRollapp(c context.Context, req *types.QueryGetProposerByRollappRequest) (*types.QueryGetProposerByRollappResponse, error) {
+	if req == nil {
+		return nil, gerrc.ErrInvalidArgument
+	}
+	ctx := sdk.UnwrapSDKContext(c)
+
+	seq, ok := k.GetProposer(ctx, req.RollappId)
+	if !ok {
+		return nil, errors.Join(gerrc.ErrNotFound, types.ErrNoProposer)
+	}
```
danwt, line 59: no need to join

x/sequencer/keeper/hook_listener.go
```diff
@@ -32,32 +41,39 @@ func (hook rollappHook) BeforeUpdateState(ctx sdk.Context, seqAddr string, rolla
...
+
 	return nil
 }
 
 // FraudSubmitted implements the RollappHooks interface
 // It slashes the sequencer and unbonds all other bonded sequencers
 func (hook rollappHook) FraudSubmitted(ctx sdk.Context, rollappID string, height uint64, seqAddr string) error {
-	err := hook.k.SlashAndJailFraud(ctx, seqAddr)
+	err := hook.k.JailSequencerOnFraud(ctx, seqAddr)
 	if err != nil {
```
danwt, line 69: I think we need a PRD for definition and intention of jail

x/sequencer/keeper/msg_server_create_sequencer.go
```diff
@@ -85,10 +83,25 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
...
-		Proposer:     !rollapp.Sealed,
 		Tokens:       bond,
 	}
 
+	// if this is the first sequencer, make it a PROPOSER
+	bondedSequencers := k.GetSequencersByRollappByStatus(ctx, msg.RollappId, types.Bonded)
+	isProposer := len(bondedSequencers) == 0
+	if isProposer {
+		k.SetProposer(ctx, sequencer.RollappId, sequencer.Address)
+	}
```
danwt, line 94: cant you just check if proposer set already?

x/sequencer/keeper/msg_server_create_sequencer.go
```diff
@@ -85,10 +83,25 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
...
+
+    // we currently only support setting next proposer (or empty one) before the rotation started. This is in order to 
+    // avoid handling the case a potential next proposer bonds in the middle of a rotation. 
+    // This will be handled in next iteration.
+	nextProposer := k.IsRotating(ctx, sequencer.RollappId) && k.ExpectedNextProposer(ctx, sequencer.RollappId).IsEmpty()
```
danwt, line 99: why expected here, why not just nextproposer?

x/sequencer/keeper/sequencer.go
```diff
@@ -172,3 +146,127 @@ func (k Keeper) removeUnbondingSequencer(ctx sdk.Context, sequencer types.Sequen
...
+	if len(b) == 0 {
+		return val, false
+	}
+
+	return k.GetSequencer(ctx, string(b))
+}
+
+func (k Keeper) removeProposer(ctx sdk.Context, rollappId string) {
+	k.SetProposer(ctx, rollappId, NO_SEQUENCER_AVAILABLE)
+}
```
danwt, line 224: maybe use same pattern for next propsoer?

x/sequencer/keeper/sequencer.go
```diff
@@ -10,6 +9,9 @@ import (
 	"github.com/dymensionxyz/dymension/v3/x/sequencer/types"
 )
 
+// used to indicate that no sequencer is available for a proposer / next proposer role
+const NO_SEQUENCER_AVAILABLE = ""
```
danwt, line 13: the sentinel only makes sense if you use it with direct value comparison, not len(x)==0

x/sequencer/keeper/unbond.go
```diff
@@ -26,158 +49,130 @@ func (k Keeper) UnbondAllMatureSequencers(ctx sdk.Context, currTime time.Time) {
...
 
-func (k Keeper) HandleBondReduction(ctx sdk.Context, currTime time.Time) {
-	unbondings := k.GetMatureDecreasingBondSequencers(ctx, currTime)
-	for _, unbonding := range unbondings {
-		wrapFn := func(ctx sdk.Context) error {
-			return k.completeBondReduction(ctx, unbonding)
-		}
-		err := osmoutils.ApplyFuncIfNoError(ctx, wrapFn)
+// InstantUnbondAllSequencers unbonds all sequencers for a rollapp
+// This is called when the rollapp is frozen
```
danwt, line 53:
```suggestion
// This is called when there is a fraud
```

x/sequencer/types/errors.go
```diff
@@ -25,6 +25,8 @@ var (
 	ErrUnknownRequest           = errorsmod.Register(ModuleName, 1017, "unknown request")
 	ErrInvalidRequest           = errorsmod.Register(ModuleName, 1018, "invalid request")
 	ErrSequencerJailed          = errorsmod.Register(ModuleName, 1019, "sequencer is jailed")
+	ErrRotationInProgress       = errorsmod.Wrap(gerrc.ErrFailedPrecondition, "sequencer rotation in progress")
+	ErrNoProposer               = errorsmod.Wrap(gerrc.ErrNotFound, "no proposer")
```
danwt, line 29:
```suggestion
	ErrNoProposer               = errorsmod.Wrap(gerrc.ErrNotFound, "proposer")
```

danwt (changes requested):
finished review
looks good

I left all important comments in previous reviews

Biggest red flag remaining for me now is a bit out of scope: definition of jail

x/sequencer/keeper/bond_reductions.go
```diff
@@ -3,11 +3,62 @@ package keeper
...
+		return err
+	}
+
+	seq.Tokens = seq.Tokens.Sub(reduction.DecreaseBondAmount)
+	k.SetSequencer(ctx, seq)
+	k.removeDecreasingBondQueue(ctx, reduction)
+
+	return nil
+}
+
```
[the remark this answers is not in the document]
mtsitrin: yes. just moved

x/sequencer/keeper/slashing.go
```diff
@@ -81,37 +74,16 @@ func (k Keeper) Slash(ctx sdk.Context, seq *types.Sequencer, amt sdk.Coins) erro
 }
 
 func (k Keeper) Jail(ctx sdk.Context, seq types.Sequencer) error {
-	oldStatus := seq.Status
```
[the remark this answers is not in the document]
mtsitrin: added godoc. left Jail as is

x/sequencer/keeper/unbond.go
```diff
@@ -26,158 +49,130 @@ func (k Keeper) UnbondAllMatureSequencers(ctx sdk.Context, currTime time.Time) {
...
 
-	return nil
+func (k Keeper) unbondSequencer(ctx sdk.Context, seqAddr string) error {
+	return k.unbond(ctx, seqAddr, false)
 }
 
-// unbondUnbondingSequencer unbonds a sequencer that currently unbonding
-func (k Keeper) unbondUnbondingSequencer(ctx sdk.Context, seqAddr string) error {
+func (k Keeper) unbond(ctx sdk.Context, seqAddr string, jail bool) error {
 	seq, found := k.GetSequencer(ctx, seqAddr)
```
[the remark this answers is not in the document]
mtsitrin: to avoid double writing to the store

x/sequencer/keeper/unbond.go
mtsitrin: fixed proto doc

x/sequencer/keeper/rotation.go
mtsitrin: i'll return an error

x/sequencer/keeper/rotation.go
mtsitrin: why?

x/sequencer/keeper/sequencer.go
mtsitrin: wdym

x/sequencer/keeper/msg_server_unbond.go
mtsitrin: its fine. it's set when forcefully unbonded

x/sequencer/keeper/msg_server_create_sequencer.go
mtsitrin: yeah possible

x/rollapp/keeper/msg_server_update_state.go
mtsitrin: removed

x/sequencer/keeper/unbond.go
mtsitrin: panic is bad practice. bugs can happen.
not sure what's wrong with it

x/sequencer/keeper/sequencer.go
mtsitrin: discussed already. for nextProposer there's a difference between set/unset and set with ""

x/sequencer/keeper/sequencer.go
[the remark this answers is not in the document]
mtsitrin: the point was to not break the api

omritoptix (approved):

x/sequencer/keeper/unbond.go
danwt: it's dead code basically

danwt (approved):
There are still couple things wrong (old unresolved comments)

https://github.com/dymensionxyz/dymension/blob/4b26022985fc19b5027eb697e2bcf96bbcd9b7cd/x/sequencer/keeper/sequencer.go#L27-L40 should not be variadic

https://github.com/dymensionxyz/dymension/blob/4b26022985fc19b5027eb697e2bcf96bbcd9b7cd/x/sequencer/keeper/sequencer.go#L12-L13 empty sentinel is weird

I cannot find doc for this https://github.com/dymensionxyz/dymension/pull/1006#discussion_r1716638405

Jail parts still doesn't make sense IMO (depending on PRD)

But these can be addressed after merge

Good job 👍

PR #1022, opened 2024-07-30
feat(rollapp): Base rollapp registration fee on alias number of characters pricing

### LINKED ISSUE #1013: Add rollapp creation price per alias characters length

zale144:
…ength

## Description

----

Closes #1013 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----;

For Reviewer:

- [ ]  Reviewers assigned

---;

After reviewer approval:


proto/dymensionxyz/dymension/rollapp/params.proto
```diff
@@ -16,7 +16,7 @@ message Params {
   uint64 dispute_period_in_blocks = 1
       [ (gogoproto.moretags) = "yaml:\"dispute_period_in_blocks\"" ];
 
-  // registration_fee is the fee that is required to register a rollapp
-  cosmos.base.v1beta1.Coin registration_fee = 2
-      [ (gogoproto.nullable) = false, (gogoproto.moretags) = "yaml:\"registration_fee\"" ];
+  // alias_fee_table is a map of alias to pricing table, which is a map of character length to price.
+  map<string, cosmos.base.v1beta1.Coin> alias_fee_table = 3
```
VictorTrustyDev, line 20: We should not use map because [the order is random](https://stackoverflow.com/a/60512308)

x/rollapp/types/rollapp.go
```diff
@@ -93,7 +89,7 @@ func (r Rollapp) ValidateBasic() error {
 }
 
 func validateAlias(alias string) error {
```
VictorTrustyDev, line 91: later please use the `dymnsutils.IsValidAlias`

proto/dymensionxyz/dymension/rollapp/tx.proto
```diff
@@ -52,12 +52,10 @@ message MsgUpdateRollappInformation {
...
   // sequencer that is allowed to initially serve this rollappId.
   string initial_sequencer_address = 3;
-  // alias is the chain alias used for display and namespace system
-  string alias = 4;
   // checksum used to verify integrity
-  string genesis_checksum = 5;
+  string genesis_checksum = 4;
   // metadata is the rollapp metadata
-  RollappMetadata metadata = 6 [(gogoproto.nullable) = true ];
+  RollappMetadata metadata = 5 [(gogoproto.nullable) = true ];
```
danwt, line 58: stop abusing proto field numbers
just reserve them please

x/rollapp/keeper/msg_server_create_rollapp_test.go
```diff
@@ -41,6 +42,41 @@ func (suite *RollappTestSuite) TestCreateRollappAlreadyExists() {
 	suite.ErrorIs(err, types.ErrRollappIDExists)
 }
 
+func (suite *RollappTestSuite) TestCreateRollappNotEnoughFundsForAlias() {
+	rollapp := types.MsgCreateRollapp{
```
danwt, line 46: can you add a high level test description in docstring?

x/rollapp/keeper/params.go
```diff
@@ -25,7 +29,35 @@ func (k Keeper) DisputePeriodInBlocks(ctx sdk.Context) (res uint64) {
 	return
 }
 
-func (k Keeper) RegistrationFee(ctx sdk.Context) (res sdk.Coin) {
-	k.paramstore.Get(ctx, types.KeyRegistrationFee, &res)
+func (k Keeper) AliasFeeTable(ctx sdk.Context) (res map[string]sdk.Coin) {
+	k.paramstore.Get(ctx, types.KeyAliasFeeTable, &res)
```
danwt, line 33: why map string instead of map int?

x/rollapp/keeper/params.go
```diff
@@ -25,7 +29,35 @@ func (k Keeper) DisputePeriodInBlocks(ctx sdk.Context) (res uint64) {
...
+	if aliasLength > maxAliasLength {
+		aliasLength = maxAliasLength
+	}
+
+	if aliasPrice, ok := aliasPricingTable[fmt.Sprint(aliasLength)]; ok {
+		return aliasPrice
+	}
+
+	return sdk.NewCoin(appparams.BaseDenom, sdk.NewInt(0))
+}
```
danwt, line 63: Please just put this in an alias_price.go file or similar

danwt: also can you you use min/max functions instead of ifs

x/rollapp/keeper/params.go
danwt: also I'd just make this return a struct `type aliasPriceMap struct` which automatically takes care of getting the max length one, so you dont need to do that logic in  GetPriceForAlias

x/rollapp/keeper/rollapp.go
```diff
@@ -24,9 +25,11 @@ func (k Keeper) RegisterRollapp(ctx sdk.Context, rollapp types.Rollapp) error {
 	}
 
 	creator, _ := sdk.AccAddressFromBech32(rollapp.Creator)
-	registrationFee := sdk.NewCoins(k.RegistrationFee(ctx))
+	registrationFee := sdk.NewCoins(k.GetPriceForAlias(ctx, rollapp.Alias))
 
 	if !registrationFee.IsZero() {
+		bal := k.bankKeeper.(bankkeeper.Keeper).SpendableCoins(ctx, creator).String()
+		fmt.Println("Balance: ", bal)
```
danwt, line 32: println

danwt (changes requested):
Would it be simler and safer to have the table actually just be a list of coins
then you can find the right one with list[len(alias)-1]

proto/dymensionxyz/dymension/rollapp/tx.proto
zale144: `MsgUpdateRollappInformation` doesn't exist in `main`, this is a branch out of `zale144/978-new-create-rollapp-design`

x/rollapp/keeper/params.go
zale144: > why map string instead of map int?

for some reason protobuf wants string as key.

zale144:
> Would it be simler and safer to have the table actually just be a list of coins then you can find the right one with list[len(alias)-1]

yeah, that was my first solution, but then I figured it's too simple.

I mean, I didn't want the alias length to be interpreted as the list index, it didn't seem too clear

proto/dymensionxyz/dymension/rollapp/tx.proto
danwt: ok sure

x/rollapp/keeper/params.go
danwt: 🤔

omritoptix:
closing as moved this logic to dymns

PR #1153, opened 2024-08-28
fix(incentives): fixed lockups by denom cache, added some optimizations from Osmosis

### LINKED ISSUE #1143: x/incentives: locks by denom cache is filled out of place
Currently, [locks by denom cache](https://github.com/dymensionxyz/dymension/blob/main/x/incentives/keeper/gauge_asset.go#L110) is cleaned and filled on every gauge processing which makes it useless. Its main idea is to persist the list of lockups throughout the whole gauge processing process to omit unnecessary requests to `x/lockup`, so filling it again and again on every gauge processing does nothing but additional overhead.

keruch:
## Description

Moved lockups by denom cache to the right place in x/incentives.

----

Closes #1143 

PR review checkboxes:

I have...

- [x]  Wrote unit and integration tests
- [x]  Wrote relevant migration scripts if necessary
- [x]  All CI checks have passed
- [x]  Added relevant `godoc` [comments](https://blog.golang.org/godoc-documenting-go-code)
- [x]  Updated the scripts for local run, e.g genesis_config_commands.sh if the PR changes parameters

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:

### DIFF
--- a/x/incentives/keeper/gauge_asset.go
+++ b/x/incentives/keeper/gauge_asset.go
@@ -187,13 +187,7 @@ func SumLocksByDenom(locks []lockuptypes.PeriodLock, denom string) (math.Int, er
 		return math.Int{}, fmt.Errorf("invalid denom used internally: %s, %w", denom, err)
 	}
 	for _, lock := range locks {
-		var amt math.Int
-		// skip a 1second cumulative runtimeEq check
-		if len(lock.Coins) == 1 {
-			amt = lock.Coins[0].Amount
-		} else {
-			amt = lock.Coins.AmountOfNoDenomValidation(denom)
-		}
+		amt := lock.Coins.AmountOfNoDenomValidation(denom)
 		sumBi.Add(sumBi, amt.BigIntMut())
 	}
 


x/incentives/keeper/gauge_asset.go
```diff
@@ -140,34 +142,92 @@ func (k Keeper) distributeToAssetGauge(ctx sdk.Context, gauge types.Gauge, currR
...
+	// validate the denom once, so we can avoid the expensive validate check in the hot loop.
+	err := sdk.ValidateDenom(denom)
+	if err != nil {
+		return math.Int{}, fmt.Errorf("invalid denom used internally: %s, %w", denom, err)
+	}
+	for _, lock := range locks {
+		var amt math.Int
+		// skip a 1second cumulative runtimeEq check
+		if len(lock.Coins) == 1 {
+			amt = lock.Coins[0].Amount
```
zale144, line 193: redundant. `AmountOfNoDenomValidation` already does this

x/incentives/keeper/gauge_asset.go
```diff
@@ -140,34 +142,92 @@ func (k Keeper) distributeToAssetGauge(ctx sdk.Context, gauge types.Gauge, currR
...
+		} else {
+			amt = lock.Coins.AmountOfNoDenomValidation(denom)
+		}
+		sumBi.Add(sumBi, amt.BigIntMut())
+	}
+
+	// handle overflow check here so we don't panic.
+	err = checkBigInt(sumBi)
```
zale144, line 201: What is the reason to return `math.Int` instead of `big.Int` ?
Could you not do the subsequent calculations (`amt := coin.Amount.Mul(denomLockAmt).Quo(lockSumTimesRemainingEpochs)`) with `big.Int`, and later cast down back into `math.Int`? Since there is a division there, I have a hunch that the end result would not overflow `math.Int`.
If that's the case, then `checkBigInto` and `maxWordLen` can be removed.

x/incentives/types/expected_keepers.go
```diff
@@ -16,19 +16,14 @@ type BankKeeper interface {
 
 	HasSupply(ctx sdk.Context, denom string) bool
 
-	SendCoinsFromModuleToAccount(
-		ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins,
-	) error
+	SendCoinsFromModuleToAccount(ctx sdk.Context, senderModule string, recipientAddr sdk.AccAddress, amt sdk.Coins) error
```
zale144, line 19: I would try using `InputOutputCoins` here if possible. 
This way, there can be a single input (one module account balance update equivalent to the sum of all locks rewards) and `n` outputs, instead of `n` inputs and `n` outputs, as long as the `sum(inputs)=sum(outputs)`. 
So,  instead of `n+n`, you would end up with `n+1` account balance updates.

zale144 (changes requested):
nice work. left some comments

x/incentives/keeper/gauge_asset.go
zale144: I was just thinking if it would be possible/beneficial to do all calculations with `big.Int`, and then at the end when you would have to add it to `sdk.Coin`, to downcast, and potentially check for overflow.
Doesn't matter, just thinking.

x/incentives/types/expected_keepers.go
zale144: I think for this use-case, it would be a good fit, I believe this scenario would be adequate for its intended purpose. Also, gas would be lower probably

zale144 (approved):

PR #1173, opened 2024-09-02
feat(streamer): distribute rewards immediately in the current block

keruch:
## Description

https://www.notion.so/dymension/ADR-x-Incentives-DoS-47238d8a29734ee686e25caef8b68c4f
----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


zale144 (changes requested):
so, I can see that the logic in both `EndBlock` and `AfterEpochEnd` roughly has some similarities:

```
1. GetActiveStreams
2. CalculateRewards
3. SendCoinsFromModuleToModule
4. ik.Distribute
```

is there a way to somehow extract that logic into a common method?
the main difference I see is that `EndBlock` iterates over all `epochPointers` while `AfterEpochEnd` distributes for the current `epochPointer`

the is that the implementation is a bit convoluted, it's not trivial to follow the logic, so maybe it would help to centralize those critical pieces of logic

x/incentives/keeper/distribute.go
```diff
@@ -4,16 +4,35 @@ import (
...
 
-// Distribute distributes coins from an array of gauges.
+// DistributeOnEpochEnd distributes coins from an array of gauges.
 // It is called at the end of each epoch to distribute coins to the gauges that are active at that time.
-func (k Keeper) Distribute(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, error) {
+func (k Keeper) DistributeOnEpochEnd(ctx sdk.Context, gauges []types.Gauge) (sdk.Coins, error) {
+	cache := types.NewDenomLocksCache()
+
+	const EpochEnd = true
+	totalDistributedCoins, err := k.Distribute(ctx, gauges, cache, EpochEnd)
```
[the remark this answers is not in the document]
zale144: not needed, but I think it makes it easier to understand what we're passing, rather than just passing `true`

x/streamer/keeper/distribute.go
```diff
@@ -5,110 +5,240 @@ import (
...
 			var err error
-			distributed, err = k.DistributeToGauge(ctx, v.Stream.EpochCoins, v.Gauge, v.Stream.DistributeTo.TotalWeight)
-			return err
-		})
+			gauge, err = k.getActiveGaugeByID(ctx, v.Gauge.GaugeId)
+			if err != nil {
+				// we don't want to fail in this case, ignore this gauge
+				k.Logger(ctx).
+					With("gaugeID", v.Gauge.GaugeId, "error", err.Error()).
+					Error("Can't distribute to gauge: failed to get active gauge")
```
danwt, line 170: when could this occur? can you document?

danwt: and why validate and then upsert?

Why not just upsert and validate later?

x/streamer/keeper/distribute.go
```diff
@@ -5,110 +5,240 @@ import (
...
-		NewPointer:       newPointer,
-		FilledStreams:    streams, // Make sure that the returning slice is always sorted
-		DistributedCoins: totalDistributed,
-		Iterations:       iterations,
+// getGaugeWeight calculates the weight of a gauge based on its type.
+// If the gauge is an asset gauge, the weight is equal to the number of associated lockups. Use cache to get this figure.
+// If the gauge is a rollapp gauge, the weight is always 1. If the gauge has an unknown type, the weight is 0 as we
+// assume that the gauge is always validated at this step and has a known type. The method also fills the cache under
+// the hood.
+func (k Keeper) getGaugeWeight(ctx sdk.Context, gauge incentivestypes.Gauge, cache incentivestypes.DenomLocksCache) uint64 {
```
danwt, line 230: the weight terminology is confusing because you have weight for people allocating different weights to different guages, and then the weight concept of iteration



Possible to use another term?

danwt (changes requested):
I will continue the review but it's better I ask questions first, mainly I don't quite understand the intent behind the linked list and its wrapper

https://github.com/dymensionxyz/dymension/pull/1173/files

Somehow I can't see the forest through the trees when it comes to the data structures and it's usage

- please docstring all public methods on InsertionOrdered
- in general better documentation of insertion ordered data structure
- why need key func in insertion ordered?
- no need for insertion ordered upsert multiple and upsert, just have upsert take variadic
- Linked lists are not performant AT ALL afaiu, use a more hardware friendly data structure? does perf matter?
- why linked list? what property there is important?
- Use data structure library instead of implement from scratch? If implement then put in sdk-utils?, (and test with rapid?)
- if it's in common utility it should be common, but data structure seems pretty specialized to use case

Other

- possible to dry out endblock and AfterEpochEnd?
- don't interpolate data into error strings
- golang support iterators natively in 1.23 did you consider?
- general weighted pagination maybe better called weightedPagination or something
- x/streamer/keeper/abci_test.go is huge, would it have been possible to use rapid?
- confusing between weight of gauages and weight of iteration

utils/cache/ordered.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// MustGet is Get that panics when the key is not found.
+func (c *InsertionOrdered[K, V]) MustGet(key K) (zero V) {
+	idx, ok := c.keyToIdx[key]
+	if ok {
+		return c.idxToValue[idx]
+	}
+	panic(fmt.Errorf("internal contract error: key is not found in the cache: %v", key))
+}
```
danwt, line 75: use .Get() internally

danwt (approved):
good work monsieur

danwt (approved):
lgtm nice work

zale144 (approved):
beautiful

VictorTrustyDev:
Hi @keruch, I have a question.

Is this PR moved incentives distribution (around 12k transfer as I see on Mainnet) from begin blocker to end blocker?

PR #1201, opened 2024-09-05
feat(IRO): IRO module implementation

### LINKED ISSUE #1164: IRO Impl
Implement IRO based on this ADR: https://www.notion.so/dymension/ADR-x-IRO-533f1c2f5a2345838150dd8e954c99c6?pvs=4

mtsitrin:
## Description

Implement IRO based on this ADR: https://www.notion.so/dymension/ADR-x-IRO-533f1c2f5a2345838150dd8e954c99c6?pvs=4

----

Closes #1164 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


x/iro/keeper/settle.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func determineLimitingFactor(unsoldTokens, raisedDYM math.Int, ratio math.LegacyDec) (tokens, dym math.Int) {
+	requiredDYM := unsoldTokens.ToLegacyDec().Mul(ratio).TruncateInt()
+
+	// if raisedDYM is less than requiredDYM, than DYM is the limiting factor
+	// we use all the raisedDYM, and the corresponding amount of tokens
+	if raisedDYM.LT(requiredDYM) {
+		dym = raisedDYM
+		tokens = raisedDYM.ToLegacyDec().Quo(ratio).TruncateInt()
```
omritoptix, line 136: do we have any gurantee tokens wont be 0?

x/iro/keeper/settle.go
```diff
@@ -0,0 +1,151 @@
...
+	incentives := sdk.NewCoins(
+		sdk.NewCoin(dymLiquidityCoin.Denom, raisedDYM.Amount.Sub(dymLiquidityCoin.Amount)),
+		sdk.NewCoin(rollappLiquidityCoin.Denom, unallocatedTokens.Sub(rollappLiquidityCoin.Amount)),
+	)
+	distrTo := lockuptypes.QueryCondition{
+		LockQueryType: lockuptypes.ByDuration,
+		Denom:         poolDenom,
+		Duration:      k.ik.GetLockableDurations(ctx)[0],
+	}
+	_, err = k.ik.CreateGauge(ctx, false, k.AK.GetModuleAddress(types.ModuleName), incentives, distrTo, ctx.BlockTime().Add(time.Hour), 30)
```
omritoptix, line 121: the 30 epochs should prob be a gov param.

omritoptix (changes requested):
looks good!
didn't review the trading logic yet though.

x/iro/keeper/create_plan.go
```diff
@@ -0,0 +1,150 @@
...
+		DenomUnits: []*banktypes.DenomUnit{
+			{Denom: baseDenom, Exponent: 0, Aliases: []string{}},
+			{Denom: displayDenom, Exponent: uint32(exponent), Aliases: []string{}}, //nolint:gosec
+		},
+		Base:    baseDenom,
+		Name:    baseDenom,
+		Display: displayDenom,
+		Symbol:  displayDenom,
+	}
+	if err := metadata.Validate(); err != nil {
```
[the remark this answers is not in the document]
mtsitrin: I guess it's not mandatory.
but as we create it with user's input, I guess it's better to validate anyway, than taking the risk of putting data in the store that can break the system later on.
(as we assume the data in the bank metadata store is valid)

x/iro/keeper/create_plan.go
```diff
@@ -0,0 +1,150 @@
...
+		return sdk.Coin{}, errorsmod.Wrap(errors.Join(gerrc.ErrInternal, err), fmt.Sprintf("metadata: %v", metadata))
+	}
+
+	if k.BK.HasDenomMetaData(ctx, baseDenom) {
```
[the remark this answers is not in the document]
mtsitrin: as above, IMO it's better to be stuck in this case rather than having corrupted data in store.
If u think it's redundent, it can be removed.

x/iro/keeper/settle.go
```diff
@@ -0,0 +1,151 @@
...
+
+	return nil
+}
+
+// Bootstrap liquidity pool with the raised DYM and unsold tokens
+func (k Keeper) bootstrapLiquidityPool(ctx sdk.Context, plan types.Plan) error {
+	unallocatedTokens := plan.TotalAllocation.Amount.Sub(plan.SoldAmt)        // assumed > 0, as we enforce it in the Buy function
+	raisedDYM := k.BK.GetBalance(ctx, plan.GetAddress(), appparams.BaseDenom) // assumed > 0, as we enforce it by IRO creation fee
+
+	// send the raised DYM to the iro module
```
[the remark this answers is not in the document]
mtsitrin: we create the pool and incentives with the `x/iro` as creator

x/iro/keeper/settle.go
```diff
@@ -0,0 +1,151 @@
...
+
+// TransfersEnabled called by the genesis transfer IBC module when a transfer is handled
+// This is a rollapp module hook
+func (k Keeper) TransfersEnabled(ctx sdk.Context, rollappId, rollappIBCDenom string) error {
+	return k.Settle(ctx, rollappId, rollappIBCDenom)
+}
+
+// Settle settles the iro plan with the given rollappId
+func (k Keeper) Settle(ctx sdk.Context, rollappId, rollappIBCDenom string) error {
+	plan, found := k.GetPlanByRollapp(ctx, rollappId)
```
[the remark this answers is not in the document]
mtsitrin: it's called even if no IRO plan exists.
`transfersEnabled` hoold will be called on the first regular transfer, if no IRO plan exist

x/iro/keeper/settle.go
mtsitrin: my bad. it's set by the plan creator

x/rollapp/transfergenesis/ibc_module.go
```diff
@@ -113,43 +123,107 @@ func (w IBCModule) OnRecvPacket(
...
+		if !noMemo {
+			l.Error("Genesis transfers already enabled.")
+			_ = w.handleDRSViolation(ctx, ra.RollappId)
+			return channeltypes.NewErrorAcknowledgement(ErrDisabled)
+		}
+		return w.IBCModule.OnRecvPacket(ctx, packet, relayer)
+	}
+
+	// check if iro plan exists. if it does, a genesis transfer is required
+	plan, found := w.iroKeeper.GetPlanByRollapp(ctx, ra.RollappId)
```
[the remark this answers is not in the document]
mtsitrin: if a plan is settled, than `TransfersEnabled` is true above

x/rollapp/keeper/rollapp.go
```diff
@@ -116,18 +117,36 @@ func (k Keeper) SetRollapp(ctx sdk.Context, rollapp types.Rollapp) {
...
+	}
+
 	if rollapp.Launched {
 		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "rollapp already launched")
 	}
 
-	if !rollapp.AllImmutableFieldsAreSet() {
-		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "seal with immutable fields not set")
+	// if genesis info is not sealed, seal it if possible
+	if !rollapp.GenesisInfo.Sealed {
```
omritoptix, line 144: imo this can be checked as part of `AllImmutableFieldsAreSet` vs having 2 seperate places for checking pre-conditions or alternatively simply call `Seal` genesis which checks the precondition and fails otherwise.

x/iro/types/msgs.go
```diff
@@ -0,0 +1,113 @@
...
+	_ sdk.Msg = &MsgUpdateParams{}
+)
+
+func (m *MsgCreatePlan) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.Owner)
+	if err != nil {
+		return sdkerrors.ErrInvalidAddress.Wrapf("invalid owner address: %s", err)
+	}
+	plan := NewPlan(1, m.RollappId, sdk.NewCoin("dummy", m.AllocatedAmount), m.BondingCurve, m.StartTime, m.PreLaunchTime, m.IncentivePlanParams)
+	return plan.ValidateBasic()
```
omritoptix, line 22: why do create a dummy plan here and validate it with dummy values vs not just validating the MsgCreatePlan?
for example, in the plan we validate that `ClaimedAmt` and `SoldAmt` are not negative, but in essence in the `MsgCreatePlan` they should be validated against 0.

x/iro/types/msgs.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (m *MsgBuy) ValidateBasic() error {
+	// buyer bech32
+	_, err := sdk.AccAddressFromBech32(m.Buyer)
+	if err != nil {
+		return sdkerrors.ErrInvalidAddress.Wrapf("invalid buyer address: %s", err)
+	}
+
+	// coin exist and valid
+	if m.Amount.IsNil() || m.Amount.IsZero() || m.Amount.IsNegative() {
```
omritoptix, line 38: why not just `m.Amount.IsPositive` ?

x/iro/types/msgs.go
```diff
@@ -0,0 +1,113 @@
...
+		return sdkerrors.ErrInvalidRequest.Wrapf("amount %v must be positive", m.Amount)
+	}
+
+	if m.ExpectedOutAmount.IsNil() || m.ExpectedOutAmount.IsZero() || m.ExpectedOutAmount.IsNegative() {
```
omritoptix, line 42: same here. just `ExpectedOutAmount.IsPositive`

also for the rest of the `ValidateBasic`

x/iro/types/bonding_curve.go
```diff
@@ -0,0 +1,80 @@
...
+	// linear bonding curve as default
+	return BondingCurve{
+		M: math.LegacyMustNewDecFromStr("0.005"),
+		N: math.LegacyOneDec(),
+		C: math.LegacyZeroDec(),
+	}
+}
+
+// validateBasic checks if the bonding curve is valid
+func (lbc BondingCurve) ValidateBasic() error {
```
omritoptix, line 34: as discussed with yishay we should probably add some sane valiation here (i.e possible ranges and num of decimals for the power)

omritoptix (changes requested):
reviewed second part. 

the only thing missing for me is to define invariants.

x/iro/keeper/settle.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// Bootstrap liquidity pool with the raised DYM and unsold tokens
+func (k Keeper) bootstrapLiquidityPool(ctx sdk.Context, plan types.Plan) error {
+	unallocatedTokens := plan.TotalAllocation.Amount.Sub(plan.SoldAmt)        // assumed > 0, as we enforce it in the Buy function
+	raisedDYM := k.BK.GetBalance(ctx, plan.GetAddress(), appparams.BaseDenom) // assumed > 0, as we enforce it by IRO creation fee
+
+	// send the raised DYM to the iro module
+	err := k.BK.SendCoinsFromAccountToModule(ctx, plan.GetAddress(), types.ModuleName, sdk.NewCoins(raisedDYM))
+	if err != nil {
```
omritoptix, line 78:
```suggestion
	// send the raised DYM to the iro module as it will be used as the pool creator
	err := k.BK.SendCoinsFromAccountToModule(ctx, plan.GetAddress(), types.ModuleName, sdk.NewCoins(raisedDYM))
```

x/iro/keeper/create_plan.go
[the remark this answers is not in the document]
omritoptix: I guess one edge case which is possible is if we created gov proposal with this base-denom for some reason, which than doesn't indicate rollapp id broken invariants. 

ok lets leave it.

x/rollapp/types/rollapp.go
```diff
@@ -144,76 +151,77 @@ func validateInitialSequencer(initialSequencer string) error {
...
 	return nil
 }
 
 func (dm DenomMetadata) Validate() error {
 	if l := len(dm.Base); l == 0 || l > maxDenomBaseLength {
-		return errorsmod.Wrap(gerrc.ErrInvalidArgument, "base denom")
+		return fmt.Errorf("base denom")
 	}
 
 	if l := len(dm.Display); l == 0 || l > maxDenomDisplayLength {
```
omritoptix, line 173: we're missing validations here of the sdk (to prevent from validating again after):

```
if err := sdk.ValidateDenom(m.Base); err != nil {
		return fmt.Errorf("invalid metadata base denom: %w", err)
	}

	if err := sdk.ValidateDenom(m.Display); err != nil {
		return fmt.Errorf("invalid metadata display denom: %w", err)
	}
```

x/iro/types/msgs.go
mtsitrin: it validates the user's inputs. the dummy plan is just to avoid code dup

x/rollapp/keeper/rollapp.go
```diff
@@ -136,21 +155,15 @@ func (k Keeper) SealRollapp(ctx sdk.Context, rollappId string) error {
...
 
-	if rollapp.GenesisInfo.Sealed {
-		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "genesis info already sealed")
-	}
-
-	if !rollapp.GenesisInfoFieldsAreSet() {
-		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "genesis info fields not set")
-	}
-
 	rollapp.GenesisInfo.Sealed = true
```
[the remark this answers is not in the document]
mtsitrin: validated by caller. this is the seal function. renamed

x/iro/keeper/create_plan.go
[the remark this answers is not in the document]
mtsitrin: we're talking here about the auto created denom (`Future token`)
currently we using `future_rollappId` as base denom, and `future_rollappSymbol` for symbol.
so i want to make sure it's valid for every `{rollappId, rollappSymbol}`

x/iro/keeper/create_plan.go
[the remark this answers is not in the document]
mtsitrin: it's custom created denom. u shouldn't proccees if for some reason it's already registered.

x/rollapp/keeper/rollapp.go
mtsitrin: refactored a bit

x/rollapp/transfergenesis/ibc_module.go
```diff
@@ -113,43 +123,107 @@ func (w IBCModule) OnRecvPacket(
...
+	l.Info("Received valid genesis transfer. Registered denom data.")
 
-	// we want to skip delayedack etc because we want the transfer to happen immediately
-	return w.IBCModule.OnRecvPacket(commontypes.SkipRollappMiddlewareContext(ctx), packet, relayer)
+	return ack
+}
+
+// validate genesis transfer amount is the same as in the `iro` plan
+// validate the destAddr is the same as `x/iro` module account address
+func (w IBCModule) validateGenesisTransfer(plan irotypes.Plan, transfer rollapptypes.TransferData, rollappDenom banktypes.Metadata) error {
```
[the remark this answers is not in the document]
mtsitrin: regarding the freeze, i think we should allow the rollapp to run, without working IBC
so it can have a fix upgrade

x/iro/keeper/settle.go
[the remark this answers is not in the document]
mtsitrin: it's not intuitive as we want to validate before `OnRecvAck` and settle afterwards.
the `genesistransfer` middleware and `x/iro` indeed coupled which makes sense IMO
only the iro handles the genesis transfer
I did the hook for transfersEnabled, which might be used by other components

x/iro/keeper/trade.go
```diff
@@ -0,0 +1,207 @@
...
+	// Validate start time started (unless the trader is the owner)
+	if ctx.BlockTime().Before(plan.StartTime) && k.rk.MustGetRollapp(ctx, plan.RollappId).Owner != trader {
+		return errorsmod.Wrapf(types.ErrPlanNotStarted, "planId: %d", plan.Id)
+	}
+
+	return nil
+}
+
+func (k Keeper) chargeTakerFee(ctx sdk.Context, cost sdk.Coin, sender sdk.AccAddress) (sdk.Coin, error) {
+	newAmt, takerFeeCoin := k.calcTakerFee(cost, k.GetParams(ctx).TakerFee)
```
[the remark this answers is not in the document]
mtsitrin: it's actually the sdk.Coin already, not only the amt

x/iro/keeper/create_plan.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil, errors.Join(gerrc.ErrFailedPrecondition, types.ErrInvalidEndTime)
+	}
+
+	// validate incentive plan params
+	incentivesMinParams := m.Keeper.GetParams(ctx).IncentivePlanMinimumParams
+	if req.IncentivePlanParams.NumEpochsPaidOver < incentivesMinParams.NumEpochsPaidOver {
+		return nil, errors.Join(gerrc.ErrFailedPrecondition, errorsmod.Wrap(types.ErrInvalidIncentivePlanParams, "num epochs paid over"))
+	}
+	if req.IncentivePlanParams.StartTimeAfterSettlement < incentivesMinParams.StartTimeAfterSettlement {
+		return nil, errors.Join(gerrc.ErrFailedPrecondition, errorsmod.Wrap(types.ErrInvalidIncentivePlanParams, "start time after settlement"))
```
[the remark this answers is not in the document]
mtsitrin: not sure how it's related. u commented on the incentives params validation.
" rollapp cannot be launched " is checked in `SealGenesisInfoWithLaunchTime`

x/iro/types/bonding_curve.go
mtsitrin: added 
```
const (
	MaxNValue    = 10
	MaxPrecision = 2
)
```

x/iro/keeper/settle.go
```diff
@@ -0,0 +1,151 @@
...
+	_, err = k.ik.CreateGauge(ctx, false, k.AK.GetModuleAddress(types.ModuleName), incentives, distrTo, ctx.BlockTime().Add(time.Hour), 30)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func determineLimitingFactor(unsoldTokens, raisedDYM math.Int, ratio math.LegacyDec) (tokens, dym math.Int) {
+	requiredDYM := unsoldTokens.ToLegacyDec().Mul(ratio).TruncateInt()
```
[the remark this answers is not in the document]
mtsitrin: changed the comment
as the calculations in base denom (e18), this scenario is very extreme

x/rollapp/transfergenesis/ibc_module.go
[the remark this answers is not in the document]
omritoptix: >  i think we should allow the rollapp to run, without working IBC

not sure how it looks like or what will the UI/UX look like in that case.

in essence he can anyway upgrade after the freeze (cause in essence freeze is basically telling the rollapp that if he wants to continue he needs to upgrade to last valid state).

how do you imagine the fix upgrade happening without rollback/freeze given his genesis is corrupted?

x/iro/keeper/trade.go
```diff
@@ -0,0 +1,215 @@
...
+	// validate the IRO have enough tokens to sell
+	// protocol will apply max limit (99.9%) to enforce initial token liquidity
+	maxSellAmt := plan.TotalAllocation.Amount.ToLegacyDec().Mul(AllocationSellLimit).TruncateInt()
+	if plan.SoldAmt.Add(amountTokensToBuy).GT(maxSellAmt) {
+		return types.ErrInsufficientTokens
+	}
+
+	// Calculate cost for buying amountTokensToBuy over fixed price curve
+	cost := plan.BondingCurve.Cost(plan.SoldAmt, plan.SoldAmt.Add(amountTokensToBuy))
+	if !cost.IsPositive() {
```
omritoptix, line 68: imo better to handle this case in the `Cost` function vs on both `buy` and `sell`

x/iro/keeper/trade.go
```diff
@@ -0,0 +1,215 @@
...
+	// Calculate cost over fixed price curve
+	cost := plan.BondingCurve.Cost(plan.SoldAmt.Sub(amountTokensToSell), plan.SoldAmt)
+	if !cost.IsPositive() {
+		return errorsmod.Wrapf(types.ErrInvalidCost, "cost: %s", cost.String())
+	}
+	costCoin := sdk.NewCoin(appparams.BaseDenom, cost)
+
+	totalCost, takerFeeCoin := k.SubtractTakerFee(costCoin, k.GetParams(ctx).TakerFee)
+	if !totalCost.IsPositive() || !takerFeeCoin.IsPositive() {
+		return errorsmod.Wrapf(types.ErrInvalidCost, "totalCost: %s, takerFeeCoin: %s", totalCost.String(), takerFeeCoin.String())
```
omritoptix, line 140: this can be dryed out by  moving to the suggested `ApplyTakerFee` function

x/iro/keeper/trade.go
```diff
@@ -0,0 +1,215 @@
...
+	return sdk.NewCoin(amt.Denom, newAmt), sdk.NewCoin(amt.Denom, takerFeeAmt)
+}
+
+// SubtractTakerFee returns the remaining amount after subtracting the taker fee and the taker fee amount
+// returns (1 - takerFee) * tokenIn, takerFee * tokenIn
+func (k Keeper) SubtractTakerFee(amt sdk.Coin, takerFee sdk.Dec) (sdk.Coin, sdk.Coin) {
+	takerFeeAmt := math.LegacyNewDecFromInt(amt.Amount).Mul(takerFee).TruncateInt()
+	newAmt := amt.Amount.Sub(takerFeeAmt)
+	return sdk.NewCoin(amt.Denom, newAmt), sdk.NewCoin(amt.Denom, takerFeeAmt)
+}
```
omritoptix, line 215: can be dried out:

func (k Keeper) ApplyTakerFee(amt sdk.Coin, takerFee sdk.Dec, isAdd bool) (sdk.Coin, sdk.Coin) {
	feeAmt := math.LegacyNewDecFromInt(amt.Amount).Mul(takerFee).TruncateInt()
	feeCoin := sdk.NewCoin(amt.Denom, feeAmt)
	
	var newAmt sdk.Int
	if isAdd {
		newAmt = amt.Amount.Add(feeAmt)
	} else {
		newAmt = amt.Amount.Sub(feeAmt)
	}
	
	return sdk.NewCoin(amt.Denom, newAmt), feeCoin
}

x/iro/types/iro.pb.go
```diff
@@ -0,0 +1,1663 @@
...
+type Params struct {
+	TakerFee    github_com_cosmos_cosmos_sdk_types.Dec `protobuf:"bytes,1,opt,name=taker_fee,json=takerFee,proto3,customtype=github.com/cosmos/cosmos-sdk/types.Dec" json:"taker_fee"`
+	CreationFee github_com_cosmos_cosmos_sdk_types.Int `protobuf:"bytes,2,opt,name=creation_fee,json=creationFee,proto3,customtype=github.com/cosmos/cosmos-sdk/types.Int" json:"creation_fee"`
+	// Minimum plan duration
+	// when creating a plan, the plan duration should be greater than or equal to this value
+	// plan duration is from the start time to the pre-launch time
+	MinPlanDuration time.Duration `protobuf:"bytes,3,opt,name=min_plan_duration,json=minPlanDuration,proto3,stdduration" json:"min_plan_duration"`
+	// Incentive plan minimum parameters
+	// when creating a plan, the incentive plan parameters should be greater than or equal to this value
+	IncentivePlanMinimumParams IncentivePlanParams `protobuf:"bytes,4,opt,name=incentive_plan_minimum_params,json=incentivePlanMinimumParams,proto3" json:"incentive_plan_minimum_params"`
```
omritoptix, line 44: this name seems very committed to the fact that every param which is incentive related will be a minimum.
I think it's cleaner to simply incroporate the min/max in the nested var names. 

so `MinStartTimeAfterSettlement` and `MinNumEpochsPaidOver` so when we see it also in genesis and queries it's easy to understand this param without the wrapping context of min.

x/iro/types/iro.pb.go
```diff
@@ -0,0 +1,1663 @@
...
+	return IncentivePlanParams{}
+}
+
+type IncentivePlanParams struct {
+	// start_time_after_settlement is the time after IRO settlement when the
+	// distribution of the remaining tokens as incentives will start
+	StartTimeAfterSettlement time.Duration `protobuf:"bytes,1,opt,name=start_time_after_settlement,json=startTimeAfterSettlement,proto3,stdduration" json:"start_time_after_settlement"`
+	// num_epochs_paid_over is the number of total epochs distribution will be
+	// completed over
+	NumEpochsPaidOver uint64 `protobuf:"varint,2,opt,name=num_epochs_paid_over,json=numEpochsPaidOver,proto3" json:"num_epochs_paid_over,omitempty"`
```
omritoptix, line 264: those params descriptions are misleading as they are not de-facto what the plan will enforce, but are boundries (i.e min)

omritoptix (changes requested):
There is a lack of consistency in variables naming in the code when it comes to describing costs as int/dec or with denom (i.e sdk.Coin)

for example:
```
costCoin := sdk.NewCoin(appparams.BaseDenom, cost)
totalCost, takerFeeCoin := k.AddTakerFee(costCoin, k.GetParams(ctx).TakerFee)
```

`costCoin` - refers to amout of coins
`totalCost` - refers to amount of coins 
`takerFeeCoin` - refers to amount of coins 

For me confusing and inconsistent. 

I suggest:
`costAmt` - the int/dec repr (so just a number) 
`cost` - the actual cost of amount + denom . 

so in the following example:

`costCoin` -> `cost`
`totalCost` -> `costWithTakerFee`
`takerFeeCoin` -> `takerFee`

x/rollapp/transfergenesis/ibc_module.go
```diff
@@ -108,43 +119,122 @@ func (w IBCModule) OnRecvPacket(
...
+	return ack
+}
+
+// IsGenesisTransferExpected checks if the genesis transfer is expected for the rollapp
+// if the prelaunch time is set, then genesis transfer is expected as we have IRO plan in place
+func (w IBCModule) IsGenesisTransferExpected(ctx sdk.Context, rollapp *rollapptypes.Rollapp) bool {
+	if rollapp.GenesisState.TransfersEnabled {
+		return false
+	}
+	if rollapp.PreLaunchTime.IsZero() {
```
omritoptix, line 199: wby do we check it implicity vs explicity, i.e a rollapp iwth an unsettled IRO plan which has finished?

x/rollapp/transfergenesis/ibc_module.go
```diff
@@ -77,20 +88,20 @@ func (w IBCModule) logger(
...
-// else:
-//
-//		 transfers must not have already been enabled.
-//	     check it's a valid genesis transfer. If it is, then register the denom
-//
+// OnRecvPacket will,
+// if iro plan exists for this rollapp:
+//	    check it's a valid genesis transfer.
+// 		If it is, then pass the packet, register the denom and settle the plan.
+// In any case, mark the transfers enabled.
```
omritoptix, line 95: why aren't we actually just setting by default on the rollapp `transfers_enabled=true` and only in the case of IRO plan creation change it to `false` vs "opening" the bridge for regualr transfers, which in essence does nothing.

x/rollapp/transfergenesis/ibc_module.go
```diff
@@ -108,43 +119,122 @@ func (w IBCModule) OnRecvPacket(
...
+		l.Error("Enable transfers.", "err", err)
+		return uevent.NewErrorAcknowledgement(ctx, errorsmod.Wrap(err, "transfer genesis: enable transfers"))
+	}
+	return ack
+}
+
+// IsGenesisTransferExpected checks if the genesis transfer is expected for the rollapp
+// if the prelaunch time is set, then genesis transfer is expected as we have IRO plan in place
+func (w IBCModule) IsGenesisTransferExpected(ctx sdk.Context, rollapp *rollapptypes.Rollapp) bool {
+	if rollapp.GenesisState.TransfersEnabled {
```
omritoptix, line 196: why dont we check in the beginning of the OnRecvPacket right after we know it's a rollapp (and if true, contine to next) vs checking if we expect a transfer.

x/iro/types/msgs.go
mtsitrin: not sure I got u. 
refactored anyway

x/rollapp/keeper/rollapp.go
```diff
@@ -116,43 +117,31 @@ func (k Keeper) SetRollapp(ctx sdk.Context, rollapp types.Rollapp) {
...
-		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "rollapp already launched")
-	}
-
+func (k Keeper) SetRollappAsLaunched(ctx sdk.Context, rollapp *types.Rollapp) error {
 	if !rollapp.AllImmutableFieldsAreSet() {
-		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "seal with immutable fields not set")
+		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "immutable fields not set")
 	}
 
+	rollapp.GenesisInfo.Sealed = true
```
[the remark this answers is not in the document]
mtsitrin: validated few lines above
`if !rollapp.AllImmutableFieldsAreSet() `

x/sequencer/keeper/msg_server_create_sequencer.go
```diff
@@ -50,7 +50,13 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
 		if !isInitialOrAllAllowed {
 			return nil, types.ErrNotInitialSequencer
 		}
-		if err := k.rollappKeeper.SealRollapp(ctx, msg.RollappId); err != nil {
+
+		// check pre launch time
+		if rollapp.PreLaunchTime.After(ctx.BlockTime()) {
+			return nil, types.ErrBeforePreLaunchTime
```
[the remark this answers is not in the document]
mtsitrin: if it's not set, the expression is always false

x/iro/keeper/trade.go
```diff
@@ -0,0 +1,207 @@
...
+		return errorsmod.Wrapf(types.ErrInvalidCost, "cost: %s", cost.String())
+	}
+
+	// Validate expected out amount
+	if cost.GT(maxCost) {
+		return errorsmod.Wrapf(types.ErrInvalidExpectedOutAmount, "maxCost: %s, cost: %s", maxCost.String(), cost.String())
+	}
+
+	// Charge taker fee
+	costC, err := k.chargeTakerFee(ctx, sdk.NewCoin(appparams.BaseDenom, cost), buyer)
```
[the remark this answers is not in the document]
mtsitrin: fixed

x/rollapp/keeper/rollapp.go
```diff
@@ -116,43 +117,31 @@ func (k Keeper) SetRollapp(ctx sdk.Context, rollapp types.Rollapp) {
...
-	if !found {
-		return gerrc.ErrNotFound
-	}
-
-	if rollapp.GenesisInfo.Sealed {
-		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "genesis info already sealed")
+// UpdateRollappWithIROPlan seals the rollapp genesis info and sets it pre launch time according to the iro plan end time
+// - GenesisInfo fields must be set
+// - Rollapp must not be Launched
+func (k Keeper) SealGenesisInfoWithLaunchTime(ctx sdk.Context, rollapp *types.Rollapp, preLaunchTime time.Time) error {
```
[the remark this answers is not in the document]
mtsitrin: genesis info sealing logic is `rollapp.GenesisInfo.Sealed = true`
not sure what there's to dry out
`SetRollappAsLaunched` and `SealGenesisInfoWithLaunchTime` have different pre-conditions.

x/rollapp/transfergenesis/ibc_module.go
mtsitrin: so we could detect violation where `TransfersEnabled=true` and we have genesis transfer.
as we discussed it can be removed, I'll refactor

x/rollapp/types/rollapp.go
```diff
@@ -76,20 +79,28 @@ func (r Rollapp) ValidateBasic() error {
...
 }
 
+func (r Rollapp) IsGenesisTransferEnabled() bool {
+	return r.GenesisState.TransfersEnabled
+}
+
 func (r Rollapp) AllImmutableFieldsAreSet() bool {
-	return r.InitialSequencer != ""
+	return r.InitialSequencer != "" || !r.GenesisInfoFieldsAreSet()
 }
```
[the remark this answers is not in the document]
mtsitrin: good catch. added UT

x/rollapp/transfergenesis/ibc_module.go
mtsitrin: it's the same. if IRO exists, preluanch is not zero. if it;s settled, than TransfersEnabled .
added comment to clarify

x/rollapp/keeper/rollapp.go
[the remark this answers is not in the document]
mtsitrin: renamed

x/rollapp/transfergenesis/ibc_module.go
mtsitrin: removed

x/rollapp/transfergenesis/ibc_module.go
mtsitrin: done

x/rollapp/transfergenesis/ibc_module.go
omritoptix: yea I understand it's the same, just don't understand whats the reasoning for checking this vs `GetPlanByRollap` result. I assume there is a reason.

x/sequencer/keeper/msg_server_create_sequencer.go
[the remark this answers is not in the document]
omritoptix: I see. imo this implicitness is confusing. we can add a comment to clarify or alternatively explictly check this in the case this sequencer is launched for rollapp with IRO. 

also thought of suggesting  it would be cleaner to move `preLaunchTime` to only live on the IRO object if it's only IRO related but than checked the IRO object and I see it also has a `preLaunchTime`. and seems like it's a duplicate var living in both places. so I think worth just leaving it on the IRO, change it's name to `EndTime` and make sure a rollapp can't start until `rollapp.iro.EndTime` is over. 

would also address https://github.com/dymensionxyz/dymension/pull/1201#discussion_r1751039489

omritoptix (dismissed):
Some var name suggestions and also a duplicate variable it seems but nothing which is a blocker. 

you can either address now or merge and open a technical debt issue with link to the  comments and address them later.

x/rollapp/keeper/rollapp.go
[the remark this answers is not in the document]
omritoptix: > genesis info sealing logic is rollapp.GenesisInfo.Sealed = true

not really, cause you always want to validate that all pre-requirements are set.
currently the pre-requirements check are delegated to the caller vs just creating a `Seal` function which validates pre-requirements in one place and seals. 

so now you're checking both independetly and sealing in:
1. `SetIROPlanToRollapp`
2. `SetRollappAsLaunched`

the next caller will also need to implement the checks and the sealing vs just calling `Seal`

x/sequencer/keeper/msg_server_create_sequencer.go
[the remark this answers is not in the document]
mtsitrin: actually I think it's better to have `preLaunchTime` as rollapp's field, as it might be enforced for different reasons in the future, not only from IRO.
I left it on the IRO plan object as well for clearity, but it's not used and can be removed from the plan store if u prefer.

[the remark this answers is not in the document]
omritoptix: > actually I think it's better to have preLaunchTime as rollapp's field, as it might be enforced for different reasons in the future, not only from IRO.

I'm fine with that.

> I left it on the IRO plan object as well for clearity, but it's not used and can be removed from the plan store if u prefer.

also fine with that, but imo we need to change it's name to `EndTime` cause I don't see the logic of `Pre-Launch time` on the IRO object, especially with it's comment related to rollapp.

x/iro/types/types.go
```diff
@@ -0,0 +1 @@
+package types
```
VictorTrustyDev, line 1: empty file

PR #1229, opened 2024-09-16
feat: register rollapp amino messages to support EIP712

ItzhakBokris:


danwt (approved):

danwt (commented):
Are you sure it's right Tzachi?

I see maybe some registration is missing like on messages

https://github.com/cosmos/cosmos-sdk/blob/d80078f721428b7482da255f2c62ffb7ccd5beec/x/evidence/types/codec.go#L20


But I'm not sure

PR #1292, opened 2024-10-07
fix(sequencers): incorrect sorting mechanism allows manipulation of proposer selection

### LINKED ISSUE #1291: Incorrect sorting mechanism allows manipulation of proposer selection
We'll enforce **single** denom for the sequencer bond to mitigate this

mtsitrin:
## Description

----

Closes #1291 

This PR does the following:
* better enforcement of single denom for sequencer bond 
* requires `MinBond` param to be > 0 (instead >= 0 currently)

### DIFF
--- a/x/sequencer/keeper/msg_server_create_sequencer.go
+++ b/x/sequencer/keeper/msg_server_create_sequencer.go
@@ -63,28 +63,22 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
 		}
 	}
 
-	bond := sdk.Coins{}
-	if minBond := k.GetParams(ctx).MinBond; !(minBond.IsNil() || minBond.IsZero()) {
-		if msg.Bond.Denom != minBond.Denom {
-			return nil, errorsmod.Wrapf(
-				types.ErrInvalidCoinDenom, "got %s, expected %s", msg.Bond.Denom, minBond.Denom,
-			)
-		}
-
-		if msg.Bond.Amount.LT(minBond.Amount) {
-			return nil, errorsmod.Wrapf(
-				types.ErrInsufficientBond, "got %s, expected %s", msg.Bond.Amount, minBond,
-			)
-		}
-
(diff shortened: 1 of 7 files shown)


x/sequencer/keeper/msg_server_create_sequencer_test.go
```diff
@@ -55,45 +49,63 @@ func (suite *SequencerTestSuite) TestMinBond() {
 			bond:          sdk.NewCoin(bond.Denom, bond.Amount.Quo(sdk.NewInt(2))),
 			expectedError: types.ErrInsufficientBond,
 		},
+		{
+			name:          "wrong bond denom",
```
omritoptix, line 53: also this test seems to test the wrong thing - i.e fail on invalid coin denom vs valid but not what we expected (i.e wrong).

omritoptix (changes requested):

x/sequencer/keeper/msg_server_create_sequencer.go
```diff
@@ -63,28 +63,22 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
...
-		}
-
-		if msg.Bond.Amount.LT(minBond.Amount) {
-			return nil, errorsmod.Wrapf(
-				types.ErrInsufficientBond, "got %s, expected %s", msg.Bond.Amount, minBond,
-			)
-		}
+	// validate bond requirement
+	minBond := k.GetParams(ctx).MinBond
+	if !msg.Bond.IsGTE(minBond) {
```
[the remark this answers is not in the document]
mtsitrin: it's checked in `IsGTE`
```
// IsGTE returns true if they are the same type and the receiver is
// an equal or greater value

x/sequencer/keeper/msg_server_create_sequencer_test.go
mtsitrin: changed the test a bit

omritoptix (approved):

PR #1311, opened 2024-10-11
feat(rollapp): move DRS version to be part of the block descriptor

### LINKED ISSUE #1295: add DRS version for each block descriptor in state info
and replace current DRS version per state info.

srene:
## Description

----

Closes #1295 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


x/rollapp/keeper/msg_server_update_state.go
```diff
@@ -34,17 +34,19 @@ func (k msgServer) UpdateState(goCtx context.Context, msg *types.MsgUpdateState)
 		return nil, errorsmod.Wrap(err, "before update state")
 	}
 
-	// verify the DRS version is not vulnerable
-	if k.IsDRSVersionVulnerable(ctx, msg.DrsVersion) {
-		// the rollapp is not marked as vulnerable yet, mark it now
-		err := k.MarkRollappAsVulnerable(ctx, msg.RollappId)
-		if err != nil {
-			return nil, fmt.Errorf("mark rollapp vulnerable: %w", err)
+	for _, bd := range msg.BDs.BD {
```
omritoptix, line 37: the iteration here can be costly and potentially lead to dos as we currently don't limit the num blocks in a batch.
I think we can simply check the first and last block, cause he chances of DRS version changing in between is highly unlikely and even somehow it did, it's not that interesting to fork the rollap for.

x/rollapp/keeper/msg_server_mark_vulnerable_rollapps.go
```diff
@@ -62,19 +62,23 @@ func (k Keeper) MarkVulnerableRollapps(ctx sdk.Context, drsVersions []string) (i
...
-		if info.DrsVersion == "" {
-			logger.With("rollapp_id", rollapp.RollappId).Info("no DRS version set for rollapp")
-		}
 
-		_, vulnerable := vulnerableVersions[info.DrsVersion]
-		if vulnerable {
-			err := k.MarkRollappAsVulnerable(ctx, rollapp.RollappId)
-			if err != nil {
-				return 0, fmt.Errorf("freeze rollapp: %w", err)
+		for _, bd := range info.BDs.BD {
```
omritoptix, line 66: the iteration here can be costly and potentially lead to dos as we currently don't limit the num blocks in a batch (though it's limited by tx size but still can be very high).
I think we can simply check the first and last block, cause he chances of DRS version changing in between is highly unlikely and even somehow it did, it's not that interesting to fork the rollap for.

omritoptix (changes requested):
unbond iteration

x/rollapp/keeper/msg_server_update_state.go
srene: modified

omritoptix (approved):

x/rollapp/keeper/msg_server_mark_vulnerable_rollapps.go
```diff
@@ -62,19 +62,31 @@ func (k Keeper) MarkVulnerableRollapps(ctx sdk.Context, drsVersions []string) (i
...
 			continue
 		}
-		// TODO: this check may be deleted once empty DRS version is marked vulnerable
-		//  https://github.com/dymensionxyz/dymension/issues/1233
-		if info.DrsVersion == "" {
-			logger.With("rollapp_id", rollapp.RollappId).Info("no DRS version set for rollapp")
+
+		// We only check first and last BD to avoid DoS attack related to iterating big number of BDs (taking into account a state update can be submitted with any numblock value)
+		// It is assumed there cannot be two upgrades in the same state update (since it requires gov proposal), if this happens it will be a fraud caught by Rollapp validators.
+		// Therefore checking first and last BD for deprecated DRS version should be enough.
```
danwt, line 68: IMO this is making things a bit more complicated without a good reason
if there is DOS by submitting a batch with loads of BDS it's a pretty serious problem IMO
So either there is a DOS and we should work out what it affect
or there isn't and it's fine
iterating small structs in memory and checking presence in a hashmap is not particularly expensive

x/rollapp/keeper/msg_server_update_state.go
```diff
@@ -34,17 +34,27 @@ func (k msgServer) UpdateState(goCtx context.Context, msg *types.MsgUpdateState)
...
+		if k.IsDRSVersionVulnerable(ctx, bd.DrsVersion) {
+			// the rollapp is not marked as vulnerable yet, mark it now
+			err := k.MarkRollappAsVulnerable(ctx, msg.RollappId)
+			if err != nil {
+				return nil, fmt.Errorf("mark rollapp vulnerable: %w", err)
+			}
+			k.Logger(ctx).With("rollapp_id", msg.RollappId, "drs_version", bd.DrsVersion).
+				Info("non-frozen rollapp tried to submit MsgUpdateState with the vulnerable DRS version, mark the rollapp as vulnerable")
+			// we must return non-error if we want the changes to be saved
+			return &types.MsgUpdateStateResponse{}, nil
```
danwt, line 56: nit: can this be  DRYer wrt to the mark_vulnerable.. ?
lots of dupe here
also ditto my comment about checking first and last

x/rollapp/keeper/msg_server_mark_vulnerable_rollapps.go
srene: this has been modified so in only checks the last bd, since the one it matters in the case of marking the rollapp as vulnerable, otherwise it means the rollapp already upgraded and is no longer vulnerable

x/rollapp/keeper/msg_server_update_state.go
srene: not sure what you mean by DRYer here...

x/rollapp/keeper/msg_server_mark_vulnerable_rollapps.go
srene: we only check last

x/rollapp/keeper/msg_server_update_state.go
srene: since the iteration does not imply any store access all bds are checked

x/rollapp/keeper/msg_server_update_state.go
danwt: reducing code duplications https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

x/rollapp/keeper/msg_server_update_state.go
```diff
@@ -34,17 +34,19 @@ func (k msgServer) UpdateState(goCtx context.Context, msg *types.MsgUpdateState)
...
-	if k.IsDRSVersionVulnerable(ctx, msg.DrsVersion) {
-		// the rollapp is not marked as vulnerable yet, mark it now
-		err := k.MarkRollappAsVulnerable(ctx, msg.RollappId)
-		if err != nil {
-			return nil, fmt.Errorf("mark rollapp vulnerable: %w", err)
+	for _, bd := range msg.BDs.BD {
+		// verify the DRS version is not vulnerable
+		if k.IsDRSVersionVulnerable(ctx, bd.DrsVersion) {
+			// the rollapp is not marked as vulnerable yet, mark it now
+			err := k.MarkRollappAsVulnerable(ctx, msg.RollappId)
```
danwt, line 41: ah I guess this is a store read
could do a happy path optimization to avoid rechecking ones which are already seenf
but IDK if we want to microoptimize

danwt (approved):
Approved but could micro optimize the state update check if we really want to

omritoptix (approved):

PR #1390, opened 2024-11-04
feat(rollapp): prefixed finalization queue with rollappID

### LINKED ISSUE #631: [rollapp] packet/states in store should be prefixed with rollappId as well
currently, `BlockHeightToFinalizationQueue` prefixed with `creation_height` and holds unsorted array of `stateIdx`

The queue should be prefixed by `height/rollappId`, so all the pending of a specific rollapp can be queried easily 

It also applies to the `DelayedAck` middleware

keruch:
## Description

----

Closes #631 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -141,26 +146,99 @@ func (k *Keeper) finalizePendingState(ctx sdk.Context, stateInfoIndex types.Stat
 	return nil
 }
 
-func (k Keeper) updateQueueForHeight(ctx sdk.Context, blockHeightToFinalizationQueue types.BlockHeightToFinalizationQueue, failedRollapps map[string]uint64) {
-	// remove the blockHeightToFinalizationQueue if all the rollapps' states are finalized
-	if len(failedRollapps) == 0 {
-		k.RemoveBlockHeightToFinalizationQueue(ctx, blockHeightToFinalizationQueue.CreationHeight)
+func (k Keeper) updateQueueForHeight(ctx sdk.Context, queue types.BlockHeightToFinalizationQueue, failedRollapps map[string]uint64) {
+	// remove the queue if all the rollapp states are finalized
```
zale144, line 150: I think this logic needs to be further reworked, in particular updating the queue from the `failedRollapps` map, because previously we had a signle queue with state updates from mixed rollapps, but now it's grouped, so `updateQueueForHeight` will do so for the current queue's rollapp. 

Also, `finalizeQueueForHeight` should be changed to simply stop finalizing at the first failure for the current queue/rollapp.

zale144 (changes requested):

app/upgrades/v4/upgrade.go
```diff
@@ -210,6 +214,51 @@ func migrateStreamer(ctx sdk.Context, sk streamerkeeper.Keeper, ek *epochskeeper
...
+		})
+	}
+
+	// sort by rollappID
+	slices.SortFunc(queues, func(a, b rollapptypes.BlockHeightToFinalizationQueue) int {
+		return strings.Compare(a.RollappId, b.RollappId)
+	})
+
+	return queues
+}
```
danwt, line 260: Is it deterministic?
Should add a comment describing why it's deterministic, if using maps

danwt (changes requested):

x/rollapp/keeper/block_height_to_finalization_queue.go
keruch: thanks for the catch, reworked `FinalizeAllPending`

app/upgrades/v4/upgrade.go
danwt: right, it's plainly obvious, sorry

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -141,26 +156,80 @@ func (k *Keeper) finalizePendingState(ctx sdk.Context, stateInfoIndex types.Stat
...
+	queue, err := k.finalizationQueue.Get(ctx, collections.Join(height, rollappID))
+	if err != nil && !errors.Is(err, collections.ErrNotFound) {
+		panic(err)
+	}
+	found := err == nil
+	return queue, found
+}
+
+// RemoveFinalizationQueue removes types.BlockHeightToFinalizationQueue.
+// Panics on encoding errors. Do not panic if the key does not exist.
```
omritoptix, line 180: seems like we panic here also if not found? don't see where we check the error.

omritoptix (changes requested):
I think we should avoid all panicing in end blockers when it comes to finalization.
the impact of panicing is halting the chain which impacts all rollapps.
in case finalization logic is corrupted, worst case the chain will not finalize states until its fixed, which is imo better than freezing all acticity.

x/rollapp/keeper/block_height_to_finalization_queue.go
```diff
@@ -60,18 +62,21 @@ func (k Keeper) FinalizeRollappStates(ctx sdk.Context) {
 	}
 	// check to see if there are pending  states to be finalized
 	finalizationHeight := uint64(ctx.BlockHeight() - int64(k.DisputePeriodInBlocks(ctx)))
-	pendingFinalizationQueue := k.GetAllFinalizationQueueUntilHeightInclusive(ctx, finalizationHeight)
+	queue, err := k.GetFinalizationQueueUntilHeightInclusive(ctx, finalizationHeight)
+	if err != nil {
```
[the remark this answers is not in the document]
keruch: i believe we cannot completely get rid of all panic in this end blocker, otherwise we would need to add some complex recovery mechanism. imagine the following:
1. we start processing some queue and it has 10 states to finalize
2. we finalize 5 states and fail on 6th (we run finalization in the cached ctx)
3. now we need to store this queue to the state
4. `SaveQueue` fails

we can't just log this and go on since the queue has already finalized states and thus will panic on the next iteration. and we can't make the finalization atomic since we support "partial" finalization (see no.2). to handle this we need to deeply refactor the method.

good news is that all panics there are encoding-related (i double checked!), so they should never happen in practice. i don't think we need to spend a lot of time for refactoring just to avoid these panics since they are not likely to happen (plus we've already had possible encoding panics in this method).

i added detailed reasoning to the lines that may panic.

x/rollapp/keeper/block_height_to_finalization_queue.go
keruch: we don't need an additional check, collections api guarantees that "not found" doesn't panic (i double-checked!)
https://github.com/cosmos/cosmos-sdk/blob/main/collections/map.go#L131

omritoptix (approved):

danwt (commented):
@keruch is this right

```
// RemoveBlockHeightToFinalizationQueue removes a blockHeightToFinalizationQueue from the store
// Deprecated: use RemoveFinalizationQueue instead. Only used in state migrations.
func (k Keeper) RemoveBlockHeightToFinalizationQueue(
	ctx sdk.Context,
	creationHeight uint64,
) {
	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.BlockHeightToFinalizationQueueKeyPrefix))
	store.Delete(types.BlockHeightToFinalizationQueueKey(
		creationHeight,
	))
}
```

does it really delete everything?

danwt (approved):

omritoptix (approved):

PR #1448, opened 2024-11-08
Danwt/fix seq params - DO NOT MERGE

danwt:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


app/upgrades/v4/upgrade.go
```diff
@@ -113,51 +100,6 @@ func migrateModuleParams(ctx sdk.Context, keepers *keepers.AppKeepers) {
...
-			keyTable = evmtypes.ParamKeyTable()
-		case feemarkettypes.ModuleName:
-			keyTable = feemarkettypes.ParamKeyTable()
-		default:
-			continue
-		}
-
-		if !subspace.HasKeyTable() {
-			subspace.WithKeyTable(keyTable)
-		}
```
danwt, line 157: LoadDeprecatedParamsSubspaces is useless and unnecessary
it does nothing because the adjusted subspace isn't passed anywhere
The only place that needs it is the consensus params upgrade but that does it by itself

app/keepers/modules.go
```diff
@@ -268,8 +269,8 @@ var maccPerms = map[string][]string{
 }
 
 var BeginBlockers = []string{
-	epochstypes.ModuleName,
 	upgradetypes.ModuleName,
+	epochstypes.ModuleName,
```
danwt, line 273: Epochs being first led to delayedack and incentives trying to access new params before they have been set in migration

app/keepers/modules.go
```diff
@@ -205,7 +205,8 @@ func (a *AppKeepers) SetupModules(
 		ibctransfer.NewAppModule(a.TransferKeeper),
 		rollappmodule.NewAppModule(appCodec, a.RollappKeeper, a.AccountKeeper, a.BankKeeper),
 		iro.NewAppModule(appCodec, *a.IROKeeper),
-		sequencermodule.NewAppModule(appCodec, a.SequencerKeeper, a.AccountKeeper, a.BankKeeper, a.GetSubspace(sequencertypes.ModuleName)),
+
+		sequencermodule.NewAppModule(appCodec, a.SequencerKeeper, a.AccountKeeper, a.BankKeeper),
```
danwt, line 209: completely deleted sequencer param store migration because we liteerally just set all params to default, so can just do that very simply, without a module specific migration

app/upgrades/v4/upgrade.go
```diff
@@ -205,9 +147,8 @@ func migrateRollappLightClients(ctx sdk.Context, rollappkeeper *rollappkeeper.Ke
 // migrateStreamer creates epoch pointers for all epoch infos and updates module params
 func migrateStreamer(ctx sdk.Context, sk streamerkeeper.Keeper, ek *epochskeeper.Keeper) error {
 	// update module params
-	oldParams := sk.GetParams(ctx)
-	oldParams.MaxIterationsPerBlock = streamertypes.DefaultMaxIterationsPerBlock
-	sk.SetParams(ctx, oldParams)
+	params := streamertypes.DefaultParams()
+	sk.SetParams(ctx, params)
```
danwt, line 151: avoid reading new param before setting it

app/upgrades/v4/upgrade.go
```diff
@@ -271,11 +212,8 @@ func ReformatFinalizationQueue(queue rollapptypes.BlockHeightToFinalizationQueue
 }
 
 func migrateIncentivesParams(ctx sdk.Context, ik *incentiveskeeper.Keeper) {
-	params := ik.GetParams(ctx)
-	defaultParams := incentivestypes.DefaultParams()
-	params.CreateGaugeBaseFee = defaultParams.CreateGaugeBaseFee
-	params.AddToGaugeBaseFee = defaultParams.AddToGaugeBaseFee
-	params.AddDenomFee = defaultParams.AddDenomFee
+	params := incentivestypes.DefaultParams()
+	params.DistrEpochIdentifier = ik.DistrEpochIdentifier(ctx)
```
danwt, line 216: same as what alex did

app/keepers/modules.go
zale144: module specific migration is the preferred way, though

app/upgrades/v4/upgrade.go
```diff
@@ -51,7 +51,46 @@ func CreateUpgradeHandler(
 	return func(ctx sdk.Context, _ upgradetypes.Plan, fromVM module.VersionMap) (module.VersionMap, error) {
 		logger := ctx.Logger().With("upgrade", UpgradeName)
 
-		LoadDeprecatedParamsSubspaces(keepers)
+		for _, subspace := range keepers.ParamsKeeper.GetSubspaces() {
```
zale144, line 54: why move this code here instead of where it was?

app/upgrades/v4/upgrade.go
zale144: outdated comment I believe

zale144 (commented):
I this it's fine but we should still remove all use of GetParamsSet, as I did in my PR, otherwise we will likely see more panics

app/keepers/modules.go
danwt: 🤷

danwt:
@zale144 re I this it's fine but we should still remove all use of GetParamsSet, as I did in my PR, otherwise we will likely see more panics

Why do you think so? They have now been migrated so..

danwt:
other pr is up https://github.com/dymensionxyz/dymension/pull/1459

PR #1455, opened 2024-11-10
feat(sequencer): hardforking when rotating to sentinel

### LINKED ISSUE #1430: force hard fork when going into "halt"
after graceful rotation to sentinel sequencer (no proposer), we should revision bump

mtsitrin:
closes #1430

### DIFF
--- a/x/sequencer/keeper/bond.go
+++ b/x/sequencer/keeper/bond.go
@@ -4,7 +4,6 @@ import (
 	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dymensionxyz/dymension/v3/x/sequencer/types"
-	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 	"github.com/dymensionxyz/sdk-utils/utils/ucoin"
 )
 
@@ -41,17 +40,14 @@ func (k Keeper) TryUnbond(ctx sdk.Context, seq *types.Sequencer, amt sdk.Coin) e
 		return errorsmod.Wrap(err, "refund")
 	}
 	if seq.Tokens.IsZero() {
-		return errorsmod.Wrap(k.unbond(ctx, seq), "unbond")
+		k.unbond(ctx, seq)
 	}
+	k.SetSequencer(ctx, *seq)
 	return nil
 }
 
 // set unbonded status and clear proposer/successor if necessary
-func (k Keeper) unbond(ctx sdk.Context, seq *types.Sequencer) error {
-	if k.IsSuccessor(ctx, *seq) {
-		return gerrc.ErrInternal.Wrap(`unbond next proposer: it shouldnt be possible because
-they cannot do frauds and they cannot unbond gracefully`)
-	}
+func (k Keeper) unbond(ctx sdk.Context, seq *types.Sequencer) {
 	seq.Status = types.Unbonded
 
 	ctx.EventManager().EmitEvent(
@@ -60,10 +56,4 @@ they cannot do frauds and they cannot unbond gracefully`)
 			sdk.NewAttribute(types.AttributeKeySequencer, seq.Address),
 		),
 	)
-	if k.IsProposer(ctx, *seq) {
-		k.SetProposer(ctx, seq.RollappId, types.SentinelSeqAddr)
-		// we assume the current successor will not be happy if the proposer suddenly unbonds
-		k.SetSuccessor(ctx, seq.RollappId, types.SentinelSeqAddr)
-	}
-	return nil
 }
--- a/x/sequencer/keeper/fraud.go
+++ b/x/sequencer/keeper/fraud.go
@@ -19,14 +19,18 @@ func (k Keeper) TryKickProposer(ctx sdk.Context, kicker types.Sequencer) error {
 	if !k.Kickable(ctx, proposer) {
 		return errorsmod.Wrap(gerrc.ErrFailedPrecondition, "not kickable")
 	}
-	// FIXME: slash remaining funds if any?
-	if err := k.unbond(ctx, &proposer); err != nil {
-		return errorsmod.Wrap(err, "unbond")
+
+	// clear the proposer
+	k.SetProposer(ctx, ra, types.SentinelSeqAddr)
+
+	err := k.TryUnbond(ctx, &proposer, proposer.TokensCoin())
+	if err != nil {
+		return errorsmod.Wrap(err, "try unbond")
 	}
 	k.SetSequencer(ctx, proposer)
 
 	// This will call hard fork on the rollapp, which will also optOut all sequencers
-	err := k.hooks.AfterKickProposer(ctx, proposer)
+	err = k.hooks.AfterKickProposer(ctx, proposer)
 	if err != nil {
 		return errorsmod.Wrap(err, "kick proposer callbacks")
 	}
@@ -47,7 +51,7 @@ func (k Keeper) TryKickProposer(ctx sdk.Context, kicker types.Sequencer) error {
 
 	// this will choose kicker as next proposer, since he is the only opted in and bonded
 	// sequencer remaining.
-	if err := k.UpdateProposerIfNeeded(ctx, ra); err != nil {
+	if err := k.RecoverFromSentinelProposerIfNeeded(ctx, ra); err != nil {
 		return errorsmod.Wrap(err, "choose proposer")
 	}
 
@@ -74,24 +78,24 @@ func (k Keeper) SlashLiveness(ctx sdk.Context, rollappID string) error {
 
 // Takes an optional rewardee addr who will receive some bounty
 func (k Keeper) PunishSequencer(ctx sdk.Context, seqAddr string, rewardee *sdk.AccAddress) error {
+	var (
+		rewardMul = sdk.ZeroDec()
+		addr      = []byte(nil)
+	)
+
 	seq, err := k.RealSequencer(ctx, seqAddr)
 	if err != nil {
 		return err
 	}
 
 	if rewardee != nil {
-		rewardMul := sdk.MustNewDecFromStr("0.5") // TODO: parameterise
-		err = k.slash(ctx, &seq, seq.TokensCoin(), rewardMul, *rewardee)
-	} else {
-		err = k.slash(ctx, &seq, seq.TokensCoin(), sdk.ZeroDec(), nil)
-	}
-	if err != nil {
-		return errorsmod.Wrap(err, "slash")
+		rewardMul = sdk.MustNewDecFromStr("0.5") // TODO: parameterise
+		addr = *rewardee
 	}
 
-	err = k.unbond(ctx, &seq)
+	err = k.slash(ctx, &seq, seq.TokensCoin(), rewardMul, addr)
 	if err != nil {
-		return errorsmod.Wrap(err, "unbond")
+		return errorsmod.Wrap(err, "slash")
 	}
 
 	k.SetSequencer(ctx, seq)
--- a/x/sequencer/keeper/fraud.go
+++ b/x/sequencer/keeper/fraud.go
@@ -41,18 +41,18 @@ func (k Keeper) TryKickProposer(ctx sdk.Context, kicker types.Sequencer) error {
 	}
 	k.SetSequencer(ctx, kicker)
 
+	// this will choose kicker as next proposer, since he is the only opted in and bonded
+	// sequencer remaining.
+	if err := k.RecoverFromHalt(ctx, ra); err != nil {
+		return errorsmod.Wrap(err, "choose proposer")
+	}
+
 	if err := uevent.EmitTypedEvent(ctx, &types.EventKickedProposer{
 		Rollapp:  ra,
 		Kicker:   kicker.Address,
 		Proposer: proposer.Address,
 	}); err != nil {
 		return err
-	}
-
-	// this will choose kicker as next proposer, since he is the only opted in and bonded
-	// sequencer remaining.
-	if err := k.RecoverFromSentinelProposerIfNeeded(ctx, ra); err != nil {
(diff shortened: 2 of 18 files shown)


danwt (commented):
TBH to me it's becoming a lot more complicated and coupling x/sequencer with hard fork

I'm not sure why you can't just trigger hard fork in AfterChooseNewProposer hook on rollapp keeper if `after` is sentinel

This is now drifting a lot from the original spec https://www.notion.so/dymension/Rollapp-Hard-Fork-Hub-POV-10aa4a51f86a80a6a1d2ca8fcca9e68b?pvs=4#c310c5683fff48b490a0a5e403aa5fc0

Which had a very clear decoupling, and was IMO, easy and simple

x/sequencer/keeper/proposer.go
```diff
@@ -37,86 +36,47 @@ func (k Keeper) RollappPotentialProposers(ctx sdk.Context, rollappId string) []t
...
 // The result can be the sentinel sequencer.
-func (k Keeper) UpdateProposerIfNeeded(ctx sdk.Context, rollapp string) error {
+func (k Keeper) RecoverFromSentinelProposerIfNeeded(ctx sdk.Context, rollapp string) error {
 	proposer := k.GetProposer(ctx, rollapp)
-	before := proposer
 
 	// a valid proposer is already set so there's no need to do anything
 	if !proposer.Sentinel() {
 		return nil
 	}
```
omritoptix, line 49: I find this function name and inner condition rather confusing as it tries to encapsulate various use cases and do multiple things in one function which make it non reusable. and from some call site (like the kick) I find it cumbersome. 

This function does:
1. check if new proposer is needed
2. set new proposer
3. set new successor 

for (1) - we can check it outside the function based on the callsite
for (2) - I think this all it should do
for (3) - we have the `setSuccessorForRollapp`. isn't it the same? shouldn't we just call it from callsite?


it's clearer to just make this function do one thing (i.e `recoverFromSentinel/UpdateProposer`) and simply check the condition and other work outside this function, where necessary. 

Currently this function is being called from 3 callsites:

1. create sequencer (need to check if current proposer is sentinel outside)
1. update sequencer (need to check if current proposer is sentinel and opted in before call)
1. kick (no need to check if current proposer is sentinel, as we're kicking)

x/sequencer/keeper/proposer.go
```diff
@@ -37,86 +36,47 @@ func (k Keeper) RollappPotentialProposers(ctx sdk.Context, rollappId string) []t
...
+
+	before := proposer
 	successor := k.GetSuccessor(ctx, rollapp)
-	k.SetProposer(ctx, rollapp, successor.Address)
-	k.SetSuccessor(ctx, rollapp, types.SentinelSeqAddr)
-	if k.GetProposer(ctx, rollapp).Sentinel() {
+	// if successor is sentinel, we attempt to find a non sentinel successor
```
omritoptix, line 53: at what point will the successor be not sentinel, if we only reached this part if the proposer is sentinel?

x/sequencer/keeper/proposer.go
```diff
@@ -37,86 +36,47 @@ func (k Keeper) RollappPotentialProposers(ctx sdk.Context, rollappId string) []t
...
+		successor = next
 	}
 
-	after := k.GetProposer(ctx, rollapp)
-	if before.Address != after.Address {
-		k.hooks.AfterChooseNewProposer(ctx, rollapp, before, after)
+	k.SetProposer(ctx, rollapp, successor.Address)
+	k.SetSuccessor(ctx, rollapp, types.SentinelSeqAddr)
+
+	if !successor.Sentinel() {
```
omritoptix, line 66: when will it not be the case given the call sites? (i.e we only call it from create/update/kick)

x/sequencer/keeper/rotation.go
```diff
@@ -84,8 +85,17 @@ func (k Keeper) OnProposerLastBlock(ctx sdk.Context, proposer types.Sequencer) e
 
 	successor := k.GetSuccessor(ctx, rollapp)
 	k.SetProposer(ctx, rollapp, successor.Address)
+
 	k.SetSuccessor(ctx, rollapp, types.SentinelSeqAddr)
 
+	// if proposer is sentinel, prepare new revision for the rollapp
+	if successor.Sentinel() {
+		err := k.rollappKeeper.HardForkToLatest(ctx, rollapp)
```
omritoptix, line 93: why do we need to hard fork here? can't we just hard fork when the new sequencer bonds? otherwise we gonna bump 2 revisions when the sequencer will come. 

if we don't hard fork here than already in the `RecoverFromHaltIfNeeded` should cover it no?

x/sequencer/keeper/rotation.go
```diff
@@ -95,3 +105,44 @@ func (k Keeper) OnProposerLastBlock(ctx sdk.Context, proposer types.Sequencer) e
 	)
 	return nil
 }
+
+// setSuccessorForRollapp will assign a successor. It won't replace an existing one.
+// It will prioritize non sentinel
+func (k Keeper) setSuccessorForRollapp(ctx sdk.Context, rollapp string) error {
```
omritoptix, line 111: this seems like a repeat of part of the `RecoverFromHaltIfNeeded` function.
also I suggest for standartization `setSuccessorForRollapp` -> `chooseSuccessor`

x/sequencer/keeper/rotation.go
```diff
@@ -95,3 +105,44 @@ func (k Keeper) OnProposerLastBlock(ctx sdk.Context, proposer types.Sequencer) e
...
+	successor := k.GetSuccessor(ctx, rollapp)
+	if !successor.Sentinel() {
+		// a valid successor is already set so there's no need to do anything
+		return nil
+	}
```
omritoptix, line 116: as to my other comment of the `RecoverFromHaltIfNeeded` would sugget to remove outside the function.

x/sequencer/keeper/rotation.go
```diff
@@ -95,3 +105,44 @@ func (k Keeper) OnProposerLastBlock(ctx sdk.Context, proposer types.Sequencer) e
...
+}
+
+// ProposerChoiceAlgo : choose the one with most bond
+// Requires sentinel to be passed in, as last resort.
+func ProposerChoiceAlgo(seqs []types.Sequencer) (types.Sequencer, error) {
+	if len(seqs) == 0 {
+		return types.Sequencer{}, gerrc.ErrInternal.Wrap("seqs must at least include sentinel")
+	}
+	// slices package is recommended over sort package
+	slices.SortStableFunc(seqs, func(a, b types.Sequencer) int {
```
omritoptix, line 134: shouldn't we filter out the opted-out one?

omritoptix (changes requested):

x/sequencer/keeper/rotation.go
mtsitrin: it's filtered out in `RollappPotentialProposers`

x/sequencer/keeper/proposer.go
mtsitrin: refactored

x/sequencer/keeper/proposer.go
mtsitrin: yeah not sure
refactored

x/sequencer/keeper/proposer.go
mtsitrin: refactored

mtsitrin:
> TBH to me it's becoming a lot more complicated and coupling x/sequencer with hard fork
> 
> I'm not sure why you can't just trigger hard fork in AfterChooseNewProposer hook on rollapp keeper if `after` is sentinel
> 
> This is now drifting a lot from the original spec https://www.notion.so/dymension/Rollapp-Hard-Fork-Hub-POV-10aa4a51f86a80a6a1d2ca8fcca9e68b?pvs=4#c310c5683fff48b490a0a5e403aa5fc0
> 
> Which had a very clear decoupling, and was IMO, easy and simple

pls be more specific
what I did was:
* when rollapp left with no proposer, prepare a new revision.
it allows the L2 to sync to the **expected revision** and just wait for a new proposer
* when new proposer bonds after halt, we only need to modify the last state info with `nextProposer`

-----
we have the fraud proposal which leaves the rollapp with bumped revision and no proposer.
I've tried to have the other flows act similar

x/sequencer/keeper/msg_server_update.go
```diff
@@ -56,8 +55,11 @@ func (k msgServer) UpdateOptInStatus(goCtx context.Context,
...
 
 	// maybe set as proposer if one is needed
-	if err := k.UpdateProposerIfNeeded(ctx, seq.RollappId); err != nil {
-		return nil, errorsmod.Wrap(err, "choose proposer")
+	proposer := k.GetProposer(ctx, seq.RollappId)
+	if proposer.Sentinel() {
+		if err := k.RecoverFromHalt(ctx, seq.RollappId); err != nil {
+			return nil, err
+		}
 	}
```
danwt, line 63: The original chooseProposer function did exactly this. It chose one if one is needed. Not sure why it's factored differently now

x/sequencer/keeper/rotation.go
```diff
@@ -95,3 +105,39 @@ func (k Keeper) OnProposerLastBlock(ctx sdk.Context, proposer types.Sequencer) e
 	)
 	return nil
 }
+
+// setSuccessorForRotatingRollapp will assign a successor to the rollapp.
+// It will prioritize non sentinel
+// called when a proposer has finished their notice period.
+func (k Keeper) setSuccessorForRotatingRollapp(ctx sdk.Context, rollapp string) error {
```
danwt, line 112: can't this just be `chooseSuccessor`

x/sequencer/keeper/msg_server_create.go
```diff
@@ -106,8 +106,11 @@ func (k msgServer) CreateSequencer(goCtx context.Context, msg *types.MsgCreateSe
 		return nil, err
 	}
 
-	if err := k.UpdateProposerIfNeeded(ctx, msg.RollappId); err != nil {
-		return nil, err
+	proposer := k.GetProposer(ctx, msg.RollappId)
+	if proposer.Sentinel() {
+		if err := k.RecoverFromHalt(ctx, msg.RollappId); err != nil {
+			return nil, err
+		}
```
danwt, line 113: ditto my other comment
This was already encapsulated in the old chooseProposer func, not sure why it's factored differently now

x/sequencer/keeper/bond.go
```diff
@@ -41,17 +40,14 @@ func (k Keeper) TryUnbond(ctx sdk.Context, seq *types.Sequencer, amt sdk.Coin) e
 		return errorsmod.Wrap(err, "refund")
 	}
 	if seq.Tokens.IsZero() {
-		return errorsmod.Wrap(k.unbond(ctx, seq), "unbond")
+		k.unbond(ctx, seq)
 	}
+	k.SetSequencer(ctx, *seq)
```
danwt, line 45: why add this?
The function is taking pointer to sequencer, it should be caller responsibility to set it

danwt (commented):
Left a few concrete comments. 

But I do have reservations about the factoring and architecture here now. In my opinion, things are now less dry, and there are more unique paths through the code, which makes things more difficult

Ive tried to illustrate how I can see things working, which is closer with the original flow as described in the research. Maybe it's interesting for you. LMK

image

x/sequencer/keeper/proposer.go
```diff
@@ -37,84 +37,43 @@ func (k Keeper) RollappPotentialProposers(ctx sdk.Context, rollappId string) []t
...
 
-// UpdateProposerIfNeeded will assign a proposer to the rollapp. It won't replace the incumbent proposer
-// if they are not sentinel. Otherwise it will prioritise a non sentinel successor. Finally, it
-// choose one based on an algorithm.
-// The result can be the sentinel sequencer.
-func (k Keeper) UpdateProposerIfNeeded(ctx sdk.Context, rollapp string) error {
+// RecoverFromHalt will assign a new proposer to the rollapp.
+// It will choose a new proposer from the list of potential proposers.
+// The rollapp must
+func (k Keeper) RecoverFromHalt(ctx sdk.Context, rollapp string) error {
```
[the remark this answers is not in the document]
mtsitrin: it's supposed to be  called only when `proposer == nil` and `successor != nil`
so it's recover from halt

x/sequencer/keeper/msg_server_update.go
mtsitrin: for clarity and readability. by @omritoptix 's suggestion
` if one is needed.` checked by the caller

x/sequencer/keeper/rotation.go
mtsitrin: it's called only in rotation flow, so I thought it will be more clear

x/sequencer/keeper/bond.go
mtsitrin: this function does different stuff (e.g refund)
seems wrong to complete the function without storing the changes on the seq object
wdyt?
I see now that I have duplicate call to `k.SetSequencer(ctx, *seq)` by the caller as well

x/sequencer/keeper/fraud.go
```diff
@@ -19,14 +19,18 @@ func (k Keeper) TryKickProposer(ctx sdk.Context, kicker types.Sequencer) error {
...
-	if err := k.unbond(ctx, &proposer); err != nil {
-		return errorsmod.Wrap(err, "unbond")
+
+	// clear the proposer
+	k.SetProposer(ctx, ra, types.SentinelSeqAddr)
+
+	err := k.TryUnbond(ctx, &proposer, proposer.TokensCoin())
+	if err != nil {
+		return errorsmod.Wrap(err, "try unbond")
 	}
```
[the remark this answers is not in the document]
mtsitrin: I'll rethink and add comment on the main PR

PR #1467, opened 2024-11-13
refactor(lightclient):  removed the need for next stateInfo for valiadition

mtsitrin:
DRY out the validation methods into 2:
`ValidateHeaderAgainstStateInfo` - validates LC updated against existing state info
`ValidateStateInfoAgainstConsensusStates` - validates State Info against existing consensus states

main changes:
* on `HandleMsgUpdateClient` 
  * refactor validation to use single state info, instead of checking state info for H+1

* on `SetCanonicalClient`
  * changed iteration key: instead of iterating all the consensus states, we iterate over state infos
  * each state info is validated


x/lightclient/keeper/canonical_client.go
```diff
@@ -78,56 +87,67 @@ func (k Keeper) validClient(ctx sdk.Context, clientID string, cs exported.Client
...
 	atLeastOneMatch := false
-	for _, consensusHeight := range res.ConsensusStateHeights {
-		h := consensusHeight.GetRevisionHeight()
-		if maxHeight < h {
-			break
+	csStore := k.ibcClientKeeper.ClientStore(ctx, clientID)
+	var err error
+	IterateConsensusStateDescending(csStore, func(h exported.Height) bool {
+		// skip future heights
+		if h.GetRevisionHeight() >= maxHeight {
```
danwt, line 99: >
```suggestion
		if h.GetRevisionHeight() > maxHeight {
```

x/lightclient/keeper/canonical_client.go
```diff
@@ -78,56 +87,67 @@ func (k Keeper) validClient(ctx sdk.Context, clientID string, cs exported.Client
...
+			return true // break
 		}
+
 		atLeastOneMatch = true
-	}
+		return true // break
+	})
 	// Need to be sure that at least one consensus state agrees with a state update
 	// (There are also no disagreeing consensus states. There may be some consensus states
 	// for future state updates, which will incur a fraud if they disagree.)
```
danwt, line 128: This comment is no longer valid since you dont check against states less than the state info, and they wont be frauded

x/lightclient/keeper/hook_listener.go
```diff
@@ -41,9 +36,12 @@ func (hook rollappHook) AfterUpdateState(
...
-		client, ok = hook.k.GetProspectiveCanonicalClient(ctx, rollappId, stateInfo.GetLatestHeight()-1)
+		client, ok = hook.k.FindMatchingClient(ctx, stateInfo)
 		if ok {
 			hook.k.SetCanonicalClient(ctx, rollappId, client)
+			if err := hook.k.PruneSignersBelow(ctx, client, stateInfo.GetLatestHeight()+1); err != nil {
+				return errorsmod.Wrap(err, "prune signers")
+			}
 		}
 		return nil
 	}
```
danwt, line 47:
```suggestion
			
		}
		
	}
```

Removing the `return nil` : makes sure people are not stuck with unbond restriction

Removing: prune signers below: it's called below too, so no need

danwt (changes requested):
The cleanups are great, but the bit about iterating descending doesn't make sense.

https://github.com/dymensionxyz/dymension/blob/1deed9f69af08bac45d50514cf3d9ce5e1e4bff9/x/lightclient/keeper/canonical_client.go#L97-L125

You're only checking the consensus states with heights in the state info, which isn't right. I'm sorry I should've thought about it more when you initially asked me on slack.

Now it's not secure; why? Because there can be invalid consensus states below the ones you check, which are opening a security hole by having a wrong nextValidatorsHash.

We need to iterate through all the consensus states and check that they're valid. In fact, we just need to check that there is a valid _prefix_ . The reason it was implemented the way it was to begin with was because it's just simpler: if we check all up to and including the ones in the state info, then (in the previous design where a mismatch in state update would automatically trigger fraud handler) then we would be guaranteed to catch a wrong one.

I would really suggest to just keep it as it was before. You can still change it to be an iterator though (I think why Spoorthi made it fetch all before was because iterator was not publicly exposed). 

The rest of the PR is good

x/rollapp/types/state_info.go
```diff
@@ -68,9 +72,16 @@ func (s *StateInfo) GetLatestBlockDescriptor() BlockDescriptor {
 	return s.BDs.BD[len(s.BDs.BD)-1]
 }
 
+func (s *StateInfo) NextSequencerForHeight(height uint64) string {
+	if height != s.GetLatestHeight() {
+		return s.Sequencer
+	}
+	return s.NextProposer
+}
+
```
danwt, line 81:
```suggestion
func (s *StateInfo) Successor(height uint64) string {
	if height != s.GetLatestHeight() {
		return s.Sequencer
	}
	return s.NextProposer
}

```

danwt (approved):

danwt:
Great job
Not from your PR but I'm not a fan of this mutable nextSequencer field on the state info

mtsitrin:
> Great job Not from your PR but I'm not a fan of this mutable nextSequencer field on the state info

agree
it was added for dymint validation process

PR #1521, opened 2024-11-20
fix(app): fix missing tx encoder

### LINKED ISSUE #1159: Panic running locally (app.txEncoder is nil)
I rebooted a local hub instance and saw this
`ERR panic recovered in PrepareProposal height=344 module=server panic="runtime error: invalid memory address or nil pointer dereference"`
I lost the rest of the logs unfortunately

294d5bcdd238567ca3e87e913f1cc669198a43cd

mtsitrin:
closes #1159

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -143,6 +143,7 @@ func New(
 	bApp.SetCommitMultiStoreTracer(traceStore)
 	bApp.SetVersion(version.Version)
 	bApp.SetInterfaceRegistry(interfaceRegistry)
+	bApp.SetTxEncoder(encodingConfig.TxConfig.TxEncoder())
 
 	app := &App{
 		BaseApp:           bApp,


app/app.go
```diff
@@ -143,6 +143,7 @@ func New(
 	bApp.SetCommitMultiStoreTracer(traceStore)
 	bApp.SetVersion(version.Version)
 	bApp.SetInterfaceRegistry(interfaceRegistry)
+	bApp.SetTxEncoder(encodingConfig.TxConfig.TxEncoder())
```
omritoptix, line 146: I see it's been set few lines above?

bApp := baseapp.NewBaseApp(appparams.Name, logger, db, encodingConfig.TxConfig.TxDecoder(), baseAppOptions...)

omritoptix (approved):

Issue #1577, opened 2024-11-29
change min bond to be relative to IRO funds raised in case such exists


omritoptix:
currently we enfoce 100DYM min bond.
this makes sense for rollapps without IRO or rollapps who raised small amount.
however in case of IRO raising significant funds, this is not enough to prevent attempt of malicious behavior. 
We want to change it to be min(IRO_DYM_RAISED/bond_factor, 100DYM) where by default bond_factor = 100.

for example, of IRO_DYM_RAISED = 100kDYM, the min bond will be 1k DYM.
similariliy if IRO_DYM_RAISED = 9KDYM, min bond will be 100DYM.

This bond factor should be a param.

omritoptix:
updated impl to make the bond per rollapp with enforced min but it can be chosen by rollapp creator

PR #1579, opened 2024-11-29
feat(rollapp): min bond is now defined on rollapp level

danwt:
## Description

----

Closes https://github.com/dymensionxyz/dymension/issues/1577

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


x/rollapp/keeper/rollapp.go
```diff
@@ -39,6 +39,10 @@ func (k Keeper) CheckAndUpdateRollappFields(ctx sdk.Context, update *types.MsgUp
 		current.InitialSequencer = update.InitialSequencer
 	}
 
+	if 0 < update.MinSequencerBond {
```
omritoptix, line 42: we should enforce a minimum I blv

proto/dymensionxyz/dymension/rollapp/rollapp.proto
```diff
@@ -53,6 +53,8 @@ message Rollapp {
   // register and serve for this rollapp. if left empty, no sequencer is allowed
   // to register. if set to "*" any sequencer can register.
   string initial_sequencer = 13;
+  // how much DYM a sequencer needs (at least) to be proposer
```
omritoptix, line 56: is this field set in base denom (i.e adym) or DYM?

x/sequencer/keeper/params.go
```diff
@@ -19,9 +19,6 @@ func (k Keeper) ValidateParams(ctx sdk.Context, params types.Params) error {
 	if err != nil {
 		return errorsmod.Wrapf(gerrc.ErrInternal, "failed to get base denom: %v", err)
 	}
```
omritoptix, line 21: I think we should still keep the min bond as a general min bond in the system. we can always set it to 0 if we wish.

x/rollapp/keeper/rollapp.go
omritoptix: if we want we can always set it to 0 but should be param like before

omritoptix (changes requested):

omritoptix (changes requested):
nice work! 
however I blv we should keep the min bond.
we can always set it to 0 at some point if we wish but I think it's good to start with sane min bond to protect unexperienced creators and maybe will change it later on.

x/rollapp/keeper/rollapp.go
danwt: fixed
0 would be weird because of kick threshold I think

proto/dymensionxyz/dymension/rollapp/rollapp.proto
danwt: its dym
please dont make me change it

VictorTrustyDev (approved):
utACK LGTM!

x/rollapp/types/message_update_rollapp.go
```diff
@@ -85,7 +87,7 @@ func (msg *MsgUpdateRollappInformation) ValidateBasic() error {
 }
 
 func (msg *MsgUpdateRollappInformation) UpdatingImmutableValues() bool {
-	return msg.InitialSequencer != ""
+	return msg.InitialSequencer != "" || 0 < msg.MinSequencerBond
```
mtsitrin, line 90: rollapp min bond can't be changed post launch?

x/sequencer/types/bond.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	commontypes "github.com/dymensionxyz/dymension/v3/x/common/types"
+	"github.com/dymensionxyz/sdk-utils/utils/ucoin"
+)
+
+var (
+	BondDenom = commontypes.DYMCoin.Denom
+
+	// for tests, real value is supplied by rollapp keeper
```
mtsitrin, line 11: why u need it in the `types` package if it's for tests?
put it in some test file

x/sequencer/keeper/funds.go
```diff
@@ -3,28 +3,31 @@ package keeper
...
+func validBondDenom(c sdk.Coin) error {
+	if c.Denom != types.BondDenom {
+		return errorsmod.Wrapf(types.ErrInvalidDenom, "expect: %s", types.BondDenom)
 	}
 	return nil
 }
 
-func (k Keeper) sufficientBond(ctx sdk.Context, c sdk.Coin) error {
-	if err := k.validBondDenom(ctx, c); err != nil {
+func (k Keeper) sufficientBond(ctx sdk.Context, rollapp string, c sdk.Coin) error {
```
mtsitrin, line 24: do we need some validation in regard of the kick threshold?
will be weird to have a kick threshold higher than the bond

proto/dymensionxyz/dymension/rollapp/rollapp.proto
mtsitrin: @danwt why make it different than other params? they all adym.
now u also need this on the caller side:
```
func (k Keeper) minBond(ctx sdk.Context, rollapp string) sdk.Coin {
	nDym := k.rollappKeeper.MinBond(ctx, rollapp)
	ret := commontypes.DYMCoin
	ret.Amount = ret.Amount.Mul(nDym)
	return ret
}
```
even better, it should be `sdk.Coin`, as in other relevant params (kick threshold e.g)

x/rollapp/types/message_update_rollapp.go
omritoptix: that's a great question.

produce wise I am still not convinced what's the way to go so prefer to keep it more conservative.

I'd say that letting them change it after launch have a few caveats:
1. edge cases where we break the current assumption that bond doesn't change after launch (not sure what bugs it can introduce)

2. a lot of power to the rollapp creator **after** launch where he can e.g dos the rollapp (Change it to so high at some point where no one can bond) or alternatively want to cheat so reduce the bond to some non proportinal amount where it's worth it's while.

The gist is - as a user it's easier to agree on the "terms" of the game before you opt-in.

future plans we can make it dynamic but imo requires more thought so wouldn't wanna venture into it currently.

x/rollapp/client/cli/tx_create_rollapp.go
```diff
@@ -22,6 +22,7 @@ func CmdCreateRollapp() *cobra.Command {
 		dymd tx rollapp create-rollapp myrollapp_12345-1 RollappAlias EVM 
 		// optional flags:
 		--init-sequencer '<seq_address1>,<seq_address2>'
+        --min-sequencer-bond 100
```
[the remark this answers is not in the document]
danwt: I dont' see this on my version

x/sequencer/types/bond.go
danwt: because it's used by apptesting

x/sequencer/keeper/funds.go
danwt: We just be careful
I'm not about to add cross module param verification

x/rollapp/keeper/bond.go
```diff
@@ -0,0 +1,21 @@
...
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
+)
+
+// Rollapp must exist - returns base units, not atto. E.g. 100 dym not 10e18 adym
+func (k *Keeper) MinBond(ctx sdk.Context, rollappID string) math.Int {
+	ra := k.MustGetRollapp(ctx, rollappID)
+	return math.NewIntFromUint64(ra.MinSequencerBond)
+}
```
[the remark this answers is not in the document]
danwt: In what sense @VictorTrustyDev ?

danwt:
@mtsitrin I'm considering your comments to be nits so I haven't changed anything

proto/dymensionxyz/dymension/rollapp/rollapp.proto
danwt: reworked

mtsitrin (dismissed):

omritoptix (approved):

PR #1592, opened 2024-12-01
feat(iro): IRO simulation tests

mtsitrin:
Generate:
* **Allocation amount**: between `100_000` and `1_000_000_000` RA tokens
* random `soldAmt` (to be on different X on the curve)
* Bonding Curve:	
  * Generate 0.5 < N < 1.5
  * M set by `targetRaiseDYM` between 10K and 100M DYM

Operation:
Try to buy the following amount of tokens:
```
testAmounts := []math.Int{
	math.NewInt(1).MulRaw(1e17),      // 0.1 token
	math.NewInt(1).MulRaw(1e18),      // 1 token
	math.NewInt(100).MulRaw(1e18),    // 100 tokens
	math.NewInt(1000).MulRaw(1e18),   // 1000 tokens
	math.NewInt(10000).MulRaw(1e18),  // 10000 tokens
	math.NewInt(100000).MulRaw(1e18), // 100000 tokens
}
```
the simulation validates that:
* manages to buy some of the amounts
* cost is increasing when amount increases
* `buyByToken` and `buyByDYM` are approx equal with 6% tolerance


x/iro/simulation/operations.go
```diff
@@ -0,0 +1,153 @@
...
+		}
+
+		if !managed {
+			err := fmt.Errorf("%s: no valid cost found for any of the test amounts", curveDesc)
+			return simtypes.NoOpMsg(types.ModuleName, "TestBondingCurve", err.Error()), nil, err
+		}
+
+		return simtypes.NewOperationMsg(&types.MsgBuy{}, true, fmt.Sprintf("%s Results:\n%s", curveDesc, results), cdc), nil, nil
+	}
+}
```
danwt, line 131: This looks like an invariant or randomized test, it's not a simulated operation?
I'm not sure I understand the approach here. Seems like a very weird use of sdk simulation

mtsitrin: used the framework as simulation /  fuzzing framework
It's too much effort to prepare all the preconditions for actual Buy/Sell Msgs

used the simulation framework to generate different curves, at simulate different cost functions

danwt: Got you. There are much better tool for that like go stdlib fuzz or rapid FYI

PR #1619, opened 2024-12-05
fix(sequencer): do not allow setting sentinel object

danwt:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:

### DIFF
--- a/x/sequencer/keeper/get_and_set.go
+++ b/x/sequencer/keeper/get_and_set.go
@@ -17,6 +17,9 @@ import (
 // SetSequencer : write to store indexed by address, and also by status
 // Note: do not call with sentinel sequencer
 func (k Keeper) SetSequencer(ctx sdk.Context, seq types.Sequencer) {
+	if seq.Sentinel() {
+		panic("set sentinel proposer")
+	}
 	store := ctx.KVStore(k.storeKey)
 	b := k.cdc.MustMarshal(&seq)
 	store.Set(types.SequencerKey(seq.Address), b)
--- a/x/sequencer/keeper/get_and_set.go
+++ b/x/sequencer/keeper/get_and_set.go
@@ -17,9 +17,6 @@ import (
 // SetSequencer : write to store indexed by address, and also by status
 // Note: do not call with sentinel sequencer
 func (k Keeper) SetSequencer(ctx sdk.Context, seq types.Sequencer) {
-	if seq.Sentinel() {
-		panic("set sentinel proposer")
-	}
 	store := ctx.KVStore(k.storeKey)
 	b := k.cdc.MustMarshal(&seq)
 	store.Set(types.SequencerKey(seq.Address), b)
--- a/x/sequencer/keeper/invariants.go
+++ b/x/sequencer/keeper/invariants.go
@@ -16,6 +16,7 @@ var invs = uinv.NamedFuncsList[Keeper]{
 	{Name: "hash-index", Func: InvariantProposerAddrIndex},
 	{Name: "status", Func: InvariantStatus},
 	{Name: "tokens", Func: InvariantTokens},
+	{Name: "do-not-expose-sentinel", Func: InvariantDoNotExposeSentinel},
 }
 
 // RegisterInvariants registers the sequencer module invariants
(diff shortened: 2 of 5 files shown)


x/sequencer/keeper/get_and_set.go
```diff
@@ -17,6 +17,9 @@ import (
 // SetSequencer : write to store indexed by address, and also by status
 // Note: do not call with sentinel sequencer
 func (k Keeper) SetSequencer(ctx sdk.Context, seq types.Sequencer) {
+	if seq.Sentinel() {
+		panic("set sentinel proposer")
+	}
```
danwt, line 22: I really feel this is a good thing to have

x/sequencer/keeper/proposer.go
```diff
@@ -12,6 +12,9 @@ import (
 
 func (k Keeper) abruptRemoveProposer(ctx sdk.Context, rollapp string) {
 	proposer := k.GetProposer(ctx, rollapp)
+	if proposer.Sentinel() {
+		return
+	}
```
danwt, line 17: this was the bug
Since kick scenario will clear the proposer then hard fork then try to clear it again (via hook)

x/sequencer/keeper/msg_server_kick_proposer_test.go
```diff
@@ -1,13 +1,14 @@
...
 import (
+	rollapptypes "github.com/dymensionxyz/dymension/v3/x/rollapp/types"
 	"github.com/dymensionxyz/dymension/v3/x/sequencer/types"
 	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 	"github.com/dymensionxyz/sdk-utils/utils/utest"
 )
 
 func (s *SequencerTestSuite) TestKickProposerBasicFlow() {
-	s.App.RollappKeeper.SetHooks(nil)
+	s.App.RollappKeeper.SetHooks(rollapptypes.NewMultiRollappHooks(s.k().RollappHooks()))
```
danwt, line 11: I've made sure that the new panic is triggered, when the new bug fix is not included. To do that, hooks are needed.

x/sequencer/keeper/get_and_set.go
omritoptix: problem is that it can be called in an end blocker (like from gov fraud proposal) which will halt the chain.
not sure about it. why not just return an err?

omritoptix (changes requested):
don't think we should panic since it can be called from end blocker (e.g gov)

x/sequencer/keeper/get_and_set.go
danwt: We could of course, but it really complicates the API. Also, I'm not sure how I'd handle the error.

Everything is horribly wrong if this error occurs. Also, I think we can class invariant breaking errors in two

1. It will be ok if the hub continues progressing while the invariant is broken. We will be able to fix it later
2. Once the invariant is broken, the behaviour can become more and more corrupted over time

I think this is case 2. I really don't want to be in a situation where this happened, but we didn't notice. In principle we'd have to recover the hub somehow, and fix corrupted state, but it could have been corrupted for a long time already.

I do think panics are warranted tbh

omritoptix: problem is - does it makes sense to take down an entire hub which serves hundreds of rollapps for a corruption in one? 

also not sure the implications are very tragic. as long as it couldn't lead to direct loss of funds not sure it's worth it.

also in the playground we had it now and it didn't seem like it caused any tragic non-recoverable state (also happened just after hard fork which is an edge case). 

if we had this code now, do you think it would worth to halt the playground for a hard-fork gone wrong? I'm not sure.

danwt:
Goes with https://github.com/dymensionxyz/dymension/pull/1624/files

x/sequencer/keeper/get_and_set.go
danwt: > problem is - does it makes sense to take down an entire hub which serves hundreds of rollapps for a corruption in one?

It's not one though, since rollapps share the sentinel sequencer

danwt: IMO it's like out of bounds access in c++ or something, it's really hard to say what could go wrong

omritoptix (approved):

x/sequencer/keeper/invariants.go
```diff
@@ -168,3 +169,25 @@ func checkSeqTokens(seq types.Sequencer) error {
 	}
 	return nil
 }
+
+// sentinel should not be available in the global index or rollapp wise index
+// (it's only available in getProposer or getSuccessor)
+func InvariantDoNotExposeSentinel(k Keeper) uinv.Func {
```
mtsitrin, line 175: this logic should be covered by UT

mtsitrin (approved):

PR #1623, opened 2024-12-05
feat(liveness): improve tests

danwt:
Add a test to check 100000000% that it's not possible to schedule event in the past

### DIFF
--- a/x/rollapp/keeper/invariants.go
+++ b/x/rollapp/keeper/invariants.go
@@ -291,17 +291,28 @@ func LivenessEventInvariant(k Keeper) sdk.Invariant {
 				msg += fmt.Sprintf("| rollapp stored event but wrong number found in queue: rollapp: %s: event height: %d: found: %d", ra.RollappId, ra.LivenessEventHeight, cnt)
 			}
 		}
-		for _, e := range k.GetLivenessEvents(ctx, nil) {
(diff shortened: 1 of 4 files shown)


x/rollapp/keeper/invariants.go
```diff
@@ -291,17 +291,28 @@ func LivenessEventInvariant(k Keeper) sdk.Invariant {
 				msg += fmt.Sprintf("| rollapp stored event but wrong number found in queue: rollapp: %s: event height: %d: found: %d", ra.RollappId, ra.LivenessEventHeight, cnt)
 			}
 		}
-		for _, e := range k.GetLivenessEvents(ctx, nil) {
+		evts := k.GetLivenessEvents(ctx, nil)
+		seen := make(map[string]struct{})
+		for i, e := range evts {
+			if 0 < i && e.HubHeight < evts[i-1].HubHeight {
+				broken = true
+				msg += fmt.Sprintf("| events not sorted by height: event: %v\n", e)
```
mtsitrin, line 299: is it guaranteed to be sorted by height? is it even an invariant?
it has no effect on the app

mtsitrin (approved):

x/rollapp/keeper/invariants.go
danwt: It is, there is a query to find in range

PR #1639, opened 2024-12-10
fix(sequencer): do not return sentinel values

mtsitrin:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:

### DIFF
--- a/x/sequencer/keeper/grpc_query_sequencers_by_rollapp.go
+++ b/x/sequencer/keeper/grpc_query_sequencers_by_rollapp.go
@@ -49,8 +49,13 @@ func (k Keeper) GetProposerByRollapp(c context.Context, req *types.QueryGetPropo
 	}
 	ctx := sdk.UnwrapSDKContext(c)
 
+	proposer, err := k.RealSequencer(ctx, k.GetProposer(ctx, req.RollappId).Address)
+	if err != nil {
+		return nil, err
+	}
+
 	return &types.QueryGetProposerByRollappResponse{
-		ProposerAddr: k.GetProposer(ctx, req.RollappId).Address,
+		ProposerAddr: proposer.Address,
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


omritoptix (commented):
will this not screw up dymint? 
afair it expects sentinel. 
cc @srene

srene:
> will this not screw up dymint? afair it expects sentinel. cc @srene

i think rotation to sentinel will be broken and it will need to adapt to it, because instead of GetNextProposer() return sentinel proposer, it will return error and rotation will not complete.

omritoptix (dismissed):

x/sequencer/keeper/grpc_query_sequencers_by_rollapp.go
```diff
@@ -60,11 +65,16 @@ func (k Keeper) GetNextProposerByRollapp(c context.Context, req *types.QueryGetN
...
 
+	addr := ""
+	successor := k.GetSuccessor(ctx, req.RollappId)
+	if !successor.Sentinel() {
+		addr = successor.Address
+	}
+
 	return &types.QueryGetNextProposerByRollappResponse{
-		NextProposerAddr:   successor.Address,
+		NextProposerAddr:   addr,
```
omritoptix, line 77: afaiu this only changes from returning `sentinel` to return `` - which is still invalid address. whats the purpose of this?

x/sequencer/keeper/grpc_query_sequencers_by_rollapp.go
```diff
@@ -83,7 +93,8 @@ func (k Keeper) Proposers(c context.Context, req *types.QueryProposersRequest) (
 	pageRes, err := query.Paginate(sequencerStore, req.Pagination, func(key []byte, value []byte) error {
 		proposer, err := k.RealSequencer(ctx, string(value))
 		if err != nil {
-			return err
+			// skip sentinel proposers
+			return nil
```
omritoptix, line 97: I think this will cause an invalid count (i.e +1 for each nil, though there isn't really another one)

omritoptix (approved):

x/sequencer/keeper/grpc_query_sequencers_by_rollapp.go
mtsitrin: seems fine for me. count used to break when limit reached.

PR #1655, opened 2024-12-15
refactor(incentives): skip epoch distribution to perpetual gauges

mtsitrin:
- [ ] custom created gagues are non-perpetual
- [ ] perpetual gauges not iterated on epoch end

### DIFF
--- a/x/incentives/keeper/hooks.go
+++ b/x/incentives/keeper/hooks.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"github.com/dymensionxyz/dymension/v3/x/incentives/types"
 	epochstypes "github.com/osmosis-labs/osmosis/v15/x/epochs/types"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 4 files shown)


x/incentives/keeper/hooks.go
```diff
@@ -25,12 +26,17 @@ func (k Keeper) AfterEpochEnd(ctx sdk.Context, epochIdentifier string, epochNumb
 			}
 		}
 
-		// if len(gauges) > 10 {
-		// 	ctx.EventManager().IncreaseCapacity(2e6)
-		// }
+		// get active, non-perpetual gauges
+		// perpetual gauges are paid directly by the protocol (i.e streamer)
+		newGauges := make([]types.Gauge, 0, len(gauges))
```
keruch, line 31: super nit: i guess most of the gauges will be perpetual, so maybe it's worth doing `make([]types.Gauge, 0)` to avoid runtime overhead. otherwise runtime will try to find a memory slice for `len(gauges)` though in fact we need much less.

keruch (approved):

danwt:
still relevant?

mtsitrin:
> still relevant?

yes, not for 3d

PR #1668, opened 2024-12-18
fix malformed tracking_packet_key by ensuring base64 encoding/decoding during Genesis export/import

### LINKED ISSUE #1576: eibc order tracking_packet_key field malformed upon export
that's because this is bytes stream and we decode it as unicode.
we need to base64 decode and encode it upon genesis export/import.

NagaTulasi:
## Description

----

Closes #1576 

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:

### DIFF
--- a/x/eibc/genesis.go
+++ b/x/eibc/genesis.go
@@ -1,6 +1,9 @@
 package eibc
 
 import (
+	"encoding/base64"
+	"fmt"
+
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dymensionxyz/dymension/v3/x/eibc/keeper"
 	"github.com/dymensionxyz/dymension/v3/x/eibc/types"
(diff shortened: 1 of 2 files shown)


danwt (approved):

mtsitrin (approved):

zale144 (commented):
This is an important change, however I think we might want to address the issue at the source. When setting the tracking packet key field value here https://github.com/dymensionxyz/dymension/blob/afc677f7b9efd4c62facdc9d5042101cda2471d4/x/eibc/types/demand_order.go#L23, we already might have data loss due to conversion of certain bytes, in particular whenever the order object is marshalled and stored. So, we could consider encoding it right there to preserve the bytes when saving to store, and perhaps renaming the field to `TrackingPacketKeyEncoded` to avoid confusion. However, changing the field might be a more sensitive change, we would need to make sure it is encoded/decoded in all necessary places.

zale144:
> This is an important change, however I think we might want to address the issue at the source. When setting the tracking packet key field value here
> 
> https://github.com/dymensionxyz/dymension/blob/afc677f7b9efd4c62facdc9d5042101cda2471d4/x/eibc/types/demand_order.go#L23
> 
> , we already might have data loss due to conversion of certain bytes, in particular whenever the order object is marshalled and stored. So, we could consider encoding it right there to preserve the bytes when saving to store, and perhaps renaming the field to `TrackingPacketKeyEncoded` to avoid confusion. However, changing the field might be a more sensitive change, we would need to make sure it is encoded/decoded in all necessary places.

Ignore this comment, I realize bytes would still be preserved, I did observe loss when creating events though.

PR #1749, opened 2025-02-23
chore(dpes):  updated math package

mtsitrin:
As preparation for v50 upgrade,
we bump all the `sdk.Int / sdk.Dec` usage, to use the dedicated cosmos math package


danwt:
bravo

PR #1750, opened 2025-02-23
feat(IRO): setting equilibrium point where the raised dym goes to the liquidity pool

mtsitrin:
This PR changes the way IRO allocated tokens are handled.
In case of IRO allocation `T`,
instead of all tokens available for sell,

we find an `equilibrium` point `X`, where:
* The raised DYM for X and the unsold tokens (T-X) will have the same price as the bonding curve

It means that IRO plan will have a `max_amount_to_sell` parameter

### DIFF
--- a/x/iro/types/bonding_curve.go
+++ b/x/iro/types/bonding_curve.go
@@ -79,6 +79,11 @@ func (lbc BondingCurve) ValidateBasic() error {
 		return errorsmod.Wrapf(ErrInvalidBondingCurve, "c: %s", lbc.C.String())
 	}
 
+	// positive C is supported only for fixed price for now (due to equilibrium calculation)
+	if !lbc.C.IsZero() && !lbc.M.IsZero() {
+		return errorsmod.Wrapf(ErrInvalidBondingCurve, "m: %s, c: %s", lbc.M.String(), lbc.C.String())
+	}
+
 	if !checkPrecision(lbc.N) {
 		return errorsmod.Wrapf(ErrInvalidBondingCurve, "N must have at most %d decimal places", MaxNPrecision)
 	}
--- a/x/iro/types/liquidity.go
+++ b/x/iro/types/liquidity.go
@@ -1,0 +1,52 @@
+package types
+
+import "cosmossdk.io/math"
+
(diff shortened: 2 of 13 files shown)


x/iro/types/liquidity.go
```diff
@@ -0,0 +1,54 @@
+package types
+
+import "cosmossdk.io/math"
+
+// CalcLiquidityPoolTokens determines the tokens and DYM to be used for bootstrapping the liquidity pool.
+//
+// This function calculates the required DYM based on the settled token price and compares it with the raised DYM.
+// It returns the amount of RA tokens and DYM to be used for bootstrapping the liquidity pool so it fulfills the last price.
+// We expect all the raised dym to be used for liquidity pool, so incentives will consist of the remaining tokens.s
```
omritoptix, line 9:
```suggestion
// We expect all the raised dym to be used for liquidity pool, so incentives will consist of the remaining tokens.
```

x/iro/types/bonding_curve.go
```diff
@@ -237,31 +242,34 @@ func (lbc BondingCurve) integral(x math.LegacyDec) math.LegacyDec {
...
 
-	// Calculate C * T
-	cTimesT := cBig.Mul(tBig)
+	// log(nominator) = log(val) + log(N + 1) + (N+1)log(N + 2)
+	lognum := valBig.LogBase2().Add(nPlusOne.LogBase2()).Add(nPlusOne.Mul(nPlusTwo.LogBase2()))
 
-	// Calculate VAL - C * Z
-	numerator := valBig.Sub(cTimesT)
+	// log(denominator) = (N+1)*log(T*(N+1)) + log(2)
+	logdenom := (nPlusOne.Mul((tBig.Mul(nPlusOne)).LogBase2())).Add(osmomath.OneDec())
```
omritoptix, line 265: it says `.Add(osmomath.OneDec)` but the comment says `+ log(2)`. shouldn't it be `.Add(osmomath.TwoDec().LogBase2())` ?

mtsitrin: it's the same actually

omritoptix: right 🫠

omritoptix (approved):

PR #1753, opened 2025-02-25
feat(iro): founder can trade before start time

mtsitrin:
rollapp founder can trade IRO even before the start time

This reverts commit 99d858091e0e2fa95d06e1b2da2d1be48eecd0a9.


omritoptix (dismissed):
Approved but lets plz add a unit test.

PR #1769, opened 2025-03-04
feat(iro): ability to launch iro on demand

### LINKED ISSUE #1763: Ability to launch IRO on-demand
currently a founder can launch IRO either right away or in a pre-determined time.

this removes the element of surprise.

the goal is to have the IRO on the portal, which is not open yet, but it's opening time is up to the founder, which he may or may not disclose.

mtsitrin:
closes #1763


omritoptix (changes requested):
I think there is a potential problem here where the sequencer can register before the owner enabled trading?

omritoptix:
as discussed we want the ability to have launch duration start only after we enable.

omritoptix (approved):

PR #1793, opened 2025-03-18
feat(iro):  iro upgrade handler

mtsitrin:
* IRO store migration by module migration flow
* Upgrade handler:
  * Lockup params
  * IRO params
  * GAMM params
  * x/gov params


x/iro/keeper/create_plan.go
```diff
@@ -231,3 +231,13 @@ func (k Keeper) MintAllocation(ctx sdk.Context, allocatedAmount math.Int, rollap
 	}
 	return minted, nil
 }
+
+// Function to check if a slice contains a string
+func contains(slice []string, str string) bool {
```
omritoptix, line 236: by mistake?

x/iro/keeper/migrations.go
```diff
@@ -0,0 +1,60 @@
...
+		// migrate bonding curve
+		plan.BondingCurve.LiquidityDenomDecimals = 18
+		plan.BondingCurve.RollappDenomDecimals = 18
+
+		// liquidity part is 1.0 for old plans
+		plan.LiquidityPart = math.LegacyOneDec()
+
+		// max amount to sell is calculated from bonding curve
+		eq := types.FindEquilibrium(plan.BondingCurve, plan.TotalAllocation.Amount, plan.LiquidityPart)
+		plan.MaxAmountToSell = math.MaxInt(plan.SoldAmt, eq)
```
omritoptix, line 41: how will this effect on going IROs if already passed their MaxAmountToSell?

should we not just set the MaxAmountToSell for existing ones to be the original?

mtsitrin: u mean if it passed the eq point?
1. when bootstraping the pool, we might have dym as incentive as in old behavior
2. `plan.MaxAmountToSell ` will be `plan.SoldAmt` and no more tokens available to sell.

PR #1814, opened 2025-04-16
feat(genesis-bridge): pessimistic validation for rollapp genesis bridge opening

mtsitrin:
Both the Canonical Client, and the Genesis Bridge, now requires that the state update will be available for all required heights
(
     canonical client - all IBC headers must be committed,
    genesis bridge - packet proof height must be committed
)

tested with https://github.com/dymensionxyz/go-relayer/pull/73


x/lightclient/keeper/canonical_client.go
```diff
@@ -103,18 +103,33 @@ func (k Keeper) expectedClient() ibctm.ClientState {
...
 	expClient := k.expectedClient()
 	if err := types.IsCanonicalClientParamsValid(cs, &expClient); err != nil {
 		return errors.Join(err, ErrParamsMismatch)
 	}
 
+	// Check if the clientID has any connections
+	_, found := k.ibcConnectionK.GetClientConnectionPaths(ctx, clientID)
+	if found {
+		return gerrc.ErrFailedPrecondition.Wrap("client already has connections")
+	}
```
danwt, line 118: duplicated, should remove the same thing in the calling func

danwt, line 118: duplicated, should remove the same thing in the calling func
also, don't need the comment

x/lightclient/keeper/canonical_client.go
```diff
@@ -103,18 +103,33 @@ func (k Keeper) expectedClient() ibctm.ClientState {
...
+
+	// validate lastest height is already committed
+	latestCommittedHeight, ok := k.rollappKeeper.GetLatestHeight(ctx, rollappId)
+	if !ok {
+		return gerrc.ErrNotFound.Wrap("latest height")
+	}
+	if latestCommittedHeight < cs.GetLatestHeight().GetRevisionHeight() {
+		return gerrc.ErrInvalidArgument.Wrap("client's latest height not committed")
```
danwt, line 126:
```suggestion
		return gerrc.ErrInvalidArgument.Wrap("clients latest height not committed")
```

danwt (changes requested):

x/lightclient/keeper/canonical_client.go
mtsitrin: done

danwt (approved):

PR #1845, opened 2025-05-12
feat(incentives):  dont distribute endorsement rewards to non active rollapps

### LINKED ISSUE #1841: don't distribute endorsement rewards to non-active rollapps
currently we check wether the rollapp is **launched** before paying endorsement rewards to rollapp gauge.

we should also check this rollapp is actually **active** (e.g by checking wether the proposer is kickable)

mtsitrin:
closes #1841


danwt (dismissed):

danwt (approved):

danwt:
I guess tbh most sequencers are kickable, even for 'active' rollapps
In retrospect IsActive should be a rollapp module method which a more careful definition

But after migration this kickable property will become more accurate due to params changes

@mtsitrin fyi

mtsitrin:
> But after migration this kickable property will become more accurate due to params changes

yes, we took it as an assumption

> In retrospect IsActive should be a rollapp module method which a more careful definition

I actually think this can and should be achieved by the dishonor mechanism

danwt:
x/rollapp should determine if rollapp is 'active', it just makes sense
if penalty points are used in such a function, sure

PR #1864, opened 2025-05-20
fix(sponsorship): added epoch identifier

keruch:
## Description

Adds an epoch identifier param to protect against abuse when claiming rewards.

Closes https://github.com/dymensionxyz/audits/issues/31


mtsitrin:
I would suggest to use the epoch identifier of `x/incentives`, instead of defining a new param
@keruch wdyt

PR #1865, opened 2025-05-20
fix(sponsorship): missing total power in the endorsement migration

keruch:
## Description

We need to carefully fill the `TotalShare` setting of endorsements for existing RAs. To do so, we iterate through the distribution gauges and check theirs total voting power. It might be zero if no one endorsed the RA. `TotalShare` equals to the total voting power cast to the RA.

Closes https://github.com/dymensionxyz/audits/issues/32

### DIFF
--- a/app/upgrades/v5/upgrade.go
+++ b/app/upgrades/v5/upgrade.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"cosmossdk.io/math"
 	upgradetypes "cosmossdk.io/x/upgrade/types"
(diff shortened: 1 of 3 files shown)


mtsitrin (approved):

danwt:
is this safe with this zero div? https://github.com/dymensionxyz/dymension/blob/3e0e9effdd628dfd8c3c21cae98ff3a14162791f/x/sponsorship/keeper/endorsements.go#L108

mtsitrin:
> is this safe with this zero div? @keruch
> 

it checks power > 0 before for this gauge, which applies that endorsment.Shares > 0

PR #1869, opened 2025-05-22
test(forward): more coverage for extended ibc

danwt:
This is supposed to hit some missing cases from https://docs.google.com/spreadsheets/d/1jaPAQcEs6FUWTulAvJ177oiNer92CZ573NmTHs5QnLI/edit?gid=1770583565#gid=1770583565
It was deprioritised

Also probably need to make sure that spreadsheet is updated (see 'PR IN DEVELOPMENT' boxes)


danwt:
abandonware

Issue #1856, opened 2025-05-19
Missing Non-Negative for Coin Type in MsgCreateGauge Msg


Hellobloc:
the MsgCreateGauge Msg don't validate non-negative properties which could lead a negative Coins.Add() process in fee charge process, this would been a risky problem which may lead logic problem.
https://github.com/dymensionxyz/dymension/blob/90824cb9a1b1066f31f10fe389c296ab08f2578b/x/incentives/types/msgs.go#L33-L62

mtsitrin:
thx for the report
handled over https://github.com/dymensionxyz/dymension/pull/1871

PR #1871, opened 2025-05-23
feat(incentives): support legacy EIP712 for `CreateGauge`

mtsitrin:
* removes the `oneof` from the proto, and use enum instead
* update the EIP712 test to produce non legacy output as well

**also fixes validation issue as reported on #1856**


artemijspavlovs (dismissed):

proto/dymensionxyz/dymension/incentives/tx.proto
```diff
@@ -45,25 +52,26 @@ message MsgCreateGauge {
...
-  google.protobuf.Timestamp start_time = 5 [
+  google.protobuf.Timestamp start_time = 7 [
     (gogoproto.stdtime) = true,
     (gogoproto.nullable) = false,
     (gogoproto.moretags) = "yaml:\"timestamp\""
   ];
   // num_epochs_paid_over is the number of epochs distribution will be completed
   // over
-  uint64 num_epochs_paid_over = 6;
+  uint64 num_epochs_paid_over = 8;
```
danwt, line 74: Please can you not change the numbers and instead add new ones and reserve old ones?

x/incentives/keeper/msg_server.go
```diff
@@ -66,17 +72,25 @@ func (server msgServer) CreateGauge(goCtx context.Context, msg *types.MsgCreateG
...
 		}
-	case *types.MsgCreateGauge_Endorsement:
-		gaugeID, err = server.keeper.CreateEndorsementGauge(ctx, msg.IsPerpetual, owner, msg.Coins, *distr.Endorsement, msg.StartTime, msg.NumEpochsPaidOver)
+		gaugeID, err = server.keeper.CreateEndorsementGauge(ctx, msg.IsPerpetual, owner, msg.Coins, *msg.Endorsement, msg.StartTime, msg.NumEpochsPaidOver)
 		if err != nil {
-			return nil, fmt.Errorf("create gauge: %w", err)
+			return nil, fmt.Errorf("create endorsement gauge: %w", err)
 		}
+	default:
+		return nil, fmt.Errorf("unsupported gauge type: %v", msg.GaugeType)
```
danwt, line 93: these errors are not grpc registered (nit)

x/incentives/types/msgs.go
```diff
@@ -44,18 +32,34 @@ func (m MsgCreateGauge) ValidateBasic() error {
...
 		}
-	case *MsgCreateGauge_Endorsement:
-		if distr.Endorsement.RollappId == "" {
+	case GaugeType_GAUGE_TYPE_ENDORSEMENT:
+		if m.Endorsement == nil {
+			return errors.New("endorsement must be set for endorsement gauge type")
+		}
+		if m.Endorsement.RollappId == "" {
 			return errors.New("rollapp id should be set")
 		}
```
danwt, line 60: this should be in a method on the type

danwt (changes requested):

proto/dymensionxyz/dymension/incentives/tx.proto
chich12: it doesn't matter, tx protos needs to be regenerated anyway
it's not store migration

mtsitrin: it doesn't matter, tx protos needs to be regenerated anyway
it's not store migration

danwt (dismissed):

PR #1872, opened 2025-05-23
feat(sponsorship): lazy accumulator for endorsement yielding

keruch:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:


x/sponsorship/keeper/endorsements.go
```diff
@@ -41,36 +83,53 @@ func (k Keeper) UpdateTotalSharesWithDistribution(ctx sdk.Context, update types.
 // 7. Update the endorsement epoch shares
 // 8. Blacklist the user from claiming rewards in this epoch
```
omritoptix: seems outdated

x/sponsorship/keeper/endorsements.go
```diff
@@ -81,37 +140,66 @@ func (k Keeper) EstimateClaim(ctx sdk.Context, claimer sdk.AccAddress, gaugeId u
...
-		return EstimateClaimResult{}, fmt.Errorf("user does not endorse respective RA gauge: %d", gaugeId)
-	}
+	// TODO: there is a problem with precision in big decimals. Imagine
+	// Accumulator = 7.(6)
+	// LastSeenAccumulator = 6
+	// Shares = 60
+	// Thus, Rewards = (7.(6) - 6) * 60 = 99,(9) is approx. 100
+	// However, 7.(6) - 6 is calculated not as 1.(6), but 1.666666667, so
+	// 1.66666666667 * 60 > 100, so the user will claim more than available.
+	// This is not probable, but still the case.
```
omritoptix: as discussed - ideally we can find some alternative solution (possibly rounding down) as this relies on incentives module to have "change" (which may be true for DYM) but is not necessarily the case with alternative tokens being used for rewards.

omritoptix (changes requested):

x/sponsorship/keeper/endorsements.go
keruch: fixed! I managed to round it down in calculations

x/incentives/keeper/gauge_endorsement.go
```diff
@@ -49,12 +51,22 @@ func (k Keeper) updateEndorsementGaugeOnEpochEnd(ctx sdk.Context, gauge types.Ga
...
-	endorsement.EpochRewards = epochRewards // we operate a pointer
-	gauge.FilledEpochs += 1
+	// Update endorsement total coins with the epoch rewards
+	endorsementGauge := gauge.DistributeTo.(*types.Gauge_Endorsement)
 
-	if err := k.setGauge(ctx, &gauge); err != nil {
-		return err
+	err := k.spk.UpdateEndorsementTotalCoins(ctx, endorsementGauge.Endorsement.RollappId, epochRewards)
+	switch {
+	case err == nil:
```
mtsitrin: err == nil can be out of the switch case

x/sponsorship/types/keys.go
```diff
@@ -21,7 +21,8 @@ const (
 	DelegatorValidatorPowerByte        // Delegator voting power by the validator: math.Int
 	VoteByte                           // User's vote: Vote
 	RAEndorsementsByte                 // RA endorsement: Endorsement
-	ClaimBlacklistIndexByte            // Claim blacklist: sdk.AccAddress
+	ClaimBlacklistIndexByte            // Deprecated. Claim blacklist: sdk.AccAddress
```
mtsitrin:
```suggestion
    _ // Deprecated. Claim blacklist: sdk.AccAddress
```

x/sponsorship/keeper/helpers.go
```diff
@@ -42,6 +42,9 @@ func (k Keeper) UpdateDistribution(ctx sdk.Context, fn func(types.Distribution)
 	// Apply the update
 	result := fn(current)
 
+	// Don't store gauges with <= 0 power
```
mtsitrin: don't u already have it in the merge?
```
    // Don't store gauges with 0 power
    if !gauge.Power.IsZero() {
        gauges = append(gauges, gauge)
    }
```

x/incentives/keeper/gauge_endorsement.go
```diff
@@ -49,12 +51,22 @@ func (k Keeper) updateEndorsementGaugeOnEpochEnd(ctx sdk.Context, gauge types.Ga
...
+		if err := k.setGauge(ctx, &gauge); err != nil {
+			return err
+		}
+	case errors.Is(err, sponsorshiptypes.ErrNoEndorsers):
+		// Don't fill this epoch, save rewards for the future
+	default:
+		return fmt.Errorf("update endorsement total coins: %w", err)
 	}
 
 	return nil
```
mtsitrin: `DistributeEndorsementRewards` not used

mtsitrin (approved):
GJ!

x/incentives/keeper/gauge_endorsement.go
keruch: removed

x/sponsorship/keeper/helpers.go
keruch: in `merge`, I don't store gauges with `power == 0`, so after the merge operation there might be gauges with `power < 0`. in `UpdateDistribution` we don't store gauges with `power <= 0`. the difference is that we don't want to store negative gauges in the `distribution` object, but we do want account for negative weights when applying updates using the `merge` function.

omritoptix (dismissed):

x/sponsorship/keeper/helpers.go
mtsitrin: I see
how it's even possible to have negative gauge after a merge?

mtsitrin (dismissed):

chich12:
@keruch build fails

mtsitrin:
@keruch build fails

x/sponsorship/keeper/helpers.go
keruch: Let's say we have a global distribution like that:

[1, 1000] [2, 2000] power 3000

Imagine you have a user that has 400 DYM staked and he votes 25% on Gauge1 and 75% on Gauge2. When we apply his weights, we convert his vote to a distribution update, which will look like the following:

[1, 100] [2, 300] power 400

So the distribution after update looks like:

[1, 1000] [2, 2000] power 3000 +
[1, 100] [2, 300] power 400 =
[1, 1100] [2, 2300] power 3400

Now imagine that this user wants to _update_ his vote (i.e. place a new one while he has the existing). He votes 25% on Gauge1 and 25% on Gauge2, so the distribution update should look like the following: 

[1, 100] [2, 100] power 400

However, since the user has the previous vote, we need to account for it as well. For that, we subtract the previous vote from the new one before applying the update (using the merge operation):

[1, 100] [2, 100] power 400 -
[1, 100] [2, 300] power 400 =
[2, -200] power 0 **<— negative power**

Then, apply this update to the global distribution:

[1, 1100] [2, 2300] power 3400 +
[2, -200] power 0 =
**[1, 1100] [2, 2100] power 3400** 

Note that the global distribution now is as if the user applied his vote for the initial global distribution:

[1, 1000] [2, 2000] power 3000 + 
[1, 100] [2, 100] power 400 = 
**[1, 1100] [2, 2100] power 3400**

omritoptix (approved):

PR #1875, opened 2025-05-27
feat(incentives): incentives distribution based on lock age

### LINKED ISSUE #1874: change incentives distirubtion to be based on lock age instead of lock duration
according to the ADR 
https://www.notion.so/dymension/ADR-Incentives-for-locked-tokens-1ffa4a51f86a8091a279dc8e136deb03?pvs=4

mtsitrin:
ADR: https://www.notion.so/dymension/ADR-Incentives-for-locked-tokens-1ffa4a51f86a8091a279dc8e136deb03

closes #1874 

`x/lockup`:
* add updated_at which updated on creation and on adding tokens to a lock

`x/incentives`:
* new params for default protocol incentives
* locks eligablitiy by lock age

TODOs:
* UT

### DIFF
--- a/x/incentives/keeper/gauge_asset_test.go
+++ b/x/incentives/keeper/gauge_asset_test.go
@@ -1,0 +1,102 @@
+package keeper_test
+
+import (
+	"time"
+
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/dymensionxyz/dymension/v3/x/incentives/types"
+	lockuptypes "github.com/dymensionxyz/dymension/v3/x/lockup/types"
+)
+
+// TestGetDistributeToBaseLocks tests Keeper.GetDistributeToBaseLocks for gauges with only duration, only lock age, and both.
+func (suite *KeeperTestSuite) TestGetDistributeToBaseLocks() {
+	suite.SetupTest()
+
+	baseDenom := defaultLPDenom
+	coins := sdk.NewCoins(sdk.NewInt64Coin(baseDenom, 10))
+	owner := suite.setupAddr(0, "", coins.MulInt(math.NewInt(3)))
+
+	// Create three locks:
+	// 1. duration = 10s, created 20s ago
+	// 2. duration = 20s, created 10s ago
+	// 3. duration = 30s, created 40s ago
+	blockTime := time.Now()
+	suite.Ctx = suite.Ctx.WithBlockTime(blockTime)
+
+	locks := []struct {
+		lockDuration     time.Duration
+		lockAmount       sdk.Coins
+		lockCreationTime time.Time
+	}{
+		{
+			lockDuration:     10 * time.Second,
+			lockAmount:       coins,
+			lockCreationTime: blockTime.Add(-20 * time.Second),
+		},
+		{
+			lockDuration:     20 * time.Second,
+			lockAmount:       coins,
+			lockCreationTime: blockTime.Add(-10 * time.Second),
+		},
+		{
+			lockDuration:     30 * time.Second,
+			lockAmount:       coins,
+			lockCreationTime: blockTime.Add(-40 * time.Second),
+		},
+	}
+
+	for _, lock := range locks {
+		suite.Ctx = suite.Ctx.WithBlockTime(lock.lockCreationTime)
+		suite.LockTokens(owner, lock.lockAmount, lock.lockDuration)
+	}
+
+	suite.Ctx = suite.Ctx.WithBlockTime(blockTime)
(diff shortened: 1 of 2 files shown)


app/upgrades/v5/upgrade.go
```diff
@@ -354,3 +373,38 @@ func migrateSequencers(ctx sdk.Context, k *sequencerkeeper.Keeper) {
...
+
+// migrateGaugeLockAges sets the min lock age for all perpetual asset gauges
+func migrateGaugeLockAges(ctx sdk.Context, incentivesKeeper *incentiveskeeper.Keeper) error {
+	params := incentivesKeeper.GetParams(ctx)
+	minLockAge := params.MinLockAge
+	gauges := incentivesKeeper.GetGauges(ctx)
+	for _, gauge := range gauges {
+		if gauge.IsPerpetual && gauge.GetAsset() != nil {
+			asset := gauge.GetAsset()
+			asset.LockAge = minLockAge
```
omritoptix, line 403: isn't in min lock age for lock to be eligable? why are we setting this on the asset gauge?

I thought asset gauge query condition would be by default 24 hours.

does the min lock age also would be 24 hours in that case (as the upper migration uses the same param to set min lock age on the locks)?

if so why do we need them both? or should this param be just for the locks and here we should use hardcoding?

omritoptix (changes requested):
nice!

keruch (commented):
left some comments in the ADR

x/incentives/keeper/gauge_asset.go
```diff
@@ -16,7 +16,7 @@ import (
 func (k Keeper) CreateAssetGauge(ctx sdk.Context, isPerpetual bool, owner sdk.AccAddress, coins sdk.Coins, distrTo lockuptypes.QueryCondition, startTime time.Time, numEpochsPaidOver uint64) (uint64, error) {
 	// Ensure that this gauge's duration is one of the allowed durations on chain
 	durations := k.GetLockableDurations(ctx)
-	if distrTo.LockQueryType == lockuptypes.ByDuration {
+	if distrTo.Duration > 0 {
```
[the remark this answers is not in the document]
mtsitrin: the opposite.
all locks will be eligable (every lock with duration > 0 and age > 0)

app/upgrades/v5/upgrade.go
mtsitrin: > I thought asset gauge query condition would be by default 24 hours.

currently asset gauges query condition is `lock duration > 60s`
i'm adding the `lock age > minLockAge` condition to **existing** gauges. new gauges will be already created like this.

the other migration sets the timestamp on the locks

app/upgrades/v5/upgrade.go
```diff
@@ -64,17 +65,27 @@ func CreateUpgradeHandler(
 			return nil, fmt.Errorf("migrate endorsements: %w", err)
 		}
 
+		// Migrate locks: set creation_timestamp (UpdatedAt) if not set
+		if err := migrateLockTimestamps(ctx, keepers.LockupKeeper, keepers.IncentivesKeeper); err != nil {
```
mtsitrin, line 69: @omritoptix this one actually not required.
We can leave existing locks with empty timestamp

mtsitrin: i'm leaving the migration in place for completeness

x/incentives/keeper/gauge_asset.go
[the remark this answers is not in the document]
omritoptix: yea. sorry. you're right. meant all locks. wondering if it makes sense or should we set a validation to protect gauge creator to specificy at least one query condition.

app/upgrades/v5/upgrade.go
omritoptix: > We can leave existing locks with empty timestamp

is this because on the code we assume they are old hence are eligable?

mtsitrin: yes. it will have empty timestamp, which is old

x/incentives/keeper/gauge_asset.go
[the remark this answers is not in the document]
mtsitrin: we can enforce that custom created gauges will have the minimal params limitation as well
(min_lock_age and min_duration)
wdyt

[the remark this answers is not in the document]
omritoptix: rethinking about it I think it's fine as is. the idea was to protect creator against mistakes but we can do it on FE validation. maybe there is a scenario the creator just wants to reward all token holders without any condition.

PR #1877, opened 2025-05-27
feat(incentives): prevent multiple partial unlocks for a single lock

mtsitrin:
closes #18


keruch (commented):
Good! Agree with Omri about tests

omritoptix (approved):

PR #1894, opened 2025-06-05
feat(lightclient): using staking unbond time as the expected canonical trust period

mtsitrin:
## Description

----

Closes #XXX

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:

### DIFF
--- a/app/keepers.go
+++ b/app/keepers.go
@@ -377,6 +377,7 @@ func (a *AppKeepers) InitKeepers(
 		a.IBCKeeper.ChannelKeeper,
 		a.SequencerKeeper,
 		a.RollappKeeper,
+		a.StakingKeeper,
 	)
 
 	a.SequencerKeeper.SetUnbondBlockers(a.RollappKeeper, a.LightClientKeeper)
--- a/testutil/keeper/lightclient.go
+++ b/testutil/keeper/lightclient.go
@@ -82,6 +82,7 @@ func LightClientKeeper(t testing.TB) (*keeper.Keeper, sdk.Context) {
 	mockIBCKeeper := NewMockIBCClientKeeper(consStates, genesisClients)
 	mockSequencerKeeper := NewMockSequencerKeeper(seqs)
 	mockRollappKeeper := NewMockRollappKeeper()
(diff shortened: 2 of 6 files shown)


omritoptix (commented):
as I see it the problem with it being a param is that it needs to be correlated with the rollapp(i.e sequencer) unbonding period (otherwise it won't pass validation). and here you are essentially forcing the rollapp unbonding period to match the validator unbonding period which I'm not sure it's an enforcement we want to have.

mtsitrin:
> as I see it the problem with it being a param is that it needs to be correlated with the rollapp(i.e sequencer) unbonding period (otherwise it won't pass validation). and here you are essentially forcing the rollapp unbonding period to match the validator unbonding period which I'm not sure it's an enforcement we want to have.

1. it's relayer's responsibility to put the expected value for a canonical client
2. the `x/sequencer` unbonding time in the RDK is useless. it's queried currently by the relayer, but we can have simpler logic for it
3. as we anyway enforce our logic through the `x/lightclient` we might even relax this constraint and have "arbitrary long period" (as we have 3 weeks now, which is not correlated with actual unbonding)
4. using `x/staking` gives us this long duration, while still changeable in tests

omritoptix (commented):
I think regardless of how it’s being set, the question is - should we tie it to the hub unbonding time .Currently the unbonding time of the rollapp LC on the hub is mostly used as an upper limit for trusting period. Trusting period main role is determine client expiration and block further updates if passed - which will now be limited by the hub staking param.

Also testing various rollup unbonding scenarios will now require changing the hub unbonding param (which has really nothing to do with the rollup unbonding param). Not sure if there are any edge cases there which could cause false positive in tests but imo it smells and hard to reason about. 

Either way since it’s been hardcoded thus far it doesn’t take any capability we had, I guess we can leave it for now if it helps with specific e2e test.

PR #1901, opened 2025-06-09
Fix incorrect link in CHANGELOG entry for rollapp issue

MamunC0der:
Corrected invalid URL in changelog:
Replaced research/issues/317 with the correct link dymension/issues/317.

The previous link pointed to a non-existent repository


mtsitrin:
it's internal repo

PR #1945, opened 2025-07-31
Fix: use fixed hyperlane cosmos dependency which properly allows hook setting

danwt:
uses https://github.com/dymensionxyz/hyperlane-cosmos/pull/13

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -248,7 +248,7 @@ require (
 )
 
 replace (
-	github.com/bcp-innovations/hyperlane-cosmos => github.com/dymensionxyz/hyperlane-cosmos v0.0.0-20250708152202-1417b7791dbd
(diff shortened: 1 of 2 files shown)


srene (commented):
Not sure this will work. I think the problem is not the hook itself, but the NewKeeper() that should return a pointer https://github.com/dymensionxyz/hyperlane-cosmos/blob/1417b7791dbdac06752a642606bd581f99856b84/x/warp/keeper/keeper.go#L72. NewKeeper() registers a pointer of the keeper in the approuter for the hl messages received, and the same pointer should be returned to be used later to set the hook, otherwise it ends up setting the hook in a different instance of the warp keeper.

danwt:
But the SetHook method now changes the memory across all instances

PR #1948, opened 2025-08-04
feat: print HL message field by field in utils (CLIENT ONLY)

danwt:


srene (approved):
shouldnt the new function be also called in the other places where the HL message is also printed?

PR #1955, opened 2025-08-10
fix(IRO): ensure approximation converges only via under-approximation

mtsitrin:


x/iro/types/bonding_curve.go
```diff
@@ -171,15 +171,14 @@ func (lbc BondingCurve) TokensApproximation(startingX, spendTokens math.LegacyDe
...
-	x := spendTokens
+	x := spendTokens.Quo(lbc.spotPriceInternal(startingX))
 
 	// Newton-Raphson iteration
 	epsilonDec := math.LegacyNewDecWithPrec(1, epsilonPrecision)
 	for i := 0; i < maxIterations; i++ {
 		fx := f(x) // diff between spendTokens and the actual cost to get to x
 		// If the function converges, return the result
-		if fx.Abs().LT(epsilonDec) {
+		if !fx.IsPositive() && fx.Abs().LT(epsilonDec) {
```
mtsitrin, line 181: This is the actual fix

danwt: still it always converges?

mtsitrin: yes. it might take one more iteration, but it converges.

PR #1973, opened 2025-08-18
docs(ai): enhance claude.md with CLI info

danwt:


CLAUDE.md
```diff
@@ -222,4 +222,718 @@ When verifying code locally, run tools **only as needed** to avoid unnecessary w
...
+```bash
+# Initialize node with moniker
+dymd init my-node --chain-id dymension_1100-1
+
+# Initialize with specific denom
+dymd init my-node --chain-id dymension_1100-1 --default-denom adym
+
+# Recover from existing mnemonic
+dymd init my-node --recover
+```
```
keruch, line 271: maybe it's better to use `bash scripts/setup_local.sh` instead? from my experience, it does a good job configuring the local node

danwt: it should be lol

mtsitrin (approved):

Issue #1979, opened 2025-08-26
Implement Pump stream


omritoptix:
Research: https://www.notion.so/dymension/Pump-Stream-Transforming-Sponsorship-from-Rewards-to-Token-Buying-253a4a51f86a8086b3f4f70bbde6ea8b

ADR: https://www.notion.so/dymension/ADR-X-Pump-Stream-257a4a51f86a80e09b2addedd16aa157

PR #1981, opened 2025-08-26
feat(streamer): pump stream

keruch:
## Description

----

Closes https://github.com/dymensionxyz/dymension/issues/1979

PR review checkboxes:

I have...

SDK Checklist
- [ ] Import/Export Genesis
- [ ] Registered Invariants
- [ ] Registered Events
- [ ] Updated openapi.yaml
- [ ] No usage of go `map`

----

For Reviewer:

- [ ]  Reviewers assigned

----

After reviewer approval:

### DIFF
--- a/utils/rand/random.go
+++ b/utils/rand/random.go
@@ -18,17 +18,17 @@ func GenerateUniformRandom(ctx sdk.Context) *big.Int {
 	return new(big.Int).SetUint64(generator.Uint64())
 }
 
+var BigMaxUint64 = osmomath.NewIntFromUint64(math.MaxUint64)
+
 // GenerateUniformRandomMod draws a uniform random variable in [0; modulo)
 func GenerateUniformRandomMod(ctx sdk.Context, modulo *big.Int) *big.Int {
-	// Generate a uniform random variable in [0; MaxUint64)
+	// Generate a uniform random variable in [0; MaxUint64]
 	u := GenerateUniformRandom(ctx)
 	// Normalize it to [0; 1) = uniform / (MAX_UINT64 + 1)
-	un := osmomath.NewDecFromBigInt(u).QuoInt(osmomath.NewIntFromUint64(math.MaxUint64).AddRaw(1))
+	un := osmomath.NewDecFromBigInt(u).QuoInt(BigMaxUint64.AddRaw(1))
 	// Scale it up to the modulo
 	return un.MulInt(osmomath.NewIntFromBigInt(modulo)).TruncateInt().BigInt()
 }
-
-var BigMaxUint64 = osmomath.NewIntFromUint64(math.MaxUint64)
 
 // GenerateExpRandomLambda draws an exp random variable in [0, +inf)
 // with lambda = numerator / denominator and mean = denominator / numerator.


proto/dymensionxyz/dymension/streamer/query.proto
```diff
@@ -45,6 +46,16 @@ service Query {
     option (google.api.http).get =
         "/dymensionxyz/dymension/streamer/upcoming_streams";
   }
+
+  rpc PumpPressure(PumpPressureRequest) returns (PumpPressureResponse) {
```
keruch, line 51: query server is not implemented so far!

simulation/simulation_test.go
```diff
@@ -75,6 +75,13 @@ func TestFullAppSimulation(t *testing.T) {
 	config := simcli.NewConfigFromFlags()
 	config.ChainID = SimulationAppChainID
 
+	// Configure SDK address prefixes before creating app
```
keruch, line 78: misc stuff. fixes simulation tests after v50.

utils/uinv/funcs.go
```diff
@@ -71,3 +71,16 @@ func (l NamedFuncsList[K]) All(module string, keeper K) sdk.Invariant {
 		return "", false
 	}
 }
+
+// Combines all invariants into one function returning error
+func (l NamedFuncsList[K]) AllFn(keeper K) Func {
```
keruch, line 76: helper to run invariants in tests

x/iro/keeper/trade.go
```diff
@@ -115,53 +115,58 @@ func (k Keeper) Buy(ctx sdk.Context, planId string, buyer sdk.AccAddress, amount
 }
 
 // BuyExactSpend uses exact amount of liquidity to buy tokens on the curve
-func (k Keeper) BuyExactSpend(ctx sdk.Context, planId string, buyer sdk.AccAddress, amountToSpend, minTokensAmt math.Int) error {
+func (k Keeper) BuyExactSpend(
+	ctx sdk.Context,
+	planId string,
+	buyer sdk.AccAddress,
+	amountToSpend, minTokensAmt math.Int,
+) (tokensOutAmt math.Int, err error) {
```
keruch, line 123: we need to know exact tokenOut to burn it further

x/sponsorship/keeper/endorsements_test.go
keruch, line 1: misc stuff. adds invariant checks to x/sponsorship test.

x/streamer/keeper/grpc_query.go
```diff
@@ -101,6 +101,36 @@ func (q Querier) UpcomingStreams(goCtx context.Context, req *types.UpcomingStrea
 	return &types.UpcomingStreamsResponse{Data: streams, Pagination: pageRes}, nil
 }
 
+func (q Querier) PumpPressure(goCtx context.Context, req *types.PumpPressureRequest) (*types.PumpPressureResponse, error) {
```
keruch, line 107: PumpPressure and PumpPressureByRollapp are not implemented so far!

x/streamer/keeper/stream_iterator.go
```diff
@@ -103,20 +103,24 @@ func (i *StreamIterator) findNextStream() {
...
+// 3. Stream is non-empty: there are some gauges assigned to this stream
+// 4. Stream epoch identifier matches the provided
+// 5. gaugeIdx is less than the number of gauges: the iterator points to the existing gauge
 func (i StreamIterator) validInvariants() bool {
 	/////  1. streamIdx is less than the number of streams
 	return i.streamIdx < len(i.data) &&
 
-		// 2. stream is non-empty
+		// 2. stream is not pumping
+		!i.data[i.streamIdx].IsPumpStream() &&
```
keruch, line 115: don't do any pagination for pump streams

x/streamer/keeper/abci.go
```diff
@@ -9,6 +9,30 @@
 	"github.com/dymensionxyz/dymension/v3/x/streamer/types"
 )
 
+// BeginBlock processes pump streams for potential pump operations
+func (k Keeper) BeginBlock(ctx sdk.Context) error {
+	// Get all active streams
+	streams := k.GetActiveStreams(ctx)
```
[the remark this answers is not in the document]
omritoptix: cc @keruch

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+		}
+	}
+
+	return nil
+}
+
+// Number of seconds in the year.
+// 60 * 60 * 8766 is how the SDK defines it:
+// https://github.com/cosmos/cosmos-sdk/blob/v0.50.14/x/mint/types/params.go#L33
+const yearSecs = 60 * 8766
```
omritoptix: aren't you missing another 60 here?

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+	mintParams, err := k.mintParams.Get(ctx)
+	if err != nil {
+		return math.ZeroInt(), fmt.Errorf("get mint params: %w", err)
+	}
+	// info.Duration might be "hour", "day", or "week" and is defined as
+	// an integer, so it's safe to cast it to uint64.
+	var (
+		year          = math.NewInt(yearSecs)
+		blocksPerYear = math.NewIntFromUint64(mintParams.BlocksPerYear)
+		epochSecs     = math.NewInt(int64(info.Duration))
```
omritoptix: not sure I understand how it's epochSecs if info.Duration is in nano seconds?

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+// https://github.com/cosmos/cosmos-sdk/blob/v0.50.14/x/mint/types/params.go#L33
+const yearSecs = 60 * 8766
+
+func (k Keeper) EpochBlocks(ctx sdk.Context, epochID string) (math.Int, error) {
+	info := k.ek.GetEpochInfo(ctx, epochID)
+	mintParams, err := k.mintParams.Get(ctx)
+	if err != nil {
+		return math.ZeroInt(), fmt.Errorf("get mint params: %w", err)
+	}
+	// info.Duration might be "hour", "day", or "week" and is defined as
```
omritoptix, line 365: I think the description is hour, day or week, but this field is in nanoseconds

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+	randBig := GenerateUnifiedRandom(ctx, modulo.BigIntMut())
+	rand := math.NewIntFromBigIntMut(randBig)
+	return math.MinInt(rand, left), nil
+}
+
+// GenerateUnifiedRandom a unified random variable by modulo.
+func GenerateUnifiedRandom(ctx sdk.Context, modulo *big.Int) *big.Int {
+	h := sha256.New()
+	h.Write(ctx.HeaderHash())
+	// h.Write(blockTimeBytes)
```
omritoptix: why is this commented out? should we delete it?

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+
+	return math.ZeroInt(), nil
+}
+
+// PumpAmt computes min(Uniform(0, 2 * Budget / PumpNum), Left).
+func PumpAmt(ctx sdk.Context, budget math.Int, left math.Int, pumpNum uint64) (math.Int, error) {
+	modulo := budget.MulRaw(2).QuoRaw(int64(pumpNum))
+	if modulo.IsZero() {
+		return math.ZeroInt(), fmt.Errorf("budget per pump is fractional: too small budget (%s) or too many pumps (%d)", budget, pumpNum)
+	}
```
omritoptix: can be validated in pump creation vs here?

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+		g, err := k.ik.GetGaugeByID(ctx, gauge.GaugeId)
+		if err != nil {
+			k.Logger(ctx).Error("failed to get gauge", "gaugeID", gauge.GaugeId, "error", err)
+			continue
+		}
+		if ra := g.GetRollapp(); ra != nil {
+			rollappRecords = append(rollappRecords, types.PumpPressure{
+				RollappId: ra.RollappId,
+				// Don't pre-calculate 'pumpBudget / distr.VotingPower' bc it loses precision
+				Pressure: gauge.Power.Mul(pumpBudget).Quo(distr.VotingPower),
```
omritoptix: I think we need to normalize by voting power of only the rollapps not the entire gauges otherwise we won't give up all the pumpBuget.

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+
+	plan, found := k.iroKeeper.GetPlanByRollapp(ctx, rollapp.RollappId)
+	if !found {
+		return sdk.Coin{}, fmt.Errorf("IRO plan not found for rollapp: %s", rollappID)
+	}
+
+	buyer := k.ak.GetModuleAddress(types.ModuleName)
+
+	var tokenOutDenom string
+	if plan.IsSettled() {
```
omritoptix: actually this is not the indication. the plan can be not settled but the rollapp would already be in an AMM position (@mtsitrin new feature). @mtsitrin - what condition should be checked here?

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+		tokenOutDenom = plan.SettledDenom
+	} else {
+		tokenOutDenom = plan.LiquidityDenom
+	}
+	var tokenOutAmt math.Int
+
+	err = osmoutils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
+		// There are several cases:
+		// - targetDenom == plan.SettledDenom =>
```
omritoptix: plan.SettledDenom means the denom of the rollapp after the plan has been settled. this is what you mean here? I think not. I believe you mean the rollap denom pair for the pool (i.e DYM or USDC)?

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+	plan, found := k.iroKeeper.GetPlanByRollapp(ctx, rollapp.RollappId)
+	if !found {
+		return sdk.Coin{}, fmt.Errorf("IRO plan not found for rollapp: %s", rollappID)
+	}
+
+	buyer := k.ak.GetModuleAddress(types.ModuleName)
+
+	var tokenOutDenom string
+	if plan.IsSettled() {
+		tokenOutDenom = plan.SettledDenom
```
omritoptix: I believe this is wrong. you always want to ge the liquidity Denom as this will be the denom on the other side of the pool (or the denom for IRO)

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+				reverseRoute,
+				pumpAmt,        // token in
+				math.ZeroInt(), // no slippage
+			)
+			if err != nil {
+				return fmt.Errorf("route exact amount in: target denom: %s, error: %w", tokenOutDenom, err)
+			}
+		}
+
+		if !plan.IsSettled() {
```
omritoptix: as my previous comment - not the right check given the new feature worked by @mtsitrin

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,287 @@
...
+			sdk.NewCoin(baseDenom, pumpAmt), // token in
+			math.ZeroInt(),                  // no slippage)
+		)
+		if err != nil {
+			return fmt.Errorf("route exact amount in: target denom: %s, error: %w", targetDenom, err)
+		}
+
+		if !plan.IsSettled() {
+			// If IRO is in progress, use liquidity tokens to buy IRO tokens
+			tokenOutAmt, err = k.iroKeeper.BuyExactSpend(
```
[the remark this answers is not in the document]
omritoptix: cc @keruch

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+				fmt.Sprintf("%d", plan.Id),
+				buyer,
+				tokenOutAmt,    // amountToSpend
+				math.ZeroInt(), // no slippage
+			)
+			if err != nil {
+				return fmt.Errorf("buy from IRO %d: %w", plan.Id, err)
+			}
+			// In that case, tokenOut is an IRO token
+			tokenOutDenom = plan.GetIRODenom()
```
omritoptix: I don't get it. you got the tokens you bought in the `tokenOutAmt`. what is this for? also you don't use the `tokenOutAmt` when you return.

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+		}
+
+		return nil
+	})
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
```
omritoptix, line 253: are we not missing the case when the plan rollapp token is already in an AMM and buying from there the token?

x/streamer/keeper/pump_stream.go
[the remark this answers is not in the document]
keruch: it's outdated, `tokenOutAmt` is returned in `ExecutePump`

x/streamer/keeper/pump_stream.go
keruch: you are right, `EpochBlocks` broke while testing. 

I unified it to use milliseconds everywhere.

x/streamer/keeper/pump_stream.go
keruch: see above

x/streamer/keeper/pump_stream.go
keruch: see above

x/streamer/keeper/pump_stream.go
keruch: removed

x/streamer/keeper/pump_stream.go
keruch: in the next PR I will limit voting on non-rollapp gauges, so those numbers will always be normalized

x/streamer/keeper/pump_stream.go
keruch: target denom is a denom that we want to get after swapping:
- for IRO, it's an IRO denom
- for AMM, it's a RA denom (plan.SettledDenom)

> the rollap denom pair for the pool (i.e DYM or USDC)

what you are saying is plan.LiquidityDenom

x/streamer/keeper/pump_stream.go
keruch: if the plan is settled, it's not necessary to swap DYM to LiquidityDenom. every RA token is a fee token and has at least one swap route to DYM (see https://github.com/dymensionxyz/osmosis/pull/91), so we can do a multihop swap directly from DYM to SettledDenom.

---

to recap, we can
1. make two swaps: DYM -> LiquidityDenom -> SettledDenom
2. or make one multihop swap: DYM -> SettledDenom

x/streamer/keeper/pump_stream.go
keruch: I do use `tokenOutAmt` as a return value of `ExecutePump`. 

`tokenOutAmt` is an amount, we need to know what the denom is, so I get it from `plan.GetIRODenom()`.

x/streamer/keeper/pump_stream.go
keruch: no we don't :) the AMM buy is covered, I tested it specifically. 

anyway, guess the function turned out to be complicated. I simplified it in the last commit. Now there are two straightforward branches:
1. If AMM => swap DYM to SettledDenom
2. If IRO => if LiquidityDenom != DYM then swap DYM to LiquidityDenom first => use LiquidityDenom to buy IRO

x/streamer/keeper/abci.go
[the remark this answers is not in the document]
keruch: Not sure we need to fix it now. `GetActiveStreams` is a legacy method in x/streamer and we've been using it in `EndBlock` and in `AfterEpochEnd`/`BeforeEpochStart` (which are also called in `BeginBlock` in x/epochs). The only case when it panics is on deserialization problems.

x/streamer/keeper/pump_stream.go
keruch: `budget` is dynamic – changes on every new epoch, so can't be verified on the creation.

proto/dymensionxyz/dymension/streamer/streamer.proto
```diff
@@ -20,4 +20,13 @@ message EpochPointer {
   // have an ability to sort the EpochPointer slice.
   google.protobuf.Duration epoch_duration = 4
       [ (gogoproto.nullable) = false, (gogoproto.stdduration) = true ];
+}
+
+// PumpPressure is how much DYM will be used for buying RA tokens if pump occurs
+message PumpPressure {
+  string rollapp_id = 1;
+  string pressure = 2 [
```
mtsitrin, line 29: we usually define 
`(cosmos_proto.scalar) = "cosmos.Int",`
as well

x/streamer/keeper/abci.go
```diff
@@ -9,6 +9,27 @@ import (
 	"github.com/dymensionxyz/dymension/v3/x/streamer/types"
 )
 
+// BeginBlock processes pump streams for potential pump operations
+func (k Keeper) BeginBlock(ctx sdk.Context) error {
+	// Get all active streams
```
mtsitrin, line 14: pump stream can be `active` even before `UpdateStreamAtEpochStart`
as `moveUpcomingStreamToActiveStream` checked for any kind of epoch, but `UpdateStreamAtEpochStart` is called for defined `epochIdentifer` only
it can cause `DistributePumpStreams` call to uninitialized pump stream

mtsitrin: u should probablty filter out those as well

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+	// Always use base denom for budget
+	baseDenom, err := k.txFeesKeeper.GetBaseDenom(ctx)
+	if err != nil {
+		return fmt.Errorf("get base denom: %w", err)
+	}
+
+	for _, stream := range pumpStreams {
+		if !stream.IsPumpStream() {
```
mtsitrin, line 274: already validated by caller

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+			// Skip non-pump streams
+			continue
+		}
+
+		epochBlocks, err := k.EpochBlocks(ctx, stream.DistrEpochIdentifier)
```
mtsitrin, line 279: don't u prefer to store it for the stream on stream creation?

x/streamer/keeper/stream.go
```diff
@@ -18,6 +18,16 @@ func (k Keeper) UpdateStreamAtEpochStart(ctx sdk.Context, stream types.Stream) (
 	remainEpochs := stream.NumEpochsPaidOver - stream.FilledEpochs
 	epochCoins := remainCoins.QuoInt(math.NewIntFromUint64(remainEpochs))
 
+	if stream.IsPumpStream() {
+		// Pump streams should always have one coin with base denom
+		stream.PumpParams.EpochBudget = epochCoins[0].Amount
+		stream.PumpParams.EpochBudgetLeft = stream.PumpParams.EpochBudget
```
mtsitrin, line 24: what happens with the `EpochBudgetLeft` of the last epoch?
we might have a finished stream with remaining coins

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+	if modulo.IsZero() {
+		return math.ZeroInt(), fmt.Errorf("budget per pump is fractional: too small budget (%s) or too many pumps (%d)", budget, pumpNum)
+	}
+	randBig := GenerateUnifiedRandom(ctx, modulo.BigIntMut())
+	rand := math.NewIntFromBigIntMut(randBig)
+	return math.MinInt(rand, left), nil
+}
+
+// GenerateUnifiedRandom a unified random variable by modulo.
+func GenerateUnifiedRandom(ctx sdk.Context, modulo *big.Int) *big.Int {
```
mtsitrin: why not having a copy of `modulo` instead of mutating it?

also, should probably move to `utils/rand` as it will probably be reused by other modules

x/streamer/keeper/pump_stream.go
keruch: x/mint param could be changed and thus block production might become faster or slower. so the number of blocks in the epochs could also change dynamically.

x/streamer/keeper/pump_stream.go
keruch: will do in the next PR: https://github.com/dymensionxyz/dymension/pull/1988

x/streamer/keeper/stream.go
keruch: it's fine. x/streamer will have excessive balance that can be used for further streams.

x/streamer/keeper/pump_stream.go
keruch: just a sanity check to prevent accidental panic

x/streamer/keeper/abci.go
keruch: you are right. guess that's also the case for usual streams. IIRC we've always had streams with the same epoch ID so far, so it wasn't a problem. 

I updated a condition in `BeforeEpochStart`. now we will call `moveUpcomingStreamToActiveStream ` only if 
```
ctx.BlockTime().After(s.StartTime) && s.DistrEpochIdentifier == epochIdentifier
```

--- 

UPD: apparently this is the expected behavior bc tests started failing when I added this condition.. instead of the condition, I make sure that the pump stream is initialized and ready to go right after the creation

x/streamer/keeper/pump_stream.go
keruch: done https://github.com/dymensionxyz/dymension/pull/1990

proto/dymensionxyz/dymension/streamer/streamer.proto
mtsitrin: pls add in other pump related protos

x/streamer/keeper/stream_iterator.go
mtsitrin: I think we can skip this entire distribution function for pump streams. 
instead of calling `Distribute` for all `ActiveStreams`, we can filter it out beforehand.

we can do it later in sepeate issue

x/streamer/keeper/stream.go
mtsitrin: I think this balance will be locked and cannot be reused
we check for 
`alreadyAllocatedCoins := k.GetModuleToDistributeCoins(ctx) `
on new stream creation.

undistributed tokens considered `alreadyAllocated`

x/streamer/keeper/pump_stream.go
mtsitrin: we will lose pumpBuget anyway in case `NumTopRollapps` < `num of voted rollapps`

I think we need to get the `NumTopRollapps` and normalize their total power after iterating over `NumTopRollapps`

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+
+		if pumpAmt.IsZero() {
+			// Shouldn't pump on this iteration
+			continue
+		}
+
+		// Get top N rollapps by cast voting power
+		pressure := k.PumpPressure(ctx, sponsorshipDistr, pumpAmt)
+		if len(pressure) > int(stream.PumpParams.NumTopRollapps) {
+			pressure = pressure[:int(stream.PumpParams.NumTopRollapps)]
```
mtsitrin: why not doing it inside `PumpPressure`?

proto/dymensionxyz/dymension/streamer/events.proto
```diff
@@ -26,4 +26,20 @@ message EventEpochEnd {
...
+      (gogoproto.customtype) = "cosmossdk.io/math.Int",
+      (gogoproto.nullable) = false
+    ];
+    cosmos.base.v1beta1.Coin tokenOut = 4 [
+      (gogoproto.nullable) = false,
+      (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coin"
+    ];
+  }
+  repeated Pump pumps = 1 [ (gogoproto.nullable) = false ];
+}
```
mtsitrin: I think it should be:

```suggestion
message EventPumped {
  message Pump {
    string rollapp_id = 1;
    string pumpAmt = 2 [
      (gogoproto.customtype) = "cosmossdk.io/math.Int",
      (gogoproto.nullable) = false
    ];
    cosmos.base.v1beta1.Coin tokenOut = 3 [
      (gogoproto.nullable) = false,
      (gogoproto.castrepeated) = "github.com/cosmos/cosmos-sdk/types.Coin"
    ];
  }

  uint64 stream_id = 1;
  repeated Pump pumps = 2 [ (gogoproto.nullable) = false ];
}
```

and have single event per stream

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+			//   2. Every RA token has an AMM route to DYM
+			//   3. Do an AMM swap: PumpDenom -> SettledDenom
+			//   4. Return Coin(SettledDenom; AMM token out)
+			tokenOutDenom := plan.SettledDenom
+			feeToken, err := k.txFeesKeeper.GetFeeToken(ctx, tokenOutDenom)
+			if err != nil {
+				return fmt.Errorf("get fee token for denom %s: %w", tokenOutDenom, err)
+			}
+
+			reverseRoute := reverseInRoute(feeToken.Route, tokenOutDenom)
```
mtsitrin, line 195: u'll have the poolID in the plan struct. u'll need the swap only if liquidity denom != base denom as below

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,370 @@
...
+
+	plan, found := k.iroKeeper.GetPlanByRollapp(ctx, rollapp.RollappId)
+	if !found {
+		return sdk.Coin{}, fmt.Errorf("IRO plan not found for rollapp: %s", rollappID)
+	}
+
+	buyer := k.ak.GetModuleAddress(types.ModuleName)
+
+	err = osmoutils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
+		if plan.IsSettled() {
```
mtsitrin, line 183: additional case is where plan.IsGraduated
it's similar to settled, but the pool denom is `plan.GetIRODenom()` instead of `plan.SettledDenom`

utils/rand/random.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Get 32 bytes hash from it
+	h := sha256.New()
+	h.Write(hash)
+	seed := [32]byte(h.Sum(nil))
+
+	// Use a PRNG to generate a float in [0; 1)
+	//nolint:gosec // math/rand/v2 is used for purpose
+	generator := rand.New(rand.NewChaCha8(seed))
+	return big.NewFloat(generator.Float64())
```
mtsitrin: don't use floats, it can cause non-determinism
uniform distribution can be easly done without it
(using `generator.Uint64()` + `.Mod()`
or
`generator.uint64n()` directly
)

mtsitrin (changes requested):

utils/rand/random.go
```diff
@@ -0,0 +1,78 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GenerateUnifiedRandom draws a uniform random variable in [0; 1)
+func GenerateUnifiedRandom(ctx sdk.Context) *big.Float {
+	// Take block hash
+	hash := ctx.HeaderInfo().Hash
+
+	// Get 32 bytes hash from it
```
mtsitrin: why we need rehashing?

x/streamer/keeper/pump_stream.go
keruch: moved

utils/rand/random.go
keruch: you are right, reworked with `osmomath.BigDec`. see https://github.com/dymensionxyz/dymension/pull/1981/commits/9083e4f32b7ab39367a49b00f527856d46372d80

utils/rand/random.go
keruch: I thought it was going to prevent potential patterns in the block hash, but it doesn't matter since I still use ChaCha8. removed! see https://github.com/dymensionxyz/dymension/pull/1981/commits/9083e4f32b7ab39367a49b00f527856d46372d80

proto/dymensionxyz/dymension/streamer/events.proto
keruch: done

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+
+	return math.ZeroInt(), nil
+}
+
+// PumpAmt computes the random pump amount.
+func PumpAmt(ctx sdk.Context, pumpParams types.PumpParams) (math.Int, error) {
+	numPumps := math.NewIntFromUint64(pumpParams.NumPumps)
+
+	if pumpParams.EpochBudget.LT(numPumps) {
```
mtsitrin, line 134: shouldn't it be checked when setting `EpochBudget`? instead of returning error and messing up the entire pump distribuition

x/streamer/keeper/pump_stream.go
```diff
@@ -0,0 +1,374 @@
...
+		return math.ZeroInt(), fmt.Errorf("budget per pump is fractional: too small budget (%s) or too many pumps (%d)", pumpParams.EpochBudget, pumpParams.NumPumps)
+	}
+
+	randBig := new(big.Int)
+
+	switch pumpParams.PumpDistr {
+	case types.PumpDistr_PUMP_DISTR_UNIFORM:
+		// Draw a Uniform(0; 2*B/N) value
```
mtsitrin, line 145: why not `Uniform(0.5 B/N; 2*B/N)`? makes more sense IMO, if u passed the `ShouldPump` probability

x/streamer/keeper/stream.go
keruch: when the stream finishes, the coins are unlocked and can be reused.

`GetModuleToDistributeCoins` doesn't account for finished streams, only upcoming and active:

https://github.com/dymensionxyz/dymension/blob/14877e9ef3b3f404e17f8a85df2bbda22d3ad4d6/x/streamer/keeper/distributed_coins.go#L10-L15

so when we create a new stream, we can reuse those coins:

https://github.com/dymensionxyz/dymension/blob/a1572407f5dac93ec860b63827b89482edd0095a/x/streamer/keeper/keeper.go#L150-L154

x/streamer/keeper/pump_stream.go
keruch: guess that's more a product-related q. @omritoptix what do you think? 

anyway I think @mtsitrin is right. I added the normalization.

x/streamer/keeper/stream_iterator.go
keruch: yeah, let's do later. refactoring it in this PR will be too dirty. created an issue: https://github.com/dymensionxyz/dymension/issues/1996.

x/streamer/keeper/pump_stream.go
keruch: it's not an issue actually. it means that the budget is too low to distribute anything. I updated the code and left a comment.

```
	if pumpParams.EpochBudget.LT(numPumps) {
		// The budget is too small to use it for pumping. It might happen
		// close to the stream end if the epoch budget is too low, but it's
		// not probable since EpochBudget ≈ 10^20, numPumps ≈ 10^5.
		return math.ZeroInt(), nil
	}
```

x/streamer/keeper/pump_stream.go
keruch: you mean not to distribute something close to 0? it's not a problem IMO. 

say we want to pump in 10 blocks during the epoch, eg 10 times a day
we distribute 10 000 DYM per epoch => 1 000 * 10^18 aDYM per pump

with those numbers, the probability of the outcome being less than 1 DYM (10^18) is 10^18/10^21 = 0.001 = 0.1% which is pretty small.

I discussed this with Omri, he said it makes sense that the random is random and you can get either 0 or 'max pump' with some low probability.

x/streamer/keeper/pump_stream.go
keruch: agreed on normalization

utils/rand/random.go
```diff
@@ -0,0 +1,52 @@
...
+	generator := rand.NewChaCha8([32]byte(seed))
+	return new(big.Int).SetUint64(generator.Uint64())
+}
+
+// GenerateUniformRandomMod draws a uniform random variable in [0; modulo)
+func GenerateUniformRandomMod(ctx sdk.Context, modulo *big.Int) *big.Int {
+	// Generate a uniform random variable in [0; MaxUint64)
+	u := GenerateUniformRandom(ctx)
+	// Normalize it to [0; 1) = uniform / (MAX_UINT64 + 1)
+	un := osmomath.NewDecFromBigInt(u).QuoInt(osmomath.NewIntFromUint64(math.MaxUint64).AddRaw(1))
```
mtsitrin: can use `BigMaxUint64`

utils/rand/random.go
```diff
@@ -0,0 +1,52 @@
...
+	// Take block hash. Tendermint uses SHA-256.
+	seed := ctx.HeaderInfo().Hash
+	//nolint:gosec // math/rand/v2 is used for a purpose: we need a custom seed
+	generator := rand.NewChaCha8([32]byte(seed))
+	return new(big.Int).SetUint64(generator.Uint64())
+}
+
+// GenerateUniformRandomMod draws a uniform random variable in [0; modulo)
+func GenerateUniformRandomMod(ctx sdk.Context, modulo *big.Int) *big.Int {
+	// Generate a uniform random variable in [0; MaxUint64)
```
mtsitrin:
```suggestion
	// Generate a uniform random variable in [0; MaxUint64]
```

utils/rand/random_test.go
```diff
@@ -0,0 +1,80 @@
...
+	slices.Sort(values)
+	for _, v := range values {
+		println(v)
+	}
+	t.Log("Target mean", modulo.QuoRaw(2))
+	t.Log("Actual mean", total.QuoRaw(iterations))
+}
+
+func TestExpRandom(t *testing.T) {
+	//t.Skip("This test is for debugging and visualizing the distribution.")
```
mtsitrin: comment back?

utils/rand/random_test.go
mtsitrin, line 1: can't we add assertions to those test cases? to check our distribution acts as expected

x/streamer/keeper/abci.go
keruch: created an issue https://github.com/dymensionxyz/dymension/issues/1997

utils/rand/random_test.go
keruch: https://github.com/dymensionxyz/dymension/issues/1998

mtsitrin (approved):

PR #1986, opened 2025-09-01
feat(IRO): graduate completed IROs into amm pool

mtsitrin:
* added `CreateFairLaunchPlan` to create a fair launch IRO plan
* added graduation flow, where IRO graduates into AMM pool once all tokens sold

### DIFF
--- a/x/incentives/keeper/genesis.go
+++ b/x/incentives/keeper/genesis.go
@@ -11,7 +11,6 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
 	k.SetParams(ctx, genState.Params)
 	k.SetLockableDurations(ctx, genState.LockableDurations)
 	for _, gauge := range genState.Gauges {
-		gauge := gauge
 		err := k.SetGaugeWithRefKey(ctx, &gauge)
 		if err != nil {
 			panic(err)
--- a/x/rollapp/keeper/block_height_to_finalization_queue_test.go
+++ b/x/rollapp/keeper/block_height_to_finalization_queue_test.go
@@ -55,7 +55,6 @@ func (s *RollappTestSuite) TestBlockHeightToFinalizationQueueGet() {
 	ctx := s.Ctx
 	items := createNBlockHeightToFinalizationQueue(k, ctx, 10)
 	for _, item := range items {
-		item := item
 		rst, found := k.GetFinalizationQueue(
 			ctx,
 			item.CreationHeight,
--- a/x/rollapp/keeper/latest_finalized_state_index_test.go
+++ b/x/rollapp/keeper/latest_finalized_state_index_test.go
@@ -37,7 +37,6 @@ func TestLatestFinalizedStateIndexGet(t *testing.T) {
 	keeper, ctx := keepertest.RollappKeeper(t)
 	items := createNLatestFinalizedStateIndex(keeper, ctx, 10)
 	for _, item := range items {
-		item := item
 		rst, found := keeper.GetLatestFinalizedStateIndex(ctx,
 			item.RollappId,
 		)
--- a/x/rollapp/keeper/latest_state_info_index_test.go
+++ b/x/rollapp/keeper/latest_state_info_index_test.go
@@ -30,7 +30,6 @@ func TestLatestStateInfoIndexGet(t *testing.T) {
 	keeper, ctx := keepertest.RollappKeeper(t)
 	items := createNLatestStateInfoIndex(keeper, ctx, 10)
 	for _, item := range items {
-		item := item
 		rst, found := keeper.GetLatestStateInfoIndex(ctx,
 			item.RollappId,
 		)
--- a/x/rollapp/keeper/state_info_test.go
+++ b/x/rollapp/keeper/state_info_test.go
@@ -42,7 +42,6 @@ func TestStateInfoGet(t *testing.T) {
 	k, ctx := keepertest.RollappKeeper(t)
 	items, _ := createNStateInfo(k, ctx, 10)
 	for _, item := range items {
-		item := item
 		rst, found := k.GetStateInfo(ctx,
 			item.StateInfoIndex.RollappId,
 			item.StateInfoIndex.Index,
--- a/x/rollapp/types/block_descriptor_test.go
+++ b/x/rollapp/types/block_descriptor_test.go
@@ -49,7 +49,6 @@ func TestBlockDescriptorsValidate(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			err := tc.bds.Validate()
 			if tc.expPass {
@@ -85,7 +84,6 @@ func TestBlockDescriptorValidate(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			err := tc.bd.Validate()
 			if tc.expPass {
--- a/x/sequencer/keeper/get_and_set_test.go
+++ b/x/sequencer/keeper/get_and_set_test.go
@@ -13,7 +13,6 @@ func TestSequencerGet(t *testing.T) {
 	k, ctx := keepertest.SequencerKeeper(t)
 	items := createNSequencers(k, ctx, 10)
 	for _, item := range items {
-		item := item
 		rst, err := k.RealSequencer(ctx,
 			item.Address,
 		)


x/iro/keeper/create_plan.go
```diff
@@ -114,6 +114,143 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+
+	// Check if a plan already exists for the rollapp
+	_, found = m.GetPlanByRollapp(ctx, rollapp.RollappId)
+	if found {
+		return nil, errors.Join(gerrc.ErrFailedPrecondition, types.ErrPlanExists)
+	}
+
+	// Validate 100% IRO allocation
+	if !rollapp.GenesisInfo.InitialSupply.Equal(params.FairLaunch.AllocationAmount) || len(rollapp.GenesisInfo.Accounts()) != 1 {
+		return nil, errorsmod.Wrapf(gerrc.ErrFailedPrecondition, "rollapp must have 100%% IRO allocation: expected %s, got %s", params.FairLaunch.AllocationAmount, rollapp.GenesisInfo.InitialSupply)
```
keruch:
```suggestion
		return nil, errorsmod.Wrapf(gerrc.ErrFailedPrecondition, "rollapp must have 100% IRO allocation: expected %s, got %s", params.FairLaunch.AllocationAmount, rollapp.GenesisInfo.InitialSupply)
```

x/iro/keeper/create_plan.go
```diff
@@ -114,6 +114,143 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+	// This is needed because params.FairLaunch.TargetRaise might be in a different denom
+	convertedTargetRaise, err := m.convertTargetRaiseToLiquidityDenom(ctx, params.FairLaunch.TargetRaise, req.LiquidityDenom)
+	if err != nil {
+		return nil, errorsmod.Wrapf(gerrc.ErrInvalidArgument, "failed to convert target raise to liquidity denom: %v", err)
+	}
+
+	// Calculate M parameter for the bonding curve
+	// Convert amounts to decimal representation for calculation
+	allocationDec := types.ScaleFromBase(params.FairLaunch.AllocationAmount, int64(rollapp.GenesisInfo.NativeDenom.Exponent))
+	evaluationDec := types.ScaleFromBase(convertedTargetRaise.Amount, int64(liqTokenExponent)).MulInt64(2)
```
keruch, line 202: why `MulInt64(2)`?

x/iro/keeper/create_plan.go
```diff
@@ -114,6 +114,143 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+	if err != nil {
+		return nil, err
+	}
+	plan := m.MustGetPlan(ctx, planId)
+	plan.FairLaunched = true
+	m.SetPlan(ctx, plan)
+
+	if err := plan.ValidateBasic(); err != nil {
+		return nil, errors.Join(gerrc.ErrInvalidArgument, err)
+	}
```
keruch: why validate the plan after creating and saving it to the store?

x/iro/keeper/create_plan.go
```diff
@@ -143,10 +280,6 @@ func (k Keeper) CreatePlan(ctx sdk.Context, liquidityDenom string, allocatedAmou
 		plan.EnableTradingWithStartTime(startTime)
 	}
 
-	if err := plan.ValidateBasic(); err != nil {
-		return "", errors.Join(gerrc.ErrInvalidArgument, err)
-	}
```
keruch: why? looks like a good place for validation

x/iro/keeper/params.go
```diff
@@ -23,6 +25,47 @@ func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParam
...
+	targetRaiseExponent := targetRaiseMetadata.DenomUnits[len(targetRaiseMetadata.DenomUnits)-1].Exponent
+
+	// validate target raise denom is allowed
+	if !slices.Contains(m.Keeper.gk.GetParams(ctx).AllowedPoolCreationDenoms, targetRaise.Denom) {
+		return nil, errorsmod.Wrap(gerrc.ErrFailedPrecondition, "denom not allowed")
+	}
+
+	// validate curve is valid
+	allocationDec := types.ScaleFromBase(fairLaunch.AllocationAmount, 18)
+	evaluationDec := types.ScaleFromBase(targetRaise.Amount, int64(targetRaiseExponent)).MulInt64(2)
```
keruch, line 48: why `MulInt64(2)`?

x/iro/types/params.go
```diff
@@ -18,10 +19,16 @@ var (
 	DefaultMinLiquidityPart                             = "0.4"                       // default: at least 40% goes to the liquidity pool
 	DefaultMinVestingDuration                           = 7 * 24 * time.Hour          // default: min 7 days
 	DefaultMinVestingStartTimeAfterSettlement           = 0 * time.Minute             // default: no enforced minimum by default
+
+	DefaultFairLaunch = FairLaunch{
+		AllocationAmount: math.NewInt(1e9).MulRaw(1e18),                      // 1B RA tokens
+		TargetRaise:      sdk.NewCoin("adym", math.NewInt(2e4).MulRaw(1e18)), // 20K DYM
```
keruch: there is a const for adym in `github.com/dymensionxyz/dymension/v3/app/params`
```suggestion
		TargetRaise:      sdk.NewCoin(params.BaseDenom, math.NewInt(2e4).MulRaw(1e18)), // 20K DYM
```

x/iro/keeper/create_plan.go
```diff
@@ -143,10 +280,6 @@ func (k Keeper) CreatePlan(ctx sdk.Context, liquidityDenom string, allocatedAmou
...
-
 	err = k.rk.SetIROPlanToRollapp(ctx, &rollapp, plan)
```
keruch: you don't set `plan.FairLaunch = true` anywhere here, but `SetIROPlanToRollapp` depends on this field

proto/dymensionxyz/dymension/iro/tx.proto
```diff
@@ -111,6 +114,15 @@ message MsgCreatePlan {
   ];
 }
 
+message MsgCreateFairLaunchPlan {
+  option (cosmos.msg.v1.signer) = "owner";
+
+  string owner = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
+  string rollapp_id = 2;
+  bool trading_enabled = 3;
```
keruch, line 122: why do you need `trading_enabled` for the fair launch plan? from the comments, I thought that the trading enabled concept and fair launch are interchangeable. for example, here:

https://github.com/dymensionxyz/dymension/blob/f36a5b0bf96117325edb9fb61fb86fdc0133eadf/x/rollapp/keeper/rollapp.go#L143-L149

x/iro/keeper/graduation.go
```diff
@@ -0,0 +1,123 @@
...
+	k.SetPlan(ctx, plan)
+
+	rollapp, found := k.rk.GetRollapp(ctx, plan.RollappId)
+	if !found {
+		return 0, nil, errorsmod.Wrap(gerrc.ErrFailedPrecondition, "rollapp not found")
+	}
+	// graduated plans can be launched, thus we need to update the pre launch time
+	if rollapp.PreLaunchTime.After(ctx.BlockTime()) {
+		k.rk.SetPreLaunchTime(ctx, &rollapp, ctx.BlockTime())
+	}
```
keruch, line 44: I don't really understand how `TradingEnabled` and `FairLaunch` should work with `rollapp.PreLaunchTime` and `plan.PreLaunchTime`. Seems like `rollapp.PreLaunchTime` and `plan.PreLaunchTime` are not in sync. I'm seeing conditions everywhere and the overall logic isn't clicking for me..

Let's say you are creating a fair launch plan.

---

Say `TradingEnabled == true` and `FairLaunch == true`.  

(in fact that couldn't happen since `FairLaunch` is always `false` there now, see my comment above)

You call `CreatePlan` and `SetIROPlanToRollapp`. `plan.PreLaunchTime == ctx.BlockTime()` but due to the condition `rollapp.PreLaunchTime == 10 years`.

https://github.com/dymensionxyz/dymension/blob/f36a5b0bf96117325edb9fb61fb86fdc0133eadf/x/rollapp/keeper/rollapp.go#L146-L149

---

Say `TradingEnabled == true` and `FairLaunch == false`. 

(that's possible now)

You call `CreatePlan` and `SetIROPlanToRollapp`. `plan.PreLaunchTime == ctx.BlockTime()` (in `CreatePlan`) and `rollapp.PreLaunchTime == ctx.BlockTime()` (in `SetIROPlanToRollapp`). 

---

Say `TradingEnabled == false` and `FairLaunch == true`. 

(again, that couldn't happen since `FairLaunch` is always `false` there now)

You call `CreatePlan` and `SetIROPlanToRollapp`. `rollapp.PreLaunchTime == 10 years` (condition in `SetIROPlanToRollapp`), `plan.PreLaunchTime` - ??

---

Say `TradingEnabled == false` and `FairLaunch == false`.

(that's possible now)

You call `CreatePlan` and `SetIROPlanToRollapp`. `rollapp.PreLaunchTime == 10 years` (condition in `SetIROPlanToRollapp`), `plan.PreLaunchTime` - ??

---

In any case, `FairLaunch == true` after you create a plan, so `EnableTranding` msg can't be used.
https://github.com/dymensionxyz/dymension/blob/b901d95a3fb9d800116ef653dd9eeab5f8a48278/x/iro/keeper/create_plan.go#L236

x/iro/keeper/trade.go
```diff
@@ -111,6 +124,23 @@ func (k Keeper) Buy(ctx sdk.Context, planId string, buyer sdk.AccAddress, amount
 		return err
 	}
 
+	// if all tokens are sold, we need to graduate the plan
+	if plan.SoldAmt.Equal(plan.MaxAmountToSell) {
```
keruch, line 128: do we need to do the same for `BuyExactSpend`?

x/iro/types/types.go
```diff
@@ -1 +1,46 @@
...
+const (
+	GraduationStatus_PRE_GRADUATION GraduationStatus = 0
+	GraduationStatus_POOL_CREATED   GraduationStatus = 1
+	GraduationStatus_SETTLED        GraduationStatus = 2
+)
+
+// graduation status string
+var GraduationStatusString = map[GraduationStatus]string{
+	GraduationStatus_PRE_GRADUATION: "pre_graduation",
+	GraduationStatus_POOL_CREATED:   "pool_created",
```
keruch: maybe `GraduationStatus_GRADUATED`? `pool_created` is a bit confusing in this context (as for `settled` the pool is also created).

x/iro/keeper/settle.go
keruch, line 1: The plan is settled once the genesis transfer arrives, correct? Just curios, is there a way to settle **before** graduation in case of the fair launch? IIRC, you can't run a sequencer during IRO, and with the fair launch the IRO is not over until required tokens are sold. Therefore, one cannot settle before graduation, am I right?

x/iro/keeper/settle.go
```diff
@@ -47,40 +45,49 @@ func (k Keeper) Settle(ctx sdk.Context, rollappId, rollappIBCDenom string) error
...
-	// mark the plan as `settled`, allowing users to claim tokens
-	plan.SettledDenom = rollappIBCDenom
+	// commit the settled plan
 	k.SetPlan(ctx, plan)
 
-	// uses the raised liquidity and unsold tokens to bootstrap the rollapp's liquidity pool
-	poolID, gaugeID, err := k.bootstrapLiquidityPool(ctx, plan, poolTokens)
+	// burn all the remaining IRO token.
+	iroTokenBalance := k.BK.GetBalance(ctx, k.AK.GetModuleAddress(types.ModuleName), plan.TotalAllocation.Denom)
+	err = k.BK.BurnCoins(ctx, types.ModuleName, sdk.NewCoins(iroTokenBalance))
```
keruch: in the fair launch, there are also leftovers in LiquidityDenom, right? 
https://github.com/dymensionxyz/dymension/blob/5956d10d16bfdce10f9983c3848b85b4d31bf863/x/iro/keeper/graduation.go#L22

do we need to burn them? maybe burn them right away during the graduation?

x/iro/keeper/trade.go
```diff
@@ -111,6 +124,23 @@ func (k Keeper) Buy(ctx sdk.Context, planId string, buyer sdk.AccAddress, amount
...
+
+		err = uevent.EmitTypedEvent(ctx, &types.EventGraduation{
+			PlanId:    planId,
+			RollappId: plan.RollappId,
+			PoolId:    poolID,
+		})
+		if err != nil {
+			return err
+		}
+	}
```
keruch, line 142: how is it gonna work for non-fair launched plans (including the existing ones)? if they are non-fair launched, they might have leftovers that used to be incentives for the pool, but those leftovers will be lost.

x/iro/keeper/migrations.go
keruch, line 1: we also need to properly migrate rollapps that sold out their IRO tokens but haven't been settled. they should be graduated now.

x/iro/keeper/create_plan.go
```diff
@@ -114,6 +114,143 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+	liqTokenExponent := liqToken.DenomUnits[len(liqToken.DenomUnits)-1].Exponent
+
+	// check liquidity denom is allowed
+	if !slices.Contains(m.Keeper.gk.GetParams(ctx).AllowedPoolCreationDenoms, req.LiquidityDenom) {
+		return nil, errorsmod.Wrap(gerrc.ErrFailedPrecondition, "denom not allowed")
+	}
+
+	// Convert target raise from its original denom to the requested liquidity denom
+	// This is needed because params.FairLaunch.TargetRaise might be in a different denom
+	convertedTargetRaise, err := m.convertTargetRaiseToLiquidityDenom(ctx, params.FairLaunch.TargetRaise, req.LiquidityDenom)
```
keruch: what if target raise increases or decreases while IRO is in progress due to the exchange rate between `TargetRaise.Denom` and `LiquidityDenom`?

x/iro/keeper/create_plan.go
mtsitrin: it's on purpose, it escapes the `%` for the format string

x/iro/keeper/create_plan.go
mtsitrin: because the curve's M is calculated with `total evaluation`, which is 2x the target raise
If u want to raise 100$, u expect the pool to be opened eventually with:
< 100$, X unsold tokens (which are worth 100$) >

x/iro/keeper/create_plan.go
mtsitrin: yeah because the validation is dependent on `fairLaunch` param, which I didn't add to the `CreatePlan` API
I though it will be easier to review, as changing the `CreatePlan` API introduces changes all over

x/iro/keeper/params.go
mtsitrin: to calc evalution from target raise
https://github.com/dymensionxyz/dymension/pull/1986#discussion_r2328551371

x/iro/keeper/create_plan.go
mtsitrin: I'll fix it

x/iro/keeper/create_plan.go
mtsitrin: fixed

x/iro/keeper/trade.go
mtsitrin: not really
`BuyExactSpend` uses approximation, so it probably won't get you to 100%.
Frontend will use `Buy` when it needs to reach 100%

x/iro/keeper/settle.go
mtsitrin: Yes, it's the product requirement. 
Can't launch fair launch rollapp, which wasn't graduated

x/iro/keeper/settle.go
mtsitrin: not supposed to. might be dust. tested in UT as well

x/iro/keeper/trade.go
mtsitrin: If a plan graduates, regardless of fair-launched or not, it shouldn't have leftovers
That's how the "MaxAmountToSell` point is defined
leftovers are expected for a plan settled before graduation (not fair launched obviously, as it's not possible)

proto/dymensionxyz/dymension/iro/tx.proto
mtsitrin: i'll rephrase the comment
- non tradable Plans => rollapp not launchable (waits for `plan.Duration` since start)
- fair launch plans => rollapp not launchable (waits for graduation)

it;s not interchangable
one can create a fair launch plan today, and "enable" it next week

x/iro/keeper/graduation.go
mtsitrin: logic should be
* custom IROs
* * if trading enabled - pre launch is IRO duration (set on `SetIROPlanToRollapp`)
* * if trading disabled - pre launch starts in the future, updated on `MsgEnableTrading`
* * rollapp luanchable on graduation anyway

* fair launch IROs
* * starts with pre launch in the future
* * rollapp launchable on graduation

x/iro/keeper/create_plan.go
```diff
@@ -127,13 +265,14 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+func (k Keeper) CreatePlan(ctx sdk.Context, liquidityDenom string, allocatedAmount math.Int, planDuration time.Duration, startTime time.Time, tradingEnabled bool, fairLaunched bool, rollapp rollapptypes.Rollapp, curve types.BondingCurve, incentivesParams types.IncentivePlanParams, liquidityPart math.LegacyDec, vestingDuration, vestingStartTimeAfterSettlement time.Duration) (string, error) {
 	allocation, err := k.MintAllocation(ctx, allocatedAmount, rollapp.RollappId, rollapp.GenesisInfo.NativeDenom.Display, uint64(rollapp.GenesisInfo.NativeDenom.Exponent))
 	if err != nil {
 		return "", err
 	}
 
 	plan := types.NewPlan(k.GetNextPlanIdAndIncrement(ctx), rollapp.RollappId, liquidityDenom, allocation, curve, planDuration, incentivesParams, liquidityPart, vestingDuration, vestingStartTimeAfterSettlement)
+	plan.FairLaunched = fairLaunched
```
omritoptix: nit, but - fair launch is very confusing name. cause I can create a plan which is fair launch (i.e all tokens go to the curve) but it won't set fair launch to `true`. that's why I am calling it V2. or standard launch. but fair launch is just wrong technically. I had to trace back the code to understand it.

x/iro/keeper/create_plan.go
```diff
@@ -127,13 +265,14 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
 
 	// if trading enabled initially, set start time and pre-launch time
 	if tradingEnabled {
```
omritoptix, line 278: I don't understand this part. if trading Enabled - why do we check start time is before block time? if it's before, what's the problem?

x/rollapp/keeper/rollapp.go
```diff
@@ -140,14 +140,10 @@ func (k Keeper) SetIROPlanToRollapp(ctx sdk.Context, rollapp *types.Rollapp, pla
...
 
-	// set pre launch time
-	// if not enabled yet, set preLaunchTime to the future
-	preLaunchTime := plan.PreLaunchTime
-	if !plan.TradingEnabled {
-		preLaunchTime = ctx.BlockTime().Add(time.Hour * 24 * 365 * 10) // 10 years
-	}
-
+	// set rollapp as unlaunchable until IRO's requirement achieved
+	preLaunchTime := ctx.BlockTime().Add(time.Hour * 24 * 365 * 10) // 10 years
```
omritoptix, line 144: if we do it like this (i.e every chain with IRO (v1 and v2) needs to graduate) - than why do we have this pre launch time at all? we can just block graduation if plan doesn't have `GraduatedPoolId`.

x/iro/keeper/create_plan.go
```diff
@@ -44,7 +44,7 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
 
 	// check minimal plan duration
 	if req.IroPlanDuration < params.MinPlanDuration {
```
omritoptix, line 46: why is plan duration needed if custom rollapps can't launch before IRO is graduated? (afaiu from the code)

x/iro/keeper/create_plan.go
```diff
@@ -109,7 +109,145 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+	if gAcc.Address != m.GetModuleAccountAddress() || !gAcc.Amount.Equal(params.FairLaunch.AllocationAmount) {
+		return nil, errorsmod.Wrap(gerrc.ErrFailedPrecondition, "allocated amount mismatch")
+	}
+
+	// Validate liquidity denom is registered and allowed
+	liqToken, ok := m.BK.GetDenomMetaData(ctx, req.LiquidityDenom)
+	if !ok {
+		return nil, errorsmod.Wrapf(gerrc.ErrInvalidArgument, "denom %s not registered", req.LiquidityDenom)
+	}
+	liqTokenExponent := liqToken.DenomUnits[len(liqToken.DenomUnits)-1].Exponent
```
omritoptix, line 185: is it safe? i.e - can there be a denom unit where the smallest decimal is not the last in the list of is it enforced?

x/iro/keeper/create_plan.go
```diff
@@ -109,7 +109,145 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+	}
+
+	// Create plan using global FairLaunch parameters
+	planId, err := m.Keeper.CreatePlan(
+		ctx,
+		req.LiquidityDenom,
+		params.FairLaunch.AllocationAmount,
+		0,           // no minimum plan duration
+		time.Time{}, // no start time
+		req.TradingEnabled,
```
omritoptix, line 231: in which case does trading enabled shouldn't be `true` here? why is it part of the request?

x/rollapp/keeper/rollapp.go
omritoptix: Ok now I saw that on the plan creation code you override the preLaunchTime if the plan is custom. nit - I think it's cumbersome to have those 2 pieces of code in 2 different places. I'd make sure they are in the same place for clarity.

x/iro/keeper/create_plan.go
```diff
@@ -177,17 +322,30 @@ func (k Keeper) CreatePlan(ctx sdk.Context, liquidityDenom string, allocatedAmou
...
+// If the denoms are the same, it returns the original target raise
+// If they're different, it attempts to find a conversion path or returns an error
+func (m msgServer) convertTargetRaiseToLiquidityDenom(ctx sdk.Context, targetRaise sdk.Coin, liquidityDenom string) (sdk.Coin, error) {
+	// if the target raise denom is the same as the liquidity denom, return the original target raise
+	if targetRaise.Denom == liquidityDenom {
+		return targetRaise, nil
+	}
+
+	// convert the target raise to the base denom (just in case it's not set in base denom)
+	baseTargetRaise, err := m.tk.CalcCoinInBaseDenom(ctx, targetRaise)
```
omritoptix, line 338: if targetRaise is in DYM (which is the default currently) - shouldn't we skip this stage?

x/iro/keeper/graduation.go
```diff
@@ -0,0 +1,122 @@
...
+
+// bootstrapLiquidityPool bootstraps the liquidity pool with the raised liquidity and unsold tokens.
+//
+// This function performs the following steps:
+// - Sends the raised liquidity to the IRO module to be used as the pool creator.
+// - Determines the required pool liquidity amounts to fulfill the last price.
+// - Creates a balancer pool with the determined tokens and liquidity.
+func (k Keeper) bootstrapLiquidityPool(ctx sdk.Context, plan types.Plan, poolTokens math.Int) (uint64, sdk.Coins, error) {
+	// claimable amount is kept in the module account and used for user's claims
+	claimableAmt := plan.SoldAmt.Sub(plan.ClaimedAmt)
```
omritoptix, line 73: can `ClaimedAmt` here be anything but 0 if we open the pool only on graduation?

x/iro/keeper/trade.go
```diff
@@ -111,6 +124,23 @@ func (k Keeper) Buy(ctx sdk.Context, planId string, buyer sdk.AccAddress, amount
 		return err
 	}
 
+	// if all tokens are sold, we need to graduate the plan
+	if plan.SoldAmt.Equal(plan.MaxAmountToSell) {
+		poolID, _, err := k.GraduatePlan(ctx, planId)
```
omritoptix, line 129: we don't do anyting with leftover tokens in the case of custom pool? (in case of V2 pool shouldn't be leftovers right? )

proto/dymensionxyz/dymension/iro/iro.proto
```diff
@@ -62,8 +62,9 @@ message Plan {
       [ (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
 
   // The time before which the rollapp cannot be started.
-  google.protobuf.Timestamp pre_launch_time = 8
-      [ (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
+  // Deprecated: use iro_plan_duration instead.
+  google.protobuf.Timestamp deprecated_pre_launch_time = 8
```
omritoptix: I see you used depcreated = true. why change the var name to deprecated_pre_launch_time? didn't see this pattern before

x/iro/keeper/params.go
```diff
@@ -23,6 +25,47 @@ func (m msgServer) UpdateParams(goCtx context.Context, req *types.MsgUpdateParam
...
+	/* -------------------------------------------------------------------------- */
+	fairLaunch := req.NewParams.FairLaunch
+
+	// validate target raise denom is allowed
+	targetRaise := fairLaunch.TargetRaise
+	targetRaiseMetadata, ok := m.BK.GetDenomMetaData(ctx, targetRaise.Denom)
+	if !ok {
+		return nil, errorsmod.Wrapf(gerrc.ErrInvalidArgument, "denom %s not registered", targetRaise.Denom)
+	}
+	targetRaiseExponent := targetRaiseMetadata.DenomUnits[len(targetRaiseMetadata.DenomUnits)-1].Exponent
```
omritoptix, line 39: same as my previos comment about being sure it's always the last in the list which has the biggest decimal

x/iro/keeper/settle.go
```diff
@@ -47,40 +45,49 @@ func (k Keeper) Settle(ctx sdk.Context, rollappId, rollappIBCDenom string) error
...
+	// if already graduated, we need to replace the IRO asset tokens with the settled denom
+	if plan.IsGraduated() {
+		// call ReplacePoolAsset to replace the IRO asset with the settled denom
+		err := k.gk.ReplacePoolAsset(ctx, k.AK.GetModuleAddress(types.ModuleName), plan.GraduatedPoolId, plan.GetIRODenom(), rollappIBCDenom)
+		if err != nil {
+			return err
+		}
+		plan.SettledDenom = rollappIBCDenom
+	} else {
+		plan.SettledDenom = rollappIBCDenom
```
omritoptix: nit - this doesn't have to be in if else as it's always the case.

omritoptix (commented):
great work!

x/iro/keeper/create_plan.go
mtsitrin: custom rollapps **can** be launched before graduation. it's the logic we have today. it's unchanged

x/iro/keeper/create_plan.go
mtsitrin: It's a known issue. we'll see how it looks in the UI first

x/iro/keeper/create_plan.go
mtsitrin: I've used `FairLaunch` in the backend for a plan created with global params (not custom)
u can suggest a better, descriptive name (not v2)

x/iro/keeper/create_plan.go
mtsitrin: it's enforced for valid metadata
```
//   - Denomination units are sorted in ascending order
```

x/iro/keeper/create_plan.go
mtsitrin: if one created a token, but he wants to enable it manually a bit later (manual IRO launch feature)

x/iro/keeper/create_plan.go
mtsitrin: it's matter as we set
```
		k.rk.SetPreLaunchTime(ctx, &rollapp, plan.StartTime.Add(plan.IroPlanDuration))
```
so we want to enforce minimal duration from at least creation time

x/iro/keeper/create_plan.go
mtsitrin: it's skipped inside the function
anyway the default for mainnet is planned to be USDC (as set in the migration)

proto/dymensionxyz/dymension/iro/iro.proto
mtsitrin: no good reason. I'll change

x/iro/keeper/graduation.go
mtsitrin: the fee is considered already claimed.
it not related to graduation anyway, claiming is possible only after settlement

x/iro/keeper/params.go
mtsitrin: yes

x/iro/keeper/trade.go
mtsitrin: `MaxAmountToSell` is the equalibrium point, regardless of v1/v2 type
in both cases no leftovers expected

x/rollapp/keeper/rollapp.go
mtsitrin: it's hard to have in same place, as it's different triggers
all triggers call `SetPreLaunchTime` so it's easier to follow:
* custom IRO (either on creation, on trading enable, or on graduation)
* IRO v2 (on graduation only)

x/iro/keeper/settle.go
mtsitrin: if I set the `SettledDenom` outside the if-else, the plan already considerd settled when checking for `if plan.IsGraduated()`

mtsitrin: i've refactored it

x/iro/keeper/create_plan.go
mtsitrin: renamed to `standard launch`

x/iro/keeper/create_plan.go
```diff
@@ -127,13 +265,14 @@ func (m msgServer) CreatePlan(goCtx context.Context, req *types.MsgCreatePlan) (
...
+func (k Keeper) CreatePlan(ctx sdk.Context, liquidityDenom string, allocatedAmount math.Int, planDuration time.Duration, startTime time.Time, tradingEnabled bool, standardLaunch bool, rollapp rollapptypes.Rollapp, curve types.BondingCurve, incentivesParams types.IncentivePlanParams, liquidityPart math.LegacyDec, vestingDuration, vestingStartTimeAfterSettlement time.Duration) (string, error) {
 	allocation, err := k.MintAllocation(ctx, allocatedAmount, rollapp.RollappId, rollapp.GenesisInfo.NativeDenom.Display, uint64(rollapp.GenesisInfo.NativeDenom.Exponent))
 	if err != nil {
 		return "", err
 	}
 
 	plan := types.NewPlan(k.GetNextPlanIdAndIncrement(ctx), rollapp.RollappId, liquidityDenom, allocation, curve, planDuration, incentivesParams, liquidityPart, vestingDuration, vestingStartTimeAfterSettlement)
+	plan.StandardLaunch = standardLaunch
```
keruch, line 275: maybe set it in `NewPlan` for consistency?

omritoptix (approved):

x/iro/keeper/migrations.go
mtsitrin: https://github.com/dymensionxyz/dymension/issues/1999

x/iro/keeper/settle.go
```diff
@@ -47,40 +45,48 @@ func (k Keeper) Settle(ctx sdk.Context, rollappId, rollappIBCDenom string) error
...
-	// mark the plan as `settled`, allowing users to claim tokens
+	// commit the settled plan
 	plan.SettledDenom = rollappIBCDenom
 	k.SetPlan(ctx, plan)
 
-	// uses the raised liquidity and unsold tokens to bootstrap the rollapp's liquidity pool
-	poolID, gaugeID, err := k.bootstrapLiquidityPool(ctx, plan, poolTokens)
+	// burn all the remaining IRO token.
+	iroTokenBalance := k.BK.GetBalance(ctx, k.AK.GetModuleAddress(types.ModuleName), plan.TotalAllocation.Denom)
+	err = k.BK.BurnCoins(ctx, types.ModuleName, sdk.NewCoins(iroTokenBalance))
```
keruch: shouldn't call if `iroTokenBalance` is empty?

x/iro/types/params.go
```diff
@@ -18,10 +20,16 @@ var (
 	DefaultMinLiquidityPart                             = "0.4"                       // default: at least 40% goes to the liquidity pool
 	DefaultMinVestingDuration                           = 7 * 24 * time.Hour          // default: min 7 days
 	DefaultMinVestingStartTimeAfterSettlement           = 0 * time.Minute             // default: no enforced minimum by default
+
+	DefaultStandardLaunch = StandardLaunch{
+		AllocationAmount: math.NewInt(1e9).MulRaw(1e18),                                // 1B RA tokens
+		TargetRaise:      sdk.NewCoin(params.BaseDenom, math.NewInt(2e4).MulRaw(1e18)), // 20K DYM
```
keruch: nit
```suggestion
		TargetRaise:      common.DymUint64(20_000), // 20K DYM
```

in the latest main
https://github.com/dymensionxyz/dymension/blob/caf84a4/x/common/types/constants.go#L25-L28

keruch (approved):
good!

x/iro/keeper/settle.go
```diff
@@ -89,66 +95,20 @@ func (k Keeper) Settle(ctx sdk.Context, rollappId, rollappIBCDenom string) error
...
-	}
-
-	// Add incentives
-	poolDenom := gammtypes.GetPoolShareDenom(poolId)
-	incentives := sdk.NewCoins(
-		sdk.NewCoin(baseLiquidityCoin.Denom, poolTokens.Sub(baseLiquidityCoin.Amount)),
-		sdk.NewCoin(rollappLiquidityCoin.Denom, unallocatedTokens.Sub(rollappLiquidityCoin.Amount)),
-	)
+// addIncentivesToPool adds incentives to the pool
+func (k Keeper) addIncentivesToPool(ctx sdk.Context, plan types.Plan, poolID uint64, incentives sdk.Coins) (gaugeID uint64, err error) {
```
keruch: you can pass just `types.IncentivePlanParams` here instead of the whole `types.Plan`

keruch (approved):

Issue #2009, opened 2025-09-11
Add option to charge bridging fee for Kaspa hyperlane bridging


omritoptix:
similar to the bridging fee we have now only we will not burn this fee but save it in the `x/bridgingfee` module.
the fee should be a param.

PR #2014, opened 2025-09-16
feat(bridgingfee): fee for HL transfers

keruch:
Closes https://github.com/dymensionxyz/dymension/issues/2009


x/bridgingfee/client/cli/tx.go
```diff
@@ -0,0 +1,207 @@
+package cli
```
danwt, line 1: why manual cli impl
autogenerated not good enough/

proto/dymensionxyz/dymension/bridgingfee/types.proto
```diff
@@ -0,0 +1,55 @@
...
+
+import "gogoproto/gogo.proto";
+import "cosmos_proto/cosmos.proto";
+
+option go_package = "github.com/dymensionxyz/dymension/v3/x/bridgingfee/types";
+
+message HLAssetFee {
+  // Hyperlane token ID
+  // https://docs.hyperlane.xyz/docs/alt-vm-implementations/cosmos-sdk#x%2Fwarp
+  string tokenID = 1;
```
mtsitrin: doesn't it need to be 
```
    (gogoproto.customtype) =
        "github.com/bcp-innovations/hyperlane-cosmos/util.HexAddress",
```

x/bridgingfee/keeper/keeper.go
```diff
@@ -0,0 +1,269 @@
...
+	return k
+}
+
+func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", "x/"+types.ModuleName)
+}
+
+// CreateFeeHook creates a new fee hook (business logic)
+func (k Keeper) CreateFeeHook(ctx context.Context, msg *types.MsgCreateBridgingFeeHook) (hyputil.HexAddress, error) {
+	err := msg.ValidateBasic()
```
mtsitrin, line 80: we should abonden the validateBasic when possible, and have the validation logic in a single flow in the handler

proto/dymensionxyz/dymension/bridgingfee/query.proto
```diff
@@ -0,0 +1,116 @@
...
+  repeated AggregationHook aggregation_hooks = 1
+      [ (gogoproto.nullable) = false ];
+  cosmos.base.query.v1beta1.PageResponse pagination = 2;
+}
+
+// QueryQuoteFeePaymentRequest is the request type for the Query/QuoteFeePayment RPC method.
+message QueryQuoteFeePaymentRequest {
+  string hook_id = 1;
+  string token_id = 2;
+  string transfer_amount = 3;
```
mtsitrin, line 107: math.Int?

x/bridgingfee/keeper/hook_fee.go
```diff
@@ -0,0 +1,139 @@
...
+func (f FeeHookHandler) PostDispatch(goCtx context.Context, _, hookId hyputil.HexAddress, metadata hyputil.StandardHookMetadata, message hyputil.HyperlaneMessage, maxFee sdk.Coins) (sdk.Coins, error) {
+	// Parse warp payload to get transfer amount
+	payload, err := warptypes.ParseWarpPayload(message.Body)
+	if err != nil {
+		return nil, fmt.Errorf("parse warp payload: %w", err)
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	fee, err := f.QuoteFeeInBase(ctx, hookId, message.Sender, math.NewIntFromBigIntMut(payload.Amount()))
```
mtsitrin: why `NewIntFromBigIntMut` and not `NewIntFromBigInt`?

x/bridgingfee/keeper/hook_fee.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if fee.IsZero() {
+		// Nothing to charge
+		return sdk.NewCoins(), nil
+	}
+
+	if !maxFee.IsAllGTE(fee) {
+		return sdk.NewCoins(), fmt.Errorf("required fee payment exceeds max fee: %v", maxFee)
```
mtsitrin:
```suggestion
		return nil, fmt.Errorf("required fee payment exceeds max fee: %v", maxFee)
```

x/bridgingfee/keeper/hook_fee.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// Accumulate fees on the x/bridgingfee account
+	err = f.k.bankKeeper.SendCoinsFromAccountToModule(ctx, metadata.Address, types.ModuleName, fee)
```
mtsitrin, line 65: why we accumalte it and not send to txfees to burn? (using `ChargeFees` interface, as in other places we charge protocol fee)

x/bridgingfee/keeper/keeper.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return hyputil.HexAddress{}, fmt.Errorf("invalid msg: %w", err)
+	}
+
+	// Verify all tokenIDs exist
+	for _, fee := range msg.Fees {
```
mtsitrin, line 86: we should validate `x/txfees`'s `FeeToken` exists for this denom

x/bridgingfee/types/types.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	return nil
+}
+
+// Validate validates the asset fee
+func (f HLAssetFee) Validate() error {
+	if f.TokenID == "" {
+		return fmt.Errorf("token id cannot be empty")
+	}
+	if f.InboundFee.IsNegative() {
```
mtsitrin, line 37: must be `< 1.0`

x/bridgingfee/keeper/hook_fee.go
```diff
@@ -0,0 +1,139 @@
...
+	fee := assetFee.OutboundFee.MulInt(transferAmt).TruncateInt()
+
+	feeCoin := sdk.NewCoin(tokenResp.Token.OriginDenom, fee)
+
+	feeBase, err := f.k.txFeesKeeper.CalcCoinInBaseDenom(ctx, feeCoin)
+	if err != nil {
+		return nil, fmt.Errorf("calc fee in base denom: %w", err)
+	}
+
+	return sdk.NewCoins(feeBase), nil
```
mtsitrin: fee is always charged in DYM? why not in the feeCoin?
otherwise if I bridge alot of kaspa out, i'll need to have enough DYM balance to pay fees

x/bridgingfee/keeper/hook_aggregation.go
```diff
@@ -0,0 +1,102 @@
...
+			if err != nil {
+				return fmt.Errorf("get post-dispatch module for %s: %w", subHookId.String(), err)
+			}
+			chargedFee, err := (*pdModule).PostDispatch(ctx, mailboxId, subHookId, metadata, message, remaining)
+			if err != nil {
+				return fmt.Errorf("execute sub-hook %s: %w", subHookId.String(), err)
+			}
+
+			totalCharged = totalCharged.Add(chargedFee...)
+			remaining = remaining.Sub(chargedFee...)
```
mtsitrin: can use `SafeSub` as in the hyperlane repo

mtsitrin (changes requested):

x/bridgingfee/keeper/keeper.go
keruch: It's already called within a single flow when creating a hook, just in a separate method for clarity. There is `CreateFeeHook` in `Keeper` and in `msgServer`, this validation is in `Keeper`, so no way the caller could miss it

proto/dymensionxyz/dymension/bridgingfee/query.proto
keruch: in HL, math fields in queries also use `string`, so I keep it the same way https://github.com/dymensionxyz/hyperlane-cosmos/blob/b71bd254db75e49abfc3891a125b6184c39f9dab/proto/hyperlane/core/post_dispatch/v1/query.proto#L104

changed to `math.int`

x/bridgingfee/keeper/hook_fee.go
keruch: `payload` is a new object created on every `PostDispatch` call. it's not used after `QuoteFeeInBase`, so we can safely reuse underlying bytes instead of copying.

x/bridgingfee/keeper/hook_fee.go
keruch: we discussed with Omri that we want to keep bridging fees on the module account and maybe use them later

proto/dymensionxyz/dymension/bridgingfee/query.proto
keruch: math.Int can't be used as an argument in the query server request, the autogenerated code doesn't compile

x/bridgingfee/keeper/hook_fee.go
keruch: discussed here https://dymensionworkspace.slack.com/archives/C090ZM6MT7W/p1758213922483189

TLDR
- relayer gas and fee is 0 for now
- bridging fee is in KAS
- maxFee is in KAS

x/bridgingfee/types/types.go
keruch: theoretically the fee might be more than 100% so why limit it?

x/bridgingfee/keeper/keeper.go
keruch: I removed all `x/txfees` interactions, see https://github.com/dymensionxyz/dymension/pull/2014#discussion_r2367501306

x/bridgingfee/client/cli/tx.go
keruch: I honestly tried to make it work but no luck.. I was stuck with (un)marshaling fields with custom repeated types (and specifically with ones containing hyperlane's HexAddress).

danwt: damn
btw you can combine them too though, like use autogenerated and then add or override specific custom ones

x/bridgingfee/client/cli/tx.go
```diff
@@ -0,0 +1,364 @@
...
+				return err
+			}
+
+			var fees []types.HLAssetFee
+			for _, feeJSON := range args {
+				var feeInput struct {
+					TokenID     string `json:"token_id"`
+					InboundFee  string `json:"inbound_fee"`
+					OutboundFee string `json:"outbound_fee"`
+				}
```
danwt: its very awkward, wouldn't it be better if it was one json list? not multiple args/

x/bridgingfee/client/cli/tx.go
```diff
@@ -0,0 +1,364 @@
...
+			}
+
+			msg := &types.MsgCreateBridgingFeeHook{
+				Owner: clientCtx.GetFromAddress().String(),
+				Fees:  fees,
+			}
+
+			if err := msg.ValidateBasic(); err != nil {
+				return err
+			}
```
danwt: IIRC this isn't required, it will called by client regardless

x/bridgingfee/client/cli/tx.go
```diff
@@ -0,0 +1,364 @@
...
+
+			renounceOwnership, err := cmd.Flags().GetBool(FlagRenounceOwnership)
+			if err != nil {
+				return err
+			}
+
+			var fees []types.HLAssetFee
+			for _, feeJSON := range feesJSON {
```
danwt: ditto previous comment

x/bridgingfee/client/cli/tx.go
danwt: not very dry this is repeated in update

x/bridgingfee/client/cli/tx.go
```diff
@@ -0,0 +1,364 @@
...
+
+Examples:
+# Update hook IDs
+dymd tx bridgingfee set-aggregation-hook 0x1234... --hook-ids 0x1234...,0x5678... --from mykey
+
+# Transfer ownership
+dymd tx bridgingfee set-aggregation-hook 0x1234... --new-owner dym1newowner... --from mykey
+
+# Renounce ownership
+dymd tx bridgingfee set-aggregation-hook 0x1234... --renounce-ownership --from mykey`,
```
danwt: dont need 'from' in the examples

x/bridgingfee/client/cli/tx.go
keruch: yeah, that's exactly what I did!

x/bridgingfee/keeper/keeper_test.go
```diff
@@ -0,0 +1,66 @@
...
+	msgServer := keeper.NewMsgServerImpl(app.BridgingFeeKeeper)
+
+	s.App = app
+	s.Ctx = ctx
+	s.queryClient = queryClient
+	s.msgServer = msgServer
+}
+
+// CreateTokenWithPool creates a token in the warp module and a corresponding pool for fee calculations
+func (s *KeeperTestSuite) CreateTokenWithPool(originDenom string, amount math.Int) string {
```
danwt: why does this return a value? also where is it called?

x/bridgingfee/keeper/keeper.go
```diff
@@ -0,0 +1,266 @@
...
+		Owner:  hook.Owner,
+	})
+	if err != nil {
+		return fmt.Errorf("emit event: %w", err)
+	}
+
+	return nil
+}
+
+// CreateAggregationHook creates a new aggregation hook (business logic)
```
danwt: Could you please hugely elaborate here on this, i.e. what is this hook type, how does it work, etc.

x/bridgingfee/keeper/keeper.go
```diff
@@ -0,0 +1,266 @@
...
+
+	return k
+}
+
+func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("module", "x/"+types.ModuleName)
+}
+
+// CreateFeeHook creates a new fee hook (business logic)
+func (k Keeper) CreateFeeHook(ctx context.Context, msg *types.MsgCreateBridgingFeeHook) (hyputil.HexAddress, error) {
```
danwt, line 79: Could you please hugely elaborate here on this, i.e. what is this hook type, how does it work, etc.

x/bridgingfee/keeper/keeper.go
```diff
@@ -0,0 +1,266 @@
...
+
+	schema, err := sb.Build()
+	if err != nil {
+		panic(err)
+	}
+	k.schema = schema
+
+	// Register HL post-dispatch hooks in x/core
+	router := coreKeeper.PostDispatchRouter()
+	router.RegisterModule(postdispatchtypes.POST_DISPATCH_HOOK_TYPE_PROTOCOL_FEE, NewFeeHookHandler(k))
```
danwt: do you think its better to define and use our own enum items, because otherwise upstream might develop a different implementation which uses these and we would find it hard to merge

danwt: it would also avoid confusion

x/bridgingfee/keeper/keeper_test.go
keruch: it was used in a previous impl where we had pools, removed

omritoptix (approved):

PR #2040, opened 2025-10-02
fix: filepath in err string when using Wrapf causes non-determinism

keruch:
Cherry pick of ef88f236a76c317b2f567295d248e863b118b4e0


omritoptix (approved):

danwt:
I'm curious where is the actual place where the error is written to the chain state?

keruch:
> I'm curious where is the actual place where the error is written to the chain state?

@danwt here it is https://github.com/cosmos/cosmos-sdk/blob/v0.50.14/x/gov/abci.go#L198

and that was our problematic line https://github.com/dymensionxyz/dymension/blob/6dc69057ebef9b9e1fbf2360180457e901918019/x/iro/keeper/params.go#L67

PR #2043, opened 2025-10-03
fix(ci): free up disk space task

artemijspavlovs:
the last binary release for latest rc failed due to the hosted workers not having enough space to download go dependencies. 
https://github.com/dymensionxyz/dymension/actions/runs/18222417075/job/51887515528

this PR adds a script which removes unused/unnecessary data from the worker file system before starting the release process.

### DIFF
--- a/.github/workflows/release_binary.yml
+++ b/.github/workflows/release_binary.yml
@@ -1,34 +1,28 @@
----
 name: Release Binary
-
 on:
   release:
     types: [created]
-      
 permissions: write-all
-
 # This workflow creates a release using goreleaser
 # via the 'make release' command.
-
 jobs:
   release:
     runs-on: ubuntu-latest
     environment: release
     steps:
+      - name: Free disk space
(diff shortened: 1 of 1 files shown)


keruch (approved):

.github/workflows/release_binary.yml
```diff
@@ -1,34 +1,28 @@
...
 # This workflow creates a release using goreleaser
 # via the 'make release' command.
-
 jobs:
   release:
     runs-on: ubuntu-latest
     environment: release
     steps:
+      - name: Free disk space
+        run: curl -fsSL https://raw.githubusercontent.com/kou/arrow/e49d8ae15583ceff03237571569099a6ad62be32/ci/scripts/util_free_space.sh | bash
```
mtsitrin: why forked version and not the vanilla `apache/arrow`?

artemijspavlovs: no reason, this is just the first thing I found that did everything I needed.

they are identical though, I'll switch to the vanilla

PR #2046, opened 2025-10-09
fix(IRO): fixed precision for accurate liquidity calculations

mtsitrin:
- Updated the liquidity pool token calculation to use SpotPriceWithPrecision for improved accuracy.
- Added SpotPriceWithPrecision method to BondingCurve and Plan types.
- Adjusted tests to validate the new precision handling in liquidity calculations.

### DIFF
--- a/x/iro/types/bonding_curve.go
+++ b/x/iro/types/bonding_curve.go
@@ -117,7 +117,7 @@ func (lbc BondingCurve) SpotPrice(x math.Int) math.LegacyDec {
 func (lbc BondingCurve) SpotPriceWithPrecision(x math.Int) math.LegacyDec {
 	precDiff := lbc.RollappDenomDecimals - lbc.LiquidityDenomDecimals
 	if precDiff > 0 {
-		return lbc.SpotPrice(x).QuoInt(math.NewIntWithDecimal(1, int(precDiff)))
+		return lbc.SpotPrice(x).QuoInt(math.NewIntWithDecimal(1, int(precDiff))) // #nosec G115
 	}
 	return lbc.SpotPrice(x)
 }


x/iro/types/bonding_curve.go
```diff
@@ -111,6 +111,17 @@ func (lbc BondingCurve) SpotPrice(x math.Int) math.LegacyDec {
...
 
+// SpotPriceWithPrecision returns the spot price at x, with precision factor for the liquidity denom
+// - x: the current supply, in the base denomination
+// - returns: the spot price at x, as price per base token, considering the precision difference between the rollapp and liquidity denoms
+func (lbc BondingCurve) SpotPriceWithPrecision(x math.Int) math.LegacyDec {
+	precDiff := lbc.RollappDenomDecimals - lbc.LiquidityDenomDecimals
+	if precDiff > 0 {
+		return lbc.SpotPrice(x).QuoInt(math.NewIntWithDecimal(1, int(precDiff))) // #nosec G115
+	}
+	return lbc.SpotPrice(x)
```
danwt, line 122: I dont get the justification for this
what about negative precDiff?

x/iro/types/plan.go
```diff
@@ -110,6 +110,12 @@ func (p Plan) SpotPrice() math.LegacyDec {
 	return p.BondingCurve.SpotPrice(p.SoldAmt)
 }
 
+// SpotPriceWithPrecision returns the spot price of the plan,
+// considering the precision difference between the rollapp and liquidity denoms
+func (p Plan) SpotPriceWithPrecision() math.LegacyDec {
+	return p.BondingCurve.SpotPriceWithPrecision(p.SoldAmt)
+}
```
danwt, line 117: Why do we keep SpotPrice? From reading the code it seems like SpotPriceWithPrecision is the one that works as expected and is correct and the other one is wrong so why have it?

x/iro/types/bonding_curve.go
mtsitrin: it won't happen. rollapp decimals are assumed and validated to be 18

x/iro/types/plan.go
mtsitrin: it's already set in the events. I didn't want to change the events to avoid messing with the indexer

x/iro/types/bonding_curve.go
danwt: better to panic then otherwise its just confusing
or at least cmment

x/iro/types/plan.go
danwt: I didn't get you what does it have to do with events?

mtsitrin: > Why do we keep SpotPrice? 

because it's already used in the events emitted on trades, and the indexer has already integrated it.
Otherwise, I would have removed it

PR #2055, opened 2025-10-14
fix(TEE): remove a check (moved from on-chain to offchain in dymint)

danwt:


keruch (commented):
we can also remove finalized height from proto https://github.com/dymensionxyz/dymension/blob/f9592ea7484dc2847b911988b1c3898e40a98978/proto/dymensionxyz/dymension/rollapp/tx.proto#L237-L238

PR #2088, opened 2025-11-04
feat(IRO): adding initial FDV for standard (v2) IRO

mtsitrin:
* added `initial_fdv` for IRO standard launch params
* support C > 0 in bonding curves
* calculate graduation point and M with support of C > 0

updated ADR: https://www.notion.so/dymension/ADR-x-IRO-533f1c2f5a2345838150dd8e954c99c6?source=copy_link#2a0a4a51f86a80baa9f4eaa4dda9a4a4

### DIFF
--- a/x/iro/types/liquidity.go
+++ b/x/iro/types/liquidity.go
@@ -135,7 +135,7 @@ func MOfX(
 	xBig := osmomath.BigDecFromSDKDec(x)
 	nBig := osmomath.BigDecFromSDKDec(N)
 
-	var xPowN osmomath.BigDec = xBig.Power(nBig) // Calculate x^N
+	xPowN := xBig.Power(nBig) // Calculate x^N
 
 	// inner = x*(N+2)/(N+1) - T
 	frac := nPlusTwo.Quo(nPlusOne) // (N+2)/(N+1)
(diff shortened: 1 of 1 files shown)


x/iro/types/liquidity.go
```diff
@@ -64,3 +74,188 @@ func FindEquilibrium(curve BondingCurve, totalAllocation math.Int, r math.Legacy
...
+	if C.IsZero() {
+		return math.LegacyZeroDec(), errors.New("curve C is zero")
+	}
+	if N.IsNegative() {
+		return math.LegacyZeroDec(), errors.New("curve N is negative")
+	}
+	// Preconditions we assume are already validated externally:
+	// - T > 0
+	// - C > 0
+	// - N >= 1 and integral (N == floor(N))
```
danwt, line 192: if N>=1 should have this in the code instead of the negative check
also if !positive for c and t

x/iro/types/liquidity.go
```diff
@@ -64,3 +74,188 @@ func FindEquilibrium(curve BondingCurve, totalAllocation math.Int, r math.Legacy
...
+	// We'll clamp to ints and keep it closed.
+
+	// low = floor(T/2)
+	lowLimit := T.QuoInt64(2)
+
+	// high = floor( T * (N+1)/(N+2) ) - 1
+	nPlusOne := N.Add(math.LegacyOneDec())        // N+1
+	nPlusTwo := nPlusOne.Add(math.LegacyOneDec()) // N+2
+	ratio := nPlusOne.Quo(nPlusTwo)               // (N+1)/(N+2)
+	highLimit := T.Mul(ratio)
```
danwt, line 210: where is the minus one from the comment?

x/iro/types/liquidity.go
mtsitrin: wrong comment. N > 0 is fine. i'll fix

x/iro/types/liquidity.go
mtsitrin: AI wanted to emphasize the limit is not inclusive. it dosen't really matter for the binary search
i'll update the comment as well

PR #2131, opened 2025-12-18
feat(forward/cli): add Kaspa recipient address display in decode-hl

danwt:
## Summary

Extends the `decode-hl` CLI command to display Kaspa recipient addresses derived from the 32-byte H256 payload. Shows both mainnet and testnet address variants so users can identify the correct one for their network.

## Example

```
dymd q forward decode-hl 0x03000002425d990b31...
```

```
=== Hyperlane Message ===
  Version:     3
  Nonce:       578
  Origin:      1570310961
  Sender:      0x726f757465725f61707000000000000000000000000000020000000000000000
  Destination: 1082673309
  Recipient:   0x0000000000000000000000000000000000000000000000000000000000000000
  Body:        0x8f03e108e69d36805a47dfd96353cb696917419cd0c62db239142f7f89c9fb1e...

=== Forwarding Memo ===
  (none)

=== Warp Payload ===
  Kaspa (mainnet): kaspa:qz8s8cggu6wndqz6gl0ajc6ned5kj96pnngvvtdj8y2z7lufe8a3uvxe5y0k0
  Kaspa (testnet): kaspatest:qz8s8cggu6wndqz6gl0ajc6ned5kj96pnngvvtdj8y2z7lufe8a3udql0t38t
  Cosmos Account: dym13up7zz8xn5mgqkj8mlvkx57td953wsvu6rrzmv3ezshhlzwflv0ql2swa2
  Amount:         4000000001
```

## Implementation

- New `kaspaaddr` package with `FromH256()` to convert 32-byte H256 to Kaspa bech32 address
- Uses `btcd/btcutil/bech32.ConvertBits` for bit conversion (already in go.mod)
- Custom checksum functions required because Kaspa uses non-standard bech32:
  - Separator `:` instead of `1`
  - 40-bit checksum (8 chars) instead of standard 30-bit (6 chars)
  - Different generator polynomial

## Test plan

- [x] Unit tests verified against Rust kaspa-tools output
- [x] Build passes
- [x] Manual test with decode-hl command

### DIFF
--- a/x/forward/cli/decode.go
+++ b/x/forward/cli/decode.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/bcp-innovations/hyperlane-cosmos/util"
 	warptypes "github.com/bcp-innovations/hyperlane-cosmos/x/warp/types"
+	"github.com/dymensionxyz/dymension/v3/x/forward/cli/kaspaaddr"
 	"github.com/dymensionxyz/dymension/v3/x/forward/types"
 )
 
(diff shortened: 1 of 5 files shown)


x/forward/cli/decode.go
```diff
@@ -242,6 +243,15 @@ func printWarpPayload(warpPL warptypes.WarpPayload) {
 	if len(recipient) == 32 && bytes.Equal(recipient[:12], make([]byte, 12)) {
 		fmt.Printf("  EVM Address:    0x%s\n", hex.EncodeToString(recipient[12:]))
 	}
+
+	// Kaspa addresses: convert 32-byte recipient to kaspa bech32m address (both networks)
```
mtsitrin, line 247: don't we need switch case over the `DestinationDomain`?

PR #2168, opened 2026-06-24
feat(eibc): time-based competitive fee escalation for demand orders

### LINKED ISSUE #2167: x/eibc: time-based competitive fee escalation (Dutch auction) for demand orders
## Problem / opportunity

eIBC is Dymension's intent-settlement layer: a recipient/originator expresses a desired outcome (fast funds) and an off-chain market maker ("solver") fulfills it before finalization. But the **fee is static**. For `ON_RECV` orders it is fixed by the originator in the IBC memo (`x/eibc/keeper/handler.go:70-78`, `EIBCMemo.Fee`); the recipient's only lever afterwards is a unilateral, manual `MsgUpdateDemandOrder` (`x/eibc/keeper/msg_server.go:187-248`). There is **no automated price discovery**.

Consequence: the originator must guess the clearing fee up front. Guess too low and the order is never filled (it sits until slow finalization — defeating the whole point of eIBC); guess too high and the recipient overpays every time. The on-demand LP system already encodes solver willingness as a threshold (`OnDemandLP.minFee`, matched in `OnDemandLPRecord.Accepts` via `GetFeePercent`, `x/eibc/types/lp.go`), but a too-cheap order simply never crosses any LP's threshold and stalls forever.

This issue adds **time-based competitive fee escalation**: the originator declares a fee *range* and a window, and the offered fee rises deterministically with block height until a solver/LP finds it profitable and fills. The recipient pays the market-clearing fee discovered by time, not a guess.

## Why now

Intent-based settlement with competitive solvers is the dominant 2026 cross-chain pattern — Across, CoW Protocol, UniswapX, and Aori all clear user intents via solver competition rather than fixed quotes, reporting ~8s median fills and ~60% lower cost than fixed-fee bridges. Time-decay/escalation auctions (Dutch auctions) are the canonical primitive these systems use for price discovery without an off-chain orderbook. eIBC is already 90% of the way there (intents = demand orders, solvers = fulfillers/LPs); it is missing exactly the competitive price-discovery step.

Sources:
- Intents & solvers, 2026 landscape (Across / CoW / UniswapX / Aori): https://eco.com/support/en/articles/11855244-what-are-intents-and-solvers-2026-guide
- Aori intent settlement (signed orders, solver fills): https://eco.com/support/en/articles/12783022-what-is-aori-intent-settlement-2026
- Intent vs bridge vs aggregator architecture comparison: https://eco.com/support/en/articles/15264370-intent-vs-bridge-vs-aggregator-2026-cross-chain-architecture-compared

## Key mechanism insight (grounded in the fulfill path)

`fulfill()` (`x/eibc/keeper/fulfill.go:36-61`) moves **only `o.Price`** from fulfiller → recipient and marks the order fulfilled; the fee is **implicit** — the fulfiller later receives the full transferred `amount` at finalization, so `fee = amount − price` is realized then. Therefore escalating the fee is equivalent to **lowering `o.Price` over time**: the fulfiller sends less to the recipient now and keeps more at finalization. The fund-movement code needs no change beyond computing the effective price; everything else is bookkeeping.

`amount` and `bridgingFee` are constant for the order, and the stored `Price` already equals `amount − baseFee − bridgingFee`. So both effective values derive from the base ones with no need to persist `amount`:
```
effectiveFee(H)   = baseFee + (maxFee − baseFee) * min(H − creationHeight, D) / D
effectivePrice(H) = basePrice − (effectiveFee(H) − baseFee)
```
All `math.Int`, deterministic, monotonic. `D` = `duration_blocks`.

### 1. Proto — `proto/dymensionxyz/dymension/eibc/demand_order.proto`
Add an optional escalation spec to `DemandOrder` (nil ⇒ static fee = today's behavior, field number after `completion_hook=13`):
```proto
message FeeEscalation {
  string max_fee_amount = 1 [(gogoproto.customtype)="cosmossdk.io/math.Int", (gogoproto.nullable)=false];
  uint64 duration_blocks = 2; // > 0
}
// in DemandOrder:
FeeEscalation fee_escalation = 14;
```
The existing `fee`/`price` hold the **base** (starting) values. Run `make proto-gen`.

### 2. Memo — `x/delayedack/types/packet_metadata.go` (`EIBCMemo`)
Add two optional JSON fields (absent ⇒ static, fully backward-compatible):
```go
FeeMax              string `json:"fee_max,omitempty"`
FeeEscalationBlocks uint64 `json:"fee_escalation_blocks,omitempty"`
```
Extend `EIBCMemo.ValidateBasic`: if either escalation field is set, **both** must be set, `FeeMax` must parse to a non-negative `math.Int` with `FeeMax >= Fee`, and `FeeEscalationBlocks > 0`; else error. Add a helper `FeeMaxInt()` mirroring `FeeInt()`.

### 3. Order creation — `x/eibc/keeper/handler.go` `CreateDemandOrderOnRecv`
After computing `demandOrderPrice` with the base `fee`, if the memo carries escalation: build a `FeeEscalation`. **Clamp** `maxFee` so the price floor stays ≥ 0: `maxFee = min(memo.FeeMax, amount − bridgingFee)` (i.e. `maxFee − baseFee <= basePrice`). Pass it into `NewDemandOrder` (extend the constructor with a trailing `*FeeEscalation` arg; pass `nil` from `CreateDemandOrderOnErrAckOrTimeout`). **`ON_TIMEOUT`/`ON_ACK` orders never escalate** (their fee is a protocol multiplier, not an originator choice).

### 4. Effective-value helpers — `x/eibc/types/demand_order.go`
```go
func (m *DemandOrder) EffectiveFeeAmount(height uint64) math.Int   // == GetFeeAmount() when FeeEscalation==nil
func (m *DemandOrder) EffectivePriceAmount(height uint64) math.Int // == PriceAmount() when FeeEscalation==nil
func (m *DemandOrder) EffectiveFeePercent(height uint64) math.LegacyDec
```
Guard: if `FeeEscalation == nil || DurationBlocks == 0` or `height <= CreationHeight`, return the base value (no panic, no div-by-zero). Cap `elapsed` at `DurationBlocks`.

### 5. Apply effective values before fund movement
Add a keeper helper that, given an order and `ctx.BlockHeight()`, returns the in-memory order with `Fee`/`Price` set to the effective values. Call it at the top of each fulfillment path so `fulfill()`'s `SendCoins(..., o.Price)` and the persisted record use the height-correct amounts:
- `FulfillOrder` (`msg_server.go:48-76`)
- `FulfillOrderAuthorized` (`msg_server.go:78-135`)
- `FulfillByOnDemandLP` (`lps.go:198-238`) — also use the effective price for `lp.Spent += o.PriceAmount()`.

### 6. Guard-check semantics (the only behavioral subtlety)
Keep today's **exact-equality** checks for non-escalating orders (so all existing tests/flows are unchanged). For escalating orders, the target moves per block, so use **bounds** that protect each party:
- `FulfillOrder.expected_fee` → treat as the fulfiller's **minimum acceptable fee**: require `EffectiveFeeAmount(H) >= expectedFee` (replaces the exact `orderFee.Equal(expectedFee)` at `msg_server.go:64-67`). Fulfiller receives the current `effectiveFee` (more is always better for them), so a lower bound is the only guard needed; this also tolerates the order escalating one block between simulate and execution.
- `FulfillOrderAuthorized` (`validateOrder`, `msg_server.go:142-151`) → `EffectivePriceAmount(H) <= msg.Price[0].Amount` (LP pays at most its quoted price) **and** `EffectiveFeeAmount(H) >= msg.ExpectedFee` (LP earns at least its quoted fee). The operator-fee split (`msg_server.go:112-122`) is computed from the effective fee.
- `FulfillByOnDemandLP` → `OnDemandLPRecord.Accepts` (`x/eibc/types/lp.go`) must use `o.EffectiveFeePercent(h)` and `o.EffectivePriceAmount(h)` (the call site `GetOrderCompatibleLPs` already has height `h`, `lps.go:190-191`). This is the payoff: a too-cheap order auto-becomes fillable once escalation lifts its fee% past an LP's `minFee`.

### 7. `MsgUpdateDemandOrder`
When the recipient manually updates the fee, **clear `FeeEscalation`** (revert to static) so the manual fee is authoritative and not re-overridden by escalation.

### 8. Events
Add an `effective_fee` attribute to the fulfilled events (`GetFulfilledEvent` / `GetFulfilledAuthorizedEvent`) so indexers see the actual cleared fee. Optionally include the escalation spec in the created event.

## Decisions (made — do not re-ask)
- **Curve:** linear integer interpolation, per-block granularity. Not exponential (simpler, deterministic, sufficient).
- **Direction:** fee only rises (`maxFee >= baseFee`); equivalently price only falls. This is the Dutch auction from the recipient's cost perspective.
- **Opt-in:** absent memo fields ⇒ byte-for-byte today's behavior. Only `ON_RECV` orders can escalate.
- **Price floor:** clamp `maxFee` to `amount − bridgingFee` so the recipient never receives a negative/zero-underflow price.
- **Guard semantics:** exact-equality preserved for static orders; bounds for escalating orders (per §6).
- **Naming:** proto `FeeEscalation { max_fee_amount, duration_blocks }`, field `fee_escalation`; memo keys `fee_max`, `fee_escalation_blocks`.
- **No new params, no governance knobs, no EndBlocker work** — everything is lazy and per-order.

## Consensus-safety
Every effective value is a pure deterministic `math.Int` function of the order's stored fields and `ctx.BlockHeight()` — no wall-clock, no randomness (the existing LP shuffle is unchanged and seeded by caller input), no map iteration, no new Begin/EndBlocker logic. Division by `duration_blocks` is guarded (`> 0` validated, plus nil/zero short-circuit). Txs remain atomic. Identical on every validator.

## Acceptance criteria
- `EffectiveFeeAmount`/`EffectivePriceAmount` unit tests: at `H == creationHeight` return base; at `H >= creationHeight + duration` return `maxFee`/min price; linear in between; `FeeEscalation == nil` returns base for all `H`; no panic at `H < creationHeight` or `duration == 0`.
- A static order (no memo escalation) behaves exactly as before — all existing `x/eibc` tests pass unchanged.
- `MsgFulfillOrder` on an escalating order at height `H`: succeeds when `expected_fee <= effectiveFee(H)`, the recipient receives `effectivePrice(H)`, and fails when `expected_fee > effectiveFee(H)`.
- `MsgFulfillOrderAuthorized` on an escalating order: succeeds when `effectivePrice(H) <= msg.Price` and `effectiveFee(H) >= msg.ExpectedFee`; operator fee split computed from `effectiveFee(H)`.
- On-demand LP: an order whose **base** fee% is below an LP's `minFee` is rejected by `Accepts` at creation height but accepted (and fulfilled by `FulfillByOnDemandLP`) at a later height once `EffectiveFeePercent(h)` crosses `minFee`; `lp.Spent` increases by the effective price.
- Creation clamps `maxFee` to `amount − bridgingFee` (test an over-large `fee_max`).
- `EIBCMemo.ValidateBasic` rejects: only one of the two escalation fields set; `fee_max < fee`; `fee_escalation_blocks == 0`; unparseable `fee_max`.
- `MsgUpdateDemandOrder` on an escalating order clears `FeeEscalation` and sets the static fee/price.
- `make proto-gen` leaves no uncommitted diff; `golangci-lint run` and `gofumpt -w .` clean on touched packages.

## Out of scope
- Escalation for `ON_TIMEOUT`/`ON_ACK` orders.
- Non-linear/exponential curves; sealed-bid or batch auctions.
- Partial fills.
- Changing the random LP shuffle/selection order in `FulfillByOnDemandLP`.
- Off-chain solver/relayer software (this only provides the on-chain primitive).

omritoptix:
Closes #2167

## What

Adds **time-based competitive fee escalation** (a Dutch auction) to eIBC `ON_RECV` demand orders. The originator declares a fee *range* and a window in the IBC memo; the offered fee then rises deterministically per block until a solver/LP finds it profitable and fills. The recipient pays the market-clearing fee discovered by time instead of a static guess.

The key insight (grounded in the fulfill path): `fulfill()` moves only `o.Price` to the recipient and the fee is realized at finalization as `amount − price`. So escalating the fee is equivalent to **lowering the price over time** — no change to fund-movement code beyond computing the effective value.

## How

- **Proto**: optional `FeeEscalation { max_fee_amount, duration_blocks }` on `DemandOrder` (nil ⇒ today's static behavior). `effective_fee` added to fulfilled events.
- **Memo**: optional `fee_max` / `fee_escalation_blocks` on `EIBCMemo`; `ValidateBasic` requires both-or-neither, `fee_max >= fee`, `blocks > 0`.
- **Effective values**: `EffectiveFeeAmount` / `EffectivePriceAmount` / `EffectiveFeePercent` — pure deterministic `math.Int` functions of stored fields and block height (linear interpolation, `elapsed` capped at duration, guarded against nil/zero/pre-creation). `ApplyEffectiveFee` locks them in and clears the spec before fund movement.
- **Creation**: `max_fee` clamped to `amount − bridgingFee` so the price floor stays ≥ 0. `ON_TIMEOUT`/`ON_ACK` never escalate.
- **Guards**: exact-equality preserved for static orders; escalating orders use bounds (fulfiller min fee, LP max price + min fee).
- **On-demand LP**: `Accepts` uses effective fee%/price, so a too-cheap order auto-becomes fillable once escalation crosses an LP's `minFee`; `lp.Spent` tracks the effective price.
- **Manual update** clears escalation so the recipient's fee is authoritative.

No new params, no governance knobs, no Begin/EndBlocker work — everything is lazy and per-order. Every effective value is deterministic (no wall-clock, no randomness, no map iteration), so it is consensus-safe.

## Proof of execution

Built with `go1.24.6`, `CGO_ENABLED=1`. `make proto-gen` regenerated the pb files with no extra diff; `golangci-lint run` and `gofumpt -w` are clean on `x/eibc` and `x/delayedack`.

### Unit tests — effective-value helpers (`x/eibc/types`)
```
=== RUN   TestEffectiveFeeAndPrice
    --- PASS: .../at_creation
    --- PASS: .../before_creation
    --- PASS: .../midpoint
    --- PASS: .../end_of_window
    --- PASS: .../past_window_(capped)
    --- PASS: .../nil_escalation
    --- PASS: .../zero_duration
--- PASS: TestEffectiveFeeAndPrice (0.00s)
--- PASS: TestEffectiveFeePercent (0.00s)
--- PASS: TestApplyEffectiveFee (0.00s)
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/types
```

### Memo validation (`x/delayedack/types`)
```
--- PASS: TestEIBCMemo_ValidateBasic_Escalation
    --- PASS: .../static_(no_escalation)
    --- PASS: .../valid_escalation
    --- PASS: .../only_fee_max_set
    --- PASS: .../only_blocks_set
    --- PASS: .../fee_max_<_fee
    --- PASS: .../fee_max_==_fee_ok
    --- PASS: .../unparseable_fee_max
    --- PASS: .../negative_fee_max
ok  	github.com/dymensionxyz/dymension/v3/x/delayedack/types
```

### Keeper integration tests (`x/eibc/keeper`)
```
--- PASS: TestKeeperTestSuite/TestCreateDemandOrderEscalationClamp        (clamps fee_max to amount-bridgingFee)
--- PASS: TestKeeperTestSuite/TestMsgFulfillOrderEscalation               (>= bound; recipient gets effective price; fails when expected_fee too high)
--- PASS: TestKeeperTestSuite/TestMsgFulfillOrderAuthorizedEscalation     (price/fee bounds; operator split from effective fee)
--- PASS: TestKeeperTestSuite/TestOnDemandLPEscalation                    (rejected at creation height, accepted later; lp.Spent = effective price)
--- PASS: TestKeeperTestSuite/TestMsgUpdateDemandOrderClearsEscalation    (manual update clears escalation)
--- PASS: TestKeeperTestSuite/TestFulfillOrderEvent                       (static order unchanged + effective_fee attribute)
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper
```

Full `go test ./x/eibc/... ./x/delayedack/...` passes — existing static-order tests are unchanged.


omritoptix (commented):
### Code review

Found 1 issue:

1. **Division-by-zero panic in `EffectiveFeePercent`.** When an escalating order's `fee_max` is at the clamp ceiling `amount - bridgingFee`, the effective price reaches exactly 0 at/after the escalation window end (`effectivePrice = basePrice - (maxFee - baseFee) = 0`). `EffectiveFeePercent` divides fee by price with no zero guard, so `math.LegacyDec.Quo(0)` panics.

https://github.com/dymensionxyz/dymension/blob/acdd54db4d2101d98cbeaaf2c7249224fa3b5519/x/eibc/types/demand_order.go#L140-L144

The clamp deliberately allows `maxFee == amount - bridgingFee` (the comment says the price floor "stays >= 0", and 0 is permitted):

https://github.com/dymensionxyz/dymension/blob/acdd54db4d2101d98cbeaaf2c7249224fa3b5519/x/eibc/keeper/handler.go#L89-L98

This is reachable from the permissionless `TryFulfillOnDemand` path: `FulfillByOnDemandLP` -> `GetOrderCompatibleLPs` -> `Accepts`, which calls `EffectiveFeePercent` unconditionally for every registered on-demand LP:

https://github.com/dymensionxyz/dymension/blob/acdd54db4d2101d98cbeaaf2c7249224fa3b5519/x/eibc/types/lp.go#L64-L69

Edge case (requires `fee_max` at the ceiling and the order surviving unfilled to its window end with at least one LP registered), but once reached it permanently traps on-demand fulfillment for that order. It's a tx-context panic (recovered by the SDK, not a chain halt) since `TryFulfillOnDemand` is a Msg handler, not BeginBlocker/EndBlocker. Fix: guard the zero divisor in `EffectiveFeePercent`, or clamp `feeMax` to keep effective price strictly positive. Note `TestCreateDemandOrderEscalationClamp` already constructs the exact zero-price scenario but never exercises the LP path at the window end, so the panic isn't caught.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

omritoptix:
Thanks for the review. Addressed the confirmed division-by-zero in `EffectiveFeePercent` (commit ac435eb3).

**Root cause:** the `maxFee` clamp used `amount - bridgingFee`, which drives the effective price to exactly `0` at the window end — both stranding the recipient (receives nothing) and dividing by zero in `EffectiveFeePercent` via the permissionless `TryFulfillOnDemand → Accepts` path.

**Fix (two-pronged):**
1. Tightened the clamp to `amount - bridgingFee - 1` so the effective price floor stays `>= 1` for all heights. `basePrice >= 1` is already guaranteed by `CalcPriceWithBridgingFee`, so the clamp never pushes `maxFee` below the base fee.
2. Added a defensive zero-price guard in `EffectiveFeePercent` (returns `1.0`, maximally attractive) so any order not created through the clamping path (e.g. genesis injection) cannot panic in this consensus path.

**Regression coverage:**
- `TestEffectiveFeePercentZeroPriceNoPanic` — direct zero-price guard.
- `TestCreateDemandOrderEscalationClamp` — now asserts clamp `989` and `NotPanics` on `EffectiveFeePercent` at and well past the window end.

All `x/eibc` and `x/delayedack` tests pass; `golangci-lint` and `gofumpt` clean on touched packages.

On the items deemed below threshold (notably `ValidateBasic` not re-validating `FeeEscalation` for genesis injection): I left those as-is per your triage, but note the new guard now makes the worst case of a malformed genesis order non-fatal.

omritoptix (commented):
### Code review — round 2 (delta)

**Resolved:** The round-1 blocker (division-by-zero panic in `EffectiveFeePercent` via the permissionless `TryFulfillOnDemand → Accepts` path) is fixed in ac435eb3, two-pronged and correct:

1. **Clamp tightened** to `amt - bridgingFee - 1`. Since `price = amt - baseFee - bridgingFee` and both the handler and `CalcPriceWithBridgingFee` compute `bridgingFee` identically (`bridgeFeeMul.MulInt(amt).TruncateInt()`), the floor is exactly `price_floor = amt - bridgingFee - maxFee >= 1`. `CalcPriceWithBridgingFee`'s positivity check also guarantees `maxAllowed >= baseFee`, so escalation stays non-decreasing — no downward-escalation regression.
2. **Defensive guard** in `EffectiveFeePercent` returns `1.0` on non-positive price; in `Accepts` this makes `feeOK` true, matching the "maximally attractive" comment — covers genesis-injected orders that bypass the clamp.

Regression coverage added and verified locally: `TestCreateDemandOrderEscalationClamp` now asserts the `989` clamp and `NotPanics` at and past the window end; `TestEffectiveFeePercentZeroPriceNoPanic` covers the guard directly. Both pass.

**No new issues** introduced by the fix.

Note: PR is still marked draft.

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #2159, opened 2026-06-23
x/agent registry: RegisterAgent + DeactivateAgent


omritoptix:
## Goal

Implement the agent registry: permissionless, fee-gated registration with an **immutable** policy, plus owner-only deactivation. No attestation logic yet — that's issue 4.

## Context

Builds on the `x/agent` scaffold (issue 2). Mirror the existing rollapp registration-fee collection pattern: rollapp's app registration charges `app_registration_fee` by `SendCoinsFromAccountToModule(owner → ModuleName, fee)` then `BurnCoins(ModuleName, fee)` (the fee is **burned**). Replicate exactly that for `agent_registration_fee`, using the agent module account (`agenttypes.ModuleName: nil`, from the scaffold) and a `BankKeeper` expected-interface with `SendCoinsFromAccountToModule` + `BurnCoins`. Message/codec/event wiring follows `x/kas` and the rollapp `tx_*` command patterns.

The policy embedded at registration is `x/common/tee` `Policy` (issue 1).

## Requirements

- `MsgRegisterAgent { string owner; string agent_id; tee.Policy policy }`:
  - Reject if `agent_id` already exists.
  - Charge `Params.agent_registration_fee` from `owner` via send-to-module-then-burn (matching rollapp app registration).
  - Store `Agent{ id, owner, policy, active: true, action_seq: 0 }`.
  - `ValidateBasic`: non-empty id/owner, well-formed policy (cert PEM parses, rego fields non-empty).
  - Emit a registration event.
- `MsgDeactivateAgent { string owner; string agent_id }`:
  - Only the registered `owner` may deactivate; otherwise reject.
  - Set `active: false`. Emit an event.
- **No policy-update message** — policy is immutable by design (a new version = a new registration).
- Register both messages in the module codec and msg server (proto Msg service with `cosmos.msg.v1.signer = "owner"`).

## Acceptance criteria

- Register a new agent → it is stored with `active: true`, `action_seq: 0`; the fee is deducted (burned) from the owner.
- Registering a duplicate `agent_id` is rejected.
- Deactivation by a non-owner is rejected; by the owner sets `active: false`.
- No message exists that mutates an existing agent's `policy`.
- Keeper tests cover: happy register, duplicate-id reject, fee deduction, non-owner deactivate reject, owner deactivate.

## Proof of execution expected

Output of `go test ./x/agent/...` showing the registry tests green.

## Manual verification

[...]

PR #2176, opened 2026-07-02
feat(agent): timelocked agent policy rotation (MsgUpdateAgentPolicy)

### LINKED ISSUE #2175: x/agent: timelocked agent policy rotation (MsgUpdateAgentPolicy) preserving identity + action-log continuity
## Problem / opportunity

`x/agent` is a **verifiable agent runtime**: an agent registers an immutable TEE attestation `Policy` (GCP Confidential Space root cert + rego structure/query/values, which pin the enclave image / model-hash the agent is allowed to run) and then appends enclave-attested entries to an append-only action log (`x/agent/keeper/registry.go`, `x/agent/keeper/msg_server.go`, `x/agent/keeper/action_log.go`).

The policy is **immutable for the agent's entire life**. The only lifecycle messages are `RegisterAgent` and `DeactivateAgent` — there is **no policy-update path** (confirmed: no `UpdateAgent`/`RotatePolicy` symbol anywhere in `x/agent`). This is a real operational gap:

- In agent infrastructure, enclave images and model weights are **patched frequently** (security fixes, model upgrades). Each patch changes the attested measurement, so the *old* policy no longer authorizes the *new* enclave.
- Today the **only** way to run a patched agent is `DeactivateAgent` (terminal — there is no reactivation) then `RegisterAgent` under a **brand-new `agent_id`**. That:
  - resets `action_seq` to `0`, **severing the continuous, auditable action-log chain** that is the whole point of the module (epic #2162: "the foundational primitive that other layers build on");
  - **destroys the agent's identity and accrued history** (a marketplace/authority layer built on top would see a different agent);
  - pays the burn-fee again for what is conceptually the same agent.

We need a way to rotate an agent's policy to a new enclave image **while preserving `agent_id`, `owner`, `action_seq`, and the action log**, without weakening the security guarantee that made the policy immutable in the first place.

## Why now

Model/enclave patch cadence is the central operational reality of the 2026 crypto-AI agent stack — agent wallets, x402 agentic payments, and intent/solver agents all assume the agent binary is a living, versioned artifact, not a frozen one. A verifiable-agent-runtime whose agents cannot upgrade their attested code without losing their on-chain identity is not usable for that stack.

- Autonomous Agents on Blockchains — standards, execution models, trust boundaries: https://arxiv.org/html/2601.04583v1
- Crypto-AI agents / agent wallets / on-chain economy 2026: https://www.vaasblock.com/news/crypto-ai-agents-onchain-x402-wallet-economy-2026/
- Coinbase x402 (agent-native payments; assumes versioned agent services): https://www.coinbase.com/developer-platform/discover/launches/x402

## Why immutability existed, and how we preserve it

Policy was made immutable (epic #2159) so an owner **cannot silently swap in a malicious policy after building reputation**. We preserve that property with a **timelocked, fully-logged rotation**: a proposed policy does not take effect for `policy_rotation_delay_blocks` blocks, and both the pending policy and its activation height are stored on the agent and emitted as an event. Any observer (a counterparty, a future marketplace/authority layer, an indexer) sees the change coming and can react before it activates. This keeps the "no silent swap" guarantee while allowing legitimate upgrades.

## Proposed change

Add a **new owner-only message `MsgUpdateAgentPolicy`** that schedules a timelocked policy rotation, plus **lazy, height-based activation** on the verification path. No ABCI hook is added (the module has no Begin/EndBlocker today, and none is needed).

### Proto — `proto/dymensionxyz/dymension/agent/agent.proto`
Add two fields to `Agent` (append, do not renumber existing fields):
```proto
message Agent {
  string id = 1;
  dymensionxyz.dymension.common.Policy policy = 2 [ (gogoproto.nullable) = false ];
  bool active = 3;
  uint64 action_seq = 4;
  string owner = 5;
  // pending_policy, if set, becomes the active policy at pending_policy_height.
  dymensionxyz.dymension.common.Policy pending_policy = 6 [ (gogoproto.nullable) = true ];
  int64 pending_policy_height = 7;
}
```
`pending_policy == nil` means "no pending rotation". Add `EventUpdateAgentPolicy`:
```proto
message EventUpdateAgentPolicy {
  string agent_id = 1;
  int64 activation_height = 2;
}
```

### Proto — `proto/dymensionxyz/dymension/agent/tx.proto`
```proto
// UpdateAgentPolicy schedules a timelocked rotation of the agent's attestation
// policy. Owner only. The new policy activates after policy_rotation_delay_blocks.
rpc UpdateAgentPolicy(MsgUpdateAgentPolicy) returns (MsgUpdateAgentPolicyResponse);

message MsgUpdateAgentPolicy {
  option (cosmos.msg.v1.signer) = "owner";
  string owner = 1 [ (cosmos_proto.scalar) = "cosmos.AddressString" ];
  string agent_id = 2;
  dymensionxyz.dymension.common.Policy new_policy = 3 [ (gogoproto.nullable) = false ];
}
message MsgUpdateAgentPolicyResponse { int64 activation_height = 1; }
```

### Param — `proto/dymensionxyz/dymension/agent/agent.proto` (`Params`) + `x/agent/types/params.go`
Add `uint64 policy_rotation_delay_blocks = 3;`. Default `100800` (≈ 7 days at 6s blocks). `Validate()` must reject `0` (a zero delay would make rotation instantaneous and defeat the timelock).

### Types — `x/agent/types/`
- `message_update_agent_policy.go`: `NewMsgUpdateAgentPolicy(owner, agentID, policy)` + `ValidateBasic()` mirroring `message_register_agent.go` — check `agent_id != ""`, `owner` is a bech32 address, and **reuse the existing package-level `validatePolicy(m.NewPolicy)`** (already in `message_register_agent.go`).
- `codec.go`: register `&MsgUpdateAgentPolicy{}` in both `RegisterCodec` (amino name `"agent/UpdateAgentPolicy"`) and `RegisterInterfaces`.
- Add exported helper on the generated type in a new `types/agent.go`:
```go
// EffectivePolicy returns the policy in force at the given block height,
// promoting a pending policy once its activation height is reached.
func (a Agent) EffectivePolicy(height int64) tee.Policy {
    if a.PendingPolicy != nil && height >= a.PendingPolicyHeight {
        return *a.PendingPolicy
    }
    return a.Policy
}
```

### Keeper — `x/agent/keeper/registry.go`
Add `func (k msgServer) UpdateAgentPolicy(goCtx, msg) (*MsgUpdateAgentPolicyResponse, error)`:
1. `GetAgent`; if not found → `ErrAgentNotFound`.
2. If `agent.Owner != msg.Owner` → `ErrUnauthorized`.
3. If `!agent.Active` → `gerrc.ErrFailedPrecondition` ("agent is not active") — deactivation is terminal, so rotating a dead agent is meaningless.
4. `delay, _ := k.PolicyRotationDelayBlocks(ctx)`; set `agent.PendingPolicy = &msg.NewPolicy`; `agent.PendingPolicyHeight = ctx.BlockHeight() + int64(delay)`. If a pending rotation already exists, **overwrite it** (re-proposing restarts the timelock — this is also the cancel/replace path).
5. `SetAgent`; emit `EventUpdateAgentPolicy{AgentId, ActivationHeight: agent.PendingPolicyHeight}` via `uevent.EmitTypedEvent` (mirror `DeactivateAgent`).
6. Return `activation_height`.

### Keeper — `x/agent/keeper/msg_server.go` (`SubmitAttestedAction`)
After the existing `GetAgent` + `agent.Active` check and **before** computing the nonce, promote a matured pending policy in-memory so verification uses the effective policy, and persist the promotion:
```go
if agent.PendingPolicy != nil && ctx.BlockHeight() >= agent.PendingPolicyHeight {
    agent.Policy = *agent.PendingPolicy
    agent.PendingPolicy = nil
    agent.PendingPolicyHeight = 0
}
```
Then verify against `agent.Policy` exactly as today. The existing tail already `SetAgent`s (it bumps `ActionSeq`), so a successful action persists the promotion in the same write; on verification failure the tx rolls back (no state change), which is correct — the effective policy is still recomputed on the next attempt because it is height-derived, not persistence-derived. (Equivalent framing: replace the direct `agent.Policy` read with `agent.EffectivePolicy(ctx.BlockHeight())` and clear the pending fields when they mature.)

### Query / CLI
- The gRPC agent query returns the raw `Agent`, which now includes `pending_policy` + `pending_policy_height` — this is intentionally **more** auditable than hiding the promotion (observers see current policy, pending policy, and activation height). No transform in the query.
- `x/agent/client/cli/tx.go`: add `cmd.AddCommand(CmdUpdateAgentPolicy())`; implement `CmdUpdateAgentPolicy()` in a new `tx_update_agent_policy.go` mirroring `CmdRegisterAgent`'s policy-input mechanism (same policy file/flag parsing), command `update-agent-policy <agent-id> <policy-input>`.

- **Timelock, not instant, not EndBlocker.** Lazy height-based activation on the submit path; no ABCI hook added.
- **`policy_rotation_delay_blocks` default = `100800`** (~7 days @ 6s blocks), governance-tunable param; `Validate()` rejects `0`.
- **Preserved across rotation:** `id`, `owner`, `active`, `action_seq`, and the entire action log. **Only** `policy` changes, and only after the delay.
- **Inactive agents cannot be rotated** (`ErrFailedPrecondition`); deactivation stays terminal.
- **Re-proposing overwrites the pending rotation and restarts the timelock** (doubles as cancel/replace). No separate cancel message.
- **Pending fields are stored on `Agent` and surfaced by the query** (transparency > hiding).
- **Policy validation reuses the existing `validatePolicy`** in `ValidateBasic` (root-cert PEM parses, non-empty query + structure).
- **Amino name:** `"agent/UpdateAgentPolicy"`.

## Consensus-safety
Fully deterministic: activation is a pure comparison of `ctx.BlockHeight()` against a stored `int64` height — no wall-clock, no randomness, no map iteration. No new Begin/EndBlocker (no panic surface). Errors return normally so txs stay atomic (rollback on failure). Every validator computes the identical effective policy at any given height.

## Acceptance criteria
- `make proto-gen` regenerates cleanly; chain builds and starts; existing `x/agent` tests still pass.
- `MsgUpdateAgentPolicy` is wired end-to-end: codec + interface registration, msg_server handler, and `dymd tx agent update-agent-policy` CLI command exist.
- New unit tests in `x/agent/keeper/msg_server_test.go` (reuse the existing `fakeVerifier`, which already captures `gotPolicy`):
  - Non-owner → `ErrUnauthorized`; unknown agent → `ErrAgentNotFound`; inactive agent → failed precondition; malformed policy → `ErrInvalidPolicy` (via `ValidateBasic`).
  - Happy path: sets `pending_policy` and `pending_policy_height == ctx.BlockHeight() + delay`; leaves `policy`, `action_seq`, and the log unchanged; response `activation_height` matches.
  - **Before** activation height: `SubmitAttestedAction` succeeds and `fakeVerifier.gotPolicy` equals the **old** policy; stored `policy` still old, pending still set.
  - **At/after** activation height: `SubmitAttestedAction` succeeds and `fakeVerifier.gotPolicy` equals the **new** policy; after the action, stored `agent.Policy` == new policy and `pending_policy == nil`.
  - Re-calling `UpdateAgentPolicy` overwrites the pending policy and resets the activation height.
  - `Agent.EffectivePolicy(height)` unit test across the boundary (`height < H` → old, `height >= H` → new, `nil pending` → active).
  - `Params.Validate()` rejects `policy_rotation_delay_blocks == 0`.

- On-chain **effects** of actions / agent authority, and the agent **marketplace** — reserved future layers of epic #2162.
- Reactivating a deactivated agent; changing an agent's `owner`; a standalone pending-cancel message; any EndBlocker-based promotion.
- The TEE verifier (`x/common/tee`), the rollapp finalize path, and the nonce/replay construction — unchanged.

---

omritoptix:
## Summary

Adds an owner-only `MsgUpdateAgentPolicy` that schedules a **timelocked** rotation of an agent's TEE attestation policy while preserving `id`, `owner`, `active`, `action_seq`, and the entire action log. Previously the only way to run a patched enclave image was `DeactivateAgent` + re-register under a new `agent_id`, which severed the append-only action-log chain and destroyed the agent's identity.

The immutability guarantee (epic #2159 — an owner cannot silently swap in a malicious policy) is preserved: a proposed policy does not activate for `policy_rotation_delay_blocks` blocks, and both the pending policy and its activation height are stored on the `Agent` and emitted via `EventUpdateAgentPolicy`, so any observer sees the change coming.

### How it works
- New `Agent.pending_policy` + `pending_policy_height` fields (`pending_policy == nil` means no pending rotation).
- New governance param `policy_rotation_delay_blocks` (default `100800`, ~7 days @ 6s blocks; `Validate()` rejects `0`).
- `UpdateAgentPolicy` handler: owner-only, active-only, overwrites any existing pending rotation (doubles as cancel/replace), returns the activation height.
- **Lazy, height-based activation** on `SubmitAttestedAction`: a matured pending policy is promoted in-memory before verification and persisted in the same write that bumps `action_seq`. No ABCI hook added.
- `Agent.EffectivePolicy(height)` helper and `dymd tx agent update-agent-policy` CLI command.

Fully deterministic (pure `ctx.BlockHeight()` comparison against a stored `int64`), atomic (rollback on verification failure).

Closes #2175

## Proof of execution

`make proto-gen` regenerated cleanly; `go build ./...` succeeds; `golangci-lint run ./x/agent/...` → `0 issues`.

New + existing `x/agent` tests (all pass, `-count=1`):

```
--- PASS: TestSubmitAttestedAction_Valid
--- PASS: TestSubmitAttestedAction_TamperedPayload
--- PASS: TestSubmitAttestedAction_WrongSeq
--- PASS: TestSubmitAttestedAction_ReplayAfterSuccess
--- PASS: TestSubmitAttestedAction_PayloadTooLarge
--- PASS: TestSubmitAttestedAction_InactiveAgent
--- PASS: TestSubmitAttestedAction_NonexistentAgent
--- PASS: TestUpdateAgentPolicy_NonOwner            (ErrUnauthorized)
--- PASS: TestUpdateAgentPolicy_UnknownAgent        (ErrAgentNotFound)
--- PASS: TestUpdateAgentPolicy_InactiveAgent       (ErrFailedPrecondition)
--- PASS: TestUpdateAgentPolicy_MalformedPolicy     (ErrInvalidPolicy via ValidateBasic)
--- PASS: TestUpdateAgentPolicy_HappyPath           (sets pending + height, leaves policy/seq/log)
--- PASS: TestUpdateAgentPolicy_Overwrite           (overwrite restarts timelock)
--- PASS: TestSubmitAttestedAction_PendingPolicyBeforeActivation  (gotPolicy == old)
--- PASS: TestSubmitAttestedAction_PendingPolicyAtActivation      (gotPolicy == new, pending cleared)
--- PASS: TestAgent_EffectivePolicy                 (boundary: <H old, >=H new, nil pending)
--- PASS: TestParams_Validate                       (rejects policy_rotation_delay_blocks == 0)
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types
```

CLI command wired (`dymd tx agent --help`):

```
Available Commands:
  deactivate-agent    Deactivate an agent you own
  register-agent      Register a new agent with an immutable attestation policy
  update-agent-policy Schedule a timelocked rotation of an agent's attestation policy
```


omritoptix (commented):
**Review — `feat(agent): timelocked agent policy rotation` (max effort: authz + attestation security surface)**

Approving. The timelock design is sound and the security-critical invariant (no silent policy swap) holds. Build + full `x/agent` suite pass locally, and the regression coverage for the new behavior is genuinely thorough.

### Verification performed
- `go build ./x/agent/...` → ok
- `go test ./x/agent/... -count=1` → `ok` (keeper + types), all listed tests pass
- Ran a dedicated security review over the authz + timelock surface

### Security — no findings
- **Authz correct**: owner-equality check mirrors `DeactivateAgent`, and `cosmos.msg.v1.signer = "owner"` enforces the signature. A non-owner cannot rotate.
- **No timelock bypass**: `PendingPolicyHeight = BlockHeight() + delay`, `delay` comes only from a genesis-validated param (`Validate()` rejects `0`, default 100800). The `>=` maturity check is consistent between the inline promotion in `SubmitAttestedAction` and `EffectivePolicy`. A policy proposed at H activates at exactly H+delay, never earlier.
- **No param-migration bypass**: the module is `ConsensusVersion() == 1` and is *not* in `v5` `StoreUpgrades.Added`, so it's pre-release — there is no on-chain 2-field param blob that could decode `policy_rotation_delay_blocks` as `0`. It ships together with `DefaultParams`. (Worth keeping in mind: if this module is ever added to a live chain in an upgrade, that upgrade must set the param, since `SetParams` doesn't call `Validate()` and stored-param decode would silently yield `0` → immediate activation. Not a bug today.)
- New policy is validated via the same `validatePolicy` used at registration, run in both `ValidateBasic` and the handler.

### Non-blocking cleanup
- `Agent.EffectivePolicy` (types/agent.go) and the inline promotion in `SubmitAttestedAction` (keeper/msg_server.go) encode the *same* maturity rule, but `EffectivePolicy` is only referenced by tests — production verifies via the inline copy. Behaviorally equivalent today, but it's the security-critical activation predicate duplicated in two places; a future edit to one risks drift. Consider having the verification path derive the effective policy through the single helper (and/or expose it in the `Agent` query so observers don't have to recompute maturity from the raw pending fields).

Proof of execution in the PR body is concrete and matches what I re-ran. LGTM.

PR #2170, opened 2026-06-25
feat(eibc): validity window + spend-velocity cap for on-demand LPs

omritoptix:
## What

Adds two opt-in autonomous-market-maker risk controls to eIBC on-demand LPs (`x/eibc`), matching the bounded/time-boxed/rate-limited authority model that agent payment rails (x402, agent wallets) have made the baseline for machine-driven spend:

1. **Validity window** — `valid_until_height`: the LP stops being matched at heights `>= valid_until_height` (exclusive expiry). `0` = no expiry.
2. **Spend-velocity cap** — `rate_limit_amount` per `rate_limit_blocks`: caps how much an LP can spend within any one window, on top of the existing lifetime `spend_limit`.

All new fields are **additive and opt-in**: when unset (`0`/nil) behavior is byte-for-byte identical to today.

### Windowing model

Deterministic **absolute-aligned tumbling windows**: `bucket(h) = h - (h % rate_limit_blocks)`. Effective in-window spend is `window_spent` only while `bucket(h) == window_start_height`, else `0` (window rolled over). Remaining capacity = `rate_limit_amount - effectiveWindowSpent`. Pure integer math over `ctx.BlockHeight()` — no wall-clock, randomness, map iteration, or new Begin/EndBlocker work. O(1), no per-fill history.

Accepted trade-off (per spec): up to ~2× `rate_limit_amount` can be spent across an exact bucket boundary — acceptable for a coarse safety rail.

### Changes
- **proto**: `OnDemandLP` gains `valid_until_height` (8), `rate_limit_amount` (9), `rate_limit_blocks` (10); `OnDemandLPRecord` gains `window_start_height` (4), `window_spent` (5). Regenerated with `make proto-gen`.
- **types** (`lp.go`): `Validate()` enforces `(rate_limit_amount>0) == (rate_limit_blocks>0)`; `Accepts()` additionally requires `validOK` and `rateOK`; new `RateAllows` / `RecordSpend` helpers; nil-`window_spent` guarded like `spent`.
- **keeper** (`lps.go`, `msg_server.go`): `FulfillByOnDemandLP` updates the window via `RecordSpend`; `CreateOnDemandLP` rejects an already-expired `valid_until_height`. Matching naturally filters expired/over-velocity LPs via the existing `Accepts(blockHeight, …)` call.
- **CLI**: `create-demand-lp` gains `--valid-until-height`, `--rate-limit-amount`, `--rate-limit-blocks`.

## Proof of execution

[...]

PR #2180, opened 2026-07-05
feat(agent): governance-gated TEE policy revocation denylist

### LINKED ISSUE #2179: x/agent: governance-gated TEE policy revocation denylist (kill a compromised enclave measurement fleet-wide)
## Problem / opportunity

`x/agent` is the verifiable agent runtime: a registry of TEE-attested agents (`x/agent/keeper/registry.go`) whose actions are appended to an attested action log only after `SubmitAttestedAction` re-derives a nonce and calls `k.verifier.Verify(ctx, agent.Policy, nonce, msg.Token)` (`x/agent/keeper/msg_server.go:57`). Each agent pins a `tee.Policy` (`x/common/tee/tee.pb.go:28`) — the GCP root cert + OPA rego + `policy_values` JSON that encodes the **expected TEE image measurement**. The measurement is what proves "this agent is running published image X".

**The gap:** there is *no way to revoke a compromised TEE image*. If a specific enclave measurement is later found vulnerable (a TDX/SGX CVE, a bad build), the chain keeps accepting attested actions from every agent pinned to that image. Today the only remedy is `DeactivateAgent`, which (a) is **owner-only** (`registry.go:55`) — an operator running the exploited image has no incentive to call it — and (b) must be done per-agent. A fleet of agents on the same bad image cannot be stopped. For a runtime whose entire trust model is "the measurement in the policy is sound", a missing global measurement-revocation path is a real security hole.

## Why now

TEE-backed autonomous agents are a live 2026 pattern, and the security literature is explicit that **on-chain accountability / revocation** — not off-chain trust — is what makes attested actors safe: bonded, on-chain-slashable actors are "strictly safer than whitelisted actors with off-chain reputation" ([Eco: Solver Networks](https://eco.com/support/en/articles/14745948-solver-networks-how-competition-replaces-routing), [Cantina: Intent-Bridge Security Failures](https://cantina.xyz/blog/intent-bridge-security-failures)). The equivalent primitive for a TEE agent runtime is measurement revocation: when an image is broken, kill it fleet-wide in one governed action rather than trusting each operator to self-deactivate.

## Proposed change

Add a **governance-gated policy revocation denylist** to `x/agent`, keyed by a deterministic fingerprint of the agent's `tee.Policy`. A revoked policy can neither register new agents nor append attested actions.

### Fingerprint (new: `x/agent/types/policy.go`)
```go
// PolicyFingerprint is the lowercase hex SHA-256 of the deterministic proto
// marshaling of the policy. Two agents pinning a byte-identical policy (same
// root cert, rego, and expected measurement) share a fingerprint; that is
// exactly the set we revoke together when an image is found vulnerable.
func PolicyFingerprint(p tee.Policy) (string, error) {
    b, err := p.Marshal()           // gogoproto generated, deterministic (no maps, all scalar fields)
    if err != nil { return "", err }
    h := sha256.Sum256(b)
    return hex.EncodeToString(h[:]), nil
}
```
`tee.Policy` contains only string fields, so `Marshal()` is stable across validators — no map iteration, no floats. The fingerprint is O(1) to compute.

### Store (`x/agent/types/keys.go`, `x/agent/keeper/keeper.go`)
- New prefix `KeyRevokedPolicies = []byte{0x03}`.
- New keeper field `revokedPolicies collections.KeySet[string]` (hex fingerprint → present == revoked).
- Add `authority string` to `Keeper` (the x/gov module account), mirroring `x/sponsorship/keeper/keeper.go:16,41,48-55`. Validate with `sdk.AccAddressFromBech32` and panic on invalid, as sponsorship does. Wire it in `app/keepers.go:596` by passing `authtypes.NewModuleAddress(govtypes.ModuleName).String()` as a new arg to `agentkeeper.NewKeeper`.
- Keeper helpers: `SetRevoked(ctx, fp) / DeleteRevoked(ctx, fp) / IsPolicyRevoked(ctx, fp) (bool, error) / AllRevokedPolicies(ctx) ([]string, error)`.

### Messages (`proto/dymensionxyz/dymension/agent/tx.proto` + regen; handlers in a new `x/agent/keeper/revoke.go`)
- `MsgRevokePolicy { string authority = 1; string fingerprint = 2; string reason = 3; }` → response empty. Handler: require `msg.Authority == k.authority` (else `gerrc.ErrPermissionDenied`); `ValidateBasic` checks `fingerprint` is 64-char lowercase hex and `authority` is a valid bech32 addr; add fingerprint to the set (idempotent — revoking an already-revoked fp succeeds as a no-op); emit `EventPolicyRevoked{fingerprint, reason}`.
- `MsgUnrevokePolicy { string authority = 1; string fingerprint = 2; }` → removes the fingerprint (no-op if absent, so a false-alarm is fully reversible with zero side effects on agents); emit `EventPolicyUnrevoked{fingerprint}`.
- Register both in `x/agent/types/codec.go` (both `RegisterConcrete` for amino and `RegisterImplementations` under `sdk.Msg`), and expose them in the CLI tx command tree (`x/agent/client/cli`).

### Enforcement (pure denylist — never mutates `agent.Active`)
- **`SubmitAttestedAction`** (`x/agent/keeper/msg_server.go`): after loading the agent and the `Active` check, before the `verifier.Verify` call, compute `fp := PolicyFingerprint(agent.Policy)` and if `k.IsPolicyRevoked(ctx, fp)` return `gerrc.ErrFailedPrecondition.Wrapf("agent policy revoked: %s", fp)`. This is the core security property: once revoked, **no** agent with that policy can ever append another action. O(1), checked every submit.
- **`RegisterAgent`** (`x/agent/keeper/registry.go`): reject registration when the incoming policy's fingerprint is already revoked, with `gerrc.ErrFailedPrecondition` — no point admitting a known-bad image.
- Deliberately do **not** flip `agent.Active`. The revocation set is the single source of truth; effective-active = `agent.Active && !revoked`. This keeps `MsgUnrevokePolicy` perfectly reversible (no orphaned `Active=false` agents needing a re-activate message that doesn't exist).

### Queries (`proto/.../agent/query.proto` + regen; `x/agent/keeper/grpc_query.go`)
- `RevokedPolicies() -> { repeated string fingerprints }` (deterministic ascending order via the KeySet iterator).
- `PolicyRevoked(fingerprint) -> { bool revoked }`.
- Extend the existing single-`Agent` query response with two derived read-only fields `string fingerprint` and `bool revoked` so operators/governance can read an agent's fingerprint straight off-chain to know what to revoke. Also surface `fingerprint` in `EventRegisterAgent` (`x/agent/types/events.go` / registry handler) for the same reason.

### Genesis (`x/agent/keeper/genesis.go`, `x/agent/types/genesis.pb.go`)
- Add `repeated string revoked_policies` to `GenesisState`; export from the KeySet in ascending order and import in `InitGenesis`. Validate each entry is 64-char hex in `GenesisState.Validate`.

- **Fingerprint = SHA-256 of `policy.Marshal()`, lowercase hex.** Coarse-but-correct: it revokes exactly the agents with a byte-identical policy, which is the honest fleet running the same published image (per-agent uniqueness comes from the attestation nonce, not the policy). Agents with intentionally customized policies get distinct fingerprints and are revoked individually — governance reads each via the `Agent` query. We do **not** parse/substring-match `policy_values` JSON to extract a measurement: there is no fixed schema for it and substring matching risks over-revocation.
- **Authority = x/gov module account.** Same pattern as `x/sponsorship`/`x/iro`. No new param.
- **Enforcement is lazy/point-lookup only; no iteration over agents at revoke time.** Avoids unbounded O(N) state iteration in a single tx. The security guarantee holds regardless because every action submit re-checks.
- **Reversible:** `MsgUnrevokePolicy` exists for false alarms and has no agent side effects.
- **Idempotent** revoke/unrevoke; **no** panics anywhere (all paths return errors); **no** Begin/EndBlocker involvement.

## Acceptance criteria
- New `revoke_test.go` (keeper) proves: (1) `MsgRevokePolicy` from the gov authority adds the fingerprint; from any other address returns `ErrPermissionDenied` and does not mutate state. (2) After revoking policy P, `SubmitAttestedAction` for every agent whose policy fingerprints to P is rejected with the revoked error and appends **nothing** to the action log (assert `action_seq` unchanged and no new log entry) — verified with the fake verifier that would otherwise accept. (3) `RegisterAgent` with a revoked policy is rejected. (4) `MsgUnrevokePolicy` re-enables submit for those agents and leaves `agent.Active` untouched throughout. (5) `PolicyFingerprint` is stable: marshaling the same policy twice yields identical fingerprints, and two agents with equal policies collide while a one-byte policy change does not.
- `RevokedPolicies` / `PolicyRevoked` queries return correct results; `Agent` query returns the right `fingerprint` and `revoked`.
- Genesis export→import round-trips the revoked set; `GenesisState.Validate` rejects a non-hex entry.
- `make proto-gen` produces no uncommitted diff; `go test ./x/agent/...` and `golangci-lint run` pass.

## Out of scope
- Do **not** modify `x/common/tee` verifier logic, OPA evaluation, or the `Policy` proto shape.
- Do **not** add bonding/slashing/economic penalties — this is revocation only.
- Do **not** touch other TEE consumers (e.g. `x/rollapp`); this denylist is local to `x/agent`.
- Do **not** add automatic on-chain expiry/timers — revocation is governance-driven and manual.
- Do **not** change `MsgDeactivateAgent` semantics or add a re-activate message.

---

omritoptix:
Closes #2179

## What

Adds a governance-gated revocation denylist to `x/agent`, keyed by a fingerprint of the agent's TEE policy (lowercase hex SHA-256 of its deterministic proto marshaling). When an enclave image is found vulnerable, one gov action kills the whole fleet pinned to it — instead of trusting each operator to self-deactivate per agent.

- **`MsgRevokePolicy` / `MsgUnrevokePolicy`** — authority is the x/gov module account (same pattern as `x/sponsorship`); any other signer gets `ErrPermissionDenied`. Both are idempotent.
- **Enforcement** — `SubmitAttestedAction` and `RegisterAgent` reject when the policy fingerprint is in the denylist (`ErrFailedPrecondition`). O(1) point lookup per submit; no iteration over agents at revoke time.
- **Pure denylist** — `agent.Active` is never mutated, so unrevoking a false alarm restores every affected agent with zero side effects.
- **Queries** — `RevokedPolicies` (ascending), `PolicyRevoked(fingerprint)`, and the single-`Agent` query now returns derived `fingerprint` + `revoked` fields so governance can obtain the fingerprint to revoke from a query instead of computing it. `EventRegisterAgent` also surfaces the fingerprint.
- **Genesis** — the revoked set round-trips through export/import; `GenesisState.Validate` rejects entries that aren't 64-char lowercase hex.
- **CLI** — `tx agent revoke-policy|unrevoke-policy`, `q agent revoked-policies|policy-revoked`.

### Revocation logic: authority gating, fleet-wide submit block, register block, reversibility, fingerprint stability, queries, genesis round-trip
**How:** `go test ./x/agent/... -run 'Revoke|Fingerprint|Genesis' -v`
**Evidence:**
```
=== RUN   TestGenesisRoundTripWithActionLog
--- PASS: TestGenesisRoundTripWithActionLog (0.01s)
=== RUN   TestPolicyFingerprint_Stable
--- PASS: TestPolicyFingerprint_Stable (0.00s)
=== RUN   TestRevokePolicy_AuthorityGated
--- PASS: TestRevokePolicy_AuthorityGated (0.00s)
=== RUN   TestRevokePolicy_ValidateBasic
--- PASS: TestRevokePolicy_ValidateBasic (0.00s)
=== RUN   TestSubmitAttestedAction_RevokedPolicy
--- PASS: TestSubmitAttestedAction_RevokedPolicy (0.00s)
=== RUN   TestRegisterAgent_RevokedPolicy
--- PASS: TestRegisterAgent_RevokedPolicy (0.00s)
=== RUN   TestRevokedPoliciesQueries
--- PASS: TestRevokedPoliciesQueries (0.00s)
=== RUN   TestGenesisRoundTripRevokedPolicies
--- PASS: TestGenesisRoundTripRevokedPolicies (0.00s)
=== RUN   TestGenesisValidate_BadFingerprint
--- PASS: TestGenesisValidate_BadFingerprint (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.151s
```

### Module and full test suite
**How:** `go test ./x/agent/...` then `go test ./...`
**Evidence:**
```
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.503s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.070s
```
Full suite: 35 packages `ok`, one failure that turned out to be a pre-existing build break on `origin/main` (`x/eibc/keeper`: `lp_test.go:165` called `types.NewDemandOrder` with the pre-#2168 arity — a semantic merge conflict between #2168 and #2170). This PR carries the one-line fix in its second commit; after it:
```
$ go test ./x/eibc/keeper/...
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper	3.192s
$ go vet ./...
exit=0
```

### Lint and generated code are clean
**How:** `golangci-lint run ./x/agent/... ./app/...` (v2.x), `gofumpt -l x/agent app/keepers.go`, re-ran `make proto-gen` and checked `git status`
**Evidence:**
```
$ golangci-lint run ./x/agent/... ./app/...
0 issues.
$ gofumpt -l x/agent app/keepers.go
(no output)
$ make proto-gen; git status --porcelain
exit=0
(only the files in this PR — no drift)
```

### Live chain end-to-end: queries, non-gov rejection, gov proposal executes the revocation
**How:** built `dymd` at HEAD, ran a throwaway single-validator chain (10s voting period), exercised the new queries and txs, then passed a real gov proposal carrying `MsgRevokePolicy`
**Evidence:**
```
=== query: revoked-policies (expect empty) ===
{}
=== query: policy-revoked f55df6c4...a640872 (expect false) ===
{}
=== query: policy-revoked with malformed fingerprint (expect error) ===
Error: rpc error: code = InvalidArgument desc = fingerprint must be 64 hex chars, got 7: invalid argument: invalid request
=== tx: revoke-policy from non-gov account (expect permission denied) ===
{
 "code": 6,
 "raw_log": "failed to execute message; message index: 0: expected authority dym10d07y265gmmuvt4z0w9aw880jnsr700jgllrna, got dym18nwg5pzyu0qkzf23ur8n9jd9v5cyjwxtydj2ev: permission denied"
}
=== tx: unrevoke-policy from non-gov account (expect permission denied) ===
{
 "code": 6,
 "raw_log": "failed to execute message; message index: 0: expected authority dym10d07y265gmmuvt4z0w9aw880jnsr700jgllrna, got dym18nwg5pzyu0qkzf23ur8n9jd9v5cyjwxtydj2ev: permission denied"
}
=== query: revoked-policies still empty after denied txs ===
{}
=== gov: submit MsgRevokePolicy proposal and vote yes ===
proposal 1 submitted and voted; waiting for voting period...
proposal status: PROPOSAL_STATUS_PASSED
=== query: policy-revoked f55df6c4...a640872 after gov proposal (expect true) ===
{
  "revoked": true
}
=== query: revoked-policies after gov proposal ===
{
  "fingerprints": [
    "f55df6c465a517854242a2be63d922b534849a52dfc64de733d3d0f84a640872"
  ]
}
```

### Rotation × revocation (post-merge of #2176): promote-then-check ordering, update-time denylist, effective-fingerprint query
**How:** merged `main` (timelocked policy rotation) into the branch; `SubmitAttestedAction` now promotes a matured pending policy *before* the denylist check, `UpdateAgentPolicy` rejects scheduling an already-revoked policy, and the `Agent` query fingerprints the effective policy. Regression tests: `go test ./x/agent/keeper/... -run 'Revoked|RevokePolicy|Fingerprint' -v`
**Evidence:**
```
=== RUN   TestUpdateAgentPolicy_RevokedPolicy
--- PASS: TestUpdateAgentPolicy_RevokedPolicy (0.00s)
=== RUN   TestSubmitAttestedAction_RevokedPendingPolicy
--- PASS: TestSubmitAttestedAction_RevokedPendingPolicy (0.00s)
=== RUN   TestSubmitAttestedAction_RevokedActiveCleanPending
--- PASS: TestSubmitAttestedAction_RevokedActiveCleanPending (0.00s)
=== RUN   TestAgentQuery_EffectiveFingerprint
--- PASS: TestAgentQuery_EffectiveFingerprint (0.00s)
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.370s

$ golangci-lint run ./x/agent/... ./app/... ./x/eibc/...
0 issues.
$ go vet ./...
exit=0
```

**Not verified:** the on-chain `SubmitAttestedAction` block/unblock path — a genuine GCP TEE attestation token cannot be minted locally, so that path is proven at the keeper layer with the fake verifier (`TestSubmitAttestedAction_RevokedPolicy` above).


omritoptix (commented):
## Review + security pass

**Verdict: Approve.**

Reviewed the diff line-by-line (generated `*.pb.go` checked against the `.proto` changes), did a separate security-focused pass, and independently reproduced the proof of execution.

### Verified
- **Authority gating** matches the x/sponsorship pattern: `cosmos.msg.v1.signer = "authority"` plus an explicit `msg.Authority != k.authority` check; the keeper panics on an invalid authority at construction; app wiring passes `govModuleAddress`.
- **Fingerprint determinism**: `tee.Policy` is four plain string fields, gogoproto `Marshal` writes fields in a fixed order, and unknown fields are discarded on unmarshal (no `XXX_unrecognized`), so there is no proto-malleability path to a divergent fingerprint and no consensus risk.
- **Enforcement** sits in both `RegisterAgent` and `SubmitAttestedAction`; the pure-denylist design (agent state untouched) makes unrevoke fully reversible, and revoke/unrevoke are idempotent via `collections.KeySet`.
- **Genesis**: `ValidateGenesis` calls `GenesisState.Validate()` (standard SDK pattern — the keeper-level `InitGenesis` not re-validating is consistent with every other module here); round-trip covered by tests; new store prefix `0x03` doesn't collide.
- **Proof of execution reproduced locally**: all 9 `Revoke|Fingerprint|Genesis` tests pass with `-count=1`, and `go build ./...` is clean. The eibc failure the PR flags is a pre-existing test-compile break (`x/eibc/keeper/lp_test.go:165: not enough arguments in call to types.NewDemandOrder`) in a package this PR doesn't touch, so it can't be caused by this change.
- **Security pass**: no high- or medium-severity security findings.

### Non-blocking notes
1. **The denylist is byte-exact while the verifier is semantic.** `PolicyValues` is JSON-parsed, the PEM is decoded, and the rego is compiled, so a byte-different but semantically identical policy verifies the same yet gets a fresh fingerprint. Existing agents are correctly killed fleet-wide, but re-registration with a trivially reformatted policy evades the register-time check. This is inherent to a denylist over a permissionless registry (and the `PolicyFingerprint` comment scopes revocation to byte-identical policies correctly), but `doc.go`'s "a revoked policy can neither register new agents" overstates it slightly — worth softening to "byte-identical policy". Measurement-keyed revocation could be a follow-up if this matters operationally.
2. **Duplicated fingerprint-check block.** The fingerprint + `IsPolicyRevoked` sequence appears three times (`RegisterAgent`, `SubmitAttestedAction`, `Agent` query). A small keeper helper returning `(fingerprint, error)` would keep the enforcement sites from drifting if a third msg type is ever added.
3. **`RevokedPolicies` query is unpaginated**, unlike the sibling `Agents`/`AgentActions` queries. Fine while the set is gov-curated and tiny; worth pagination only if that assumption changes.

omritoptix (commented):
## Re-review

**Verdict: Approve.** The only change since the prior review is commit 6693508e (one-line eibc test-arity fix); I re-ran a full-effort review over the whole PR anyway, including a security pass, and independently re-ran the test evidence.

### Prior findings — accounting

All three prior notes were explicitly non-blocking; none were addressed by the new commit, so all three remain open as follow-ups:

1. **Byte-exact denylist vs. semantic verifier / `doc.go` wording** — unchanged. `doc.go` still says "a revoked policy can neither register new agents"; the suggested softening to "byte-identical policy" (or a follow-up on measurement-keyed revocation) is still worth doing but stays non-blocking: the limitation is inherent to the design and correctly scoped in the `PolicyFingerprint` comment.
2. **Duplicated fingerprint + `IsPolicyRevoked` block** — still present in `RegisterAgent`, `SubmitAttestedAction`, and the `Agent` query. Extract a small keeper helper when convenient.
3. **Unpaginated `RevokedPolicies` query** — unchanged; fine while the set is gov-curated and tiny.

### New commit (6693508e)

Verified correct and complete: `NewDemandOrder` now takes 8 args (`feeEscalation` added by #2168); the fixed call passes `nil`, which doesn't change what the LP test asserts; it was the only stale call site (all 20 call sites audited); `go vet` and `go test ./x/eibc/keeper/...` pass. Carrying the fix here is justified — it repairs a semantic merge conflict on `main` (#2168 × #2170) that this PR's full-suite proof would otherwise trip over, and the PR body documents it transparently.

### Fresh review pass (7 angles + verification)

No new blocking issues. Candidates surfaced and refuted:
- **Missing nil-request guards in new query handlers** — consistent with the module's pre-existing handlers, and the SDK query router never delivers a nil request; not reachable.
- **x/rollapp TEE path not covered by the denylist** — not a gap: rollapp's TEE policy is a single gov-controlled param (`Params.TeeConfig`), so governance already has a kill switch there via param update; the denylist exists precisely for x/agent's permissionless per-agent policies.
- **Fingerprint recomputation per submit** — the policy is four small strings; recompute-over-cached-state is the right trade-off here.

Minor non-blocking nit: `SetRevoked`/`DeleteRevoked` are one-line pass-throughs over the `KeySet`; could be inlined.

### Security pass

Clean. Re-verified the key paths: authority equality check present in **both** `RevokePolicy` and `UnrevokePolicy` handlers, and each handler calls `ValidateBasic()` itself, so the lowercase-64-hex fingerprint validation holds even on the gov execution path; genesis fingerprints are validated via `ValidateGenesis`; `tee.Policy` is four plain string fields with deterministic marshal (no consensus risk); store prefix `0x03` doesn't collide.

### Proof of execution

Concrete and reproduced: I re-ran `go test ./x/agent/... ./x/eibc/keeper/... -count=1` locally — all `ok`. The PR body's live-chain evidence (non-gov rejection, gov proposal executing the revocation, query round-trips) is specific and self-consistent, and honestly scopes what wasn't verifiable locally (real GCP attestation). Behavior changes carry committed regression tests (`revoke_test.go`, 9 tests).

omritoptix (commented):
## Re-review (3rd pass)

**Verdict: Request changes.** The head is unchanged since the last approval (6693508e), and the PR's own code remains sound — the blocker is new: #2176 landed on `main` and conflicts with this PR at a security-relevant spot, so the required rebase is not mechanical.

### Prior findings — accounting

No new commits since the last review, so all four prior notes are unchanged and stay **non-blocking**:

1. **Byte-exact denylist vs. semantic verifier / `doc.go` wording** — still open. Note it now compounds with the rotation issue below: `MsgUpdateAgentPolicy` adds another same-class evasion path (rotate to a trivially reformatted policy), so the `doc.go` invariant wording should be revisited during the rebase anyway.
2. **Duplicated fingerprint + `IsPolicyRevoked` block** (RegisterAgent, SubmitAttestedAction, Agent query) — still open. The rebase is a good moment to extract the helper, since a fourth enforcement site may be needed (see below).
3. **Unpaginated `RevokedPolicies` query** — still open, fine for a gov-curated set.
4. **`SetRevoked`/`DeleteRevoked` one-line pass-throughs** — still open, cosmetic.

### New blocking issue — conflict with #2176 (timelocked policy rotation)

The PR is currently **unmergeable (CONFLICTING)**, and the central conflict sits exactly at the enforcement point: on `main`, `SubmitAttestedAction` now promotes a matured `PendingPolicy` immediately after the `!agent.Active` check — the same spot where this PR inserts its revocation check. The resolution ordering matters:

- **Required order: promote, then check.** If the revocation check runs before promotion, an agent whose matured *pending* policy has been revoked gets it promoted and verified in that same tx (the check passes against the still-active clean policy, then promotion swaps in the revoked one) — one attested action executes under a revoked policy before subsequent submits are blocked. The reverse case is also wrong under check-first: an agent with a revoked stale active policy but a clean matured pending policy is blocked when it should promote and proceed.
- **`MsgUpdateAgentPolicy` has no denylist check on the new policy.** With promote-then-check ordering, submit-time enforcement still catches a queued revoked policy, so there is no hard bypass — but decide explicitly whether queueing a revoked policy should be rejected at update time, consistent with register-time blocking. Silence here means the denylist gates register-time and submit-time but not update-time — two of the three paths by which a policy is set (`RegisterAgent`, `UpdateAgentPolicy`, promotion in `SubmitAttestedAction`).
- **Regression coverage:** the rebased PR needs a test for the rotation × revocation interaction (matured revoked pending policy must not verify; matured clean pending policy must verify even if the old active policy is revoked). The existing `revoke_test.go` cannot cover this because rotation didn't exist on this branch.

None of this is a defect in the code as written — it predates #2176 — but since the merge cannot happen without resolving it, and a naive textual resolution produces the check-before-promote gap, it has to be addressed before approval.

### Fresh pass + security review

Re-ran a full multi-angle review and a dedicated security pass over this head: no new findings in the PR's own code. Authority gating, fingerprint determinism (four plain string fields, canonical lowercase hex), genesis validation, and store-prefix isolation all re-verified clean.

### Proof of execution

Unchanged from the prior review and previously reproduced. Re-ran `go test ./x/agent/... ./x/eibc/keeper/... -count=1` on this head today:
```
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.857s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.075s
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper	4.649s
```

### Requested changes (summary)

1. Rebase onto `main`; in `SubmitAttestedAction`, place the revocation check **after** the pending-policy promotion.
2. Decide and implement denylist behavior for `MsgUpdateAgentPolicy` (reject queueing a revoked policy, or document why submit-time enforcement suffices).
3. Add a regression test for rotation × revocation ordering.

omritoptix:
Addressed the requested changes in merge commit eb280356 (merged `main` rather than rebasing, per repo convention):

- **Promote-then-check ordering**: `SubmitAttestedAction` promotes a matured pending policy first, then runs the denylist check on the effective policy. A matured revoked rotation is blocked even if the old policy was clean (no action verifies under it — promotion is in-memory and only persists via the final `SetAgent`, so nothing lands on the error path), and a clean matured rotation escapes a revoked active policy.
- **Update-time denylist behavior**: `UpdateAgentPolicy` rejects scheduling a rotation to an already-revoked policy (`ErrFailedPrecondition`, same rationale as `RegisterAgent`); the submit-time re-check remains the backstop for policies revoked after scheduling.
- **Rotation × revocation regression tests**: `TestSubmitAttestedAction_RevokedPendingPolicy` (blocked at maturity, nothing appended or persisted, unrevoke restores promotion), `TestSubmitAttestedAction_RevokedActiveCleanPending`, `TestUpdateAgentPolicy_RevokedPolicy`, `TestAgentQuery_EffectiveFingerprint` — all passing.
- Also picked up two of the prior non-blocking notes while in there: the duplicated fingerprint-check block is now a single `fingerprintNotRevoked` keeper helper used by all three handlers, and the `Agent` query fingerprints the effective (post-promotion) policy so its `revoked` field matches what enforcement would do.

PR body's proof-of-execution section updated with the new test run; `golangci-lint` and `go vet ./...` clean.

omritoptix (commented):
## Re-review (4th pass) — after merge of main (eb280356)

**Verdict: Approve.** The merge commit resolves all three blocking items from the previous review, and I verified each in code and by re-running the tests. One new non-blocking finding surfaced from the merge combination (eibc LP binding × revocation) — flagged below for an explicit follow-up decision.

### Prior blocking items (3rd review) — all resolved

1. **Promote-then-check ordering in `SubmitAttestedAction`** — ✅ resolved correctly. The merge places the denylist check *after* the matured-pending-policy promotion, so enforcement runs on the effective policy: a matured revoked rotation is blocked even when the old policy was clean, and a clean matured rotation escapes a revoked old policy. Verified in the handler and via `TestSubmitAttestedAction_RevokedPendingPolicy` / `TestSubmitAttestedAction_RevokedActiveCleanPending`, which assert exactly these two scenarios (including that a blocked promotion persists nothing).
2. **`MsgUpdateAgentPolicy` denylist behavior** — ✅ decided and implemented: scheduling a rotation to an already-revoked policy is rejected at update time, with a comment documenting that the submit-time check remains the backstop for policies revoked after scheduling. Covered by `TestUpdateAgentPolicy_RevokedPolicy`.
3. **Rotation × revocation regression coverage** — ✅ added: the two ordering tests above plus `TestAgentQuery_EffectiveFingerprint` (the Agent query fingerprints `EffectivePolicy`, so reported revocation status matches what submit-time enforcement would apply). All pass with `-count=1`.

### Prior non-blocking notes — accounting

1. **Byte-exact denylist vs. semantic verifier / `doc.go` wording** — still open. `doc.go` still says "a revoked policy can neither register new agents"; the suggested "byte-identical" softening wasn't applied. Update-time rejection narrows the rotation evasion path for byte-identical policies, but the reformatted-policy evasion is inherent and remains. Follow-up.
2. **Duplicated fingerprint + `IsPolicyRevoked` block** — ✅ resolved: extracted into `keeper.fingerprintNotRevoked`, used by all three enforcement sites (`RegisterAgent`, `UpdateAgentPolicy`, `SubmitAttestedAction`). The `Agent` query intentionally keeps its own sequence since it needs the boolean rather than an error — fine.
3. **Unpaginated `RevokedPolicies` query** — still open, still fine for a gov-curated set.
4. **`SetRevoked`/`DeleteRevoked` one-line pass-throughs** — still open, cosmetic.

### New finding (non-blocking, decision needed): eibc agent-bound LPs don't see revocation

Main's #2183 (merged in here) gates on-demand LP matching on agent liveness, and its liveness check is `found && agent.Active` only. Because revocation is deliberately a pure denylist that never mutates `Active`, an agent whose policy is revoked is fully dead in x/agent (cannot submit attested actions) yet still reports "live" to x/eibc — its bound LPs keep auto-fulfilling. This partially undercuts the feature's rationale ("one gov action kills the whole fleet — instead of trusting each operator to self-deactivate"): on the LP surface, operators must still self-deactivate. I assess it non-blocking because no attacker capability is gained — LP funds and parameters are gated by owner keys, not the enclave, and fulfillment triggering is permissionless regardless — but the inconsistency between "can the agent act" and "is the agent live for LP purposes" deserves an explicit decision as a follow-up: either have the eibc liveness check also consult the denylist on the effective policy (requires extending eibc's `AgentKeeper` expected-keeper interface), or document that revocation intentionally does not affect LP matching.

Minor nit (pre-existing on main, not introduced here): the promotion condition in `SubmitAttestedAction` inlines the same maturity predicate as `Agent.EffectivePolicy`; a shared helper would prevent drift.

### Security pass

Clean — no high/medium findings. Re-verified on the merged head: authority equality check present in both `RevokePolicy` and `UnrevokePolicy` with explicit `ValidateBasic()` calls (so lowercase-64-hex canonicalization holds on the gov path — no case-mismatch route to an ineffective revocation); all writes to the denylist go through validated entry points (two handlers + validated genesis); fingerprint is SHA-256 over deterministic marshaling of a strings-only proto (no consensus risk); enforcement sits before any irreversible state change in all three handlers.

### Proof of execution

Concrete and independently reproduced on the merged head:
```
$ go test ./x/agent/... ./x/eibc/keeper/... -count=1
ok  github.com/dymensionxyz/dymension/v3/x/agent/keeper   1.912s
ok  github.com/dymensionxyz/dymension/v3/x/agent/types    0.264s
ok  github.com/dymensionxyz/dymension/v3/x/eibc/keeper    6.972s
$ go build ./...   # exit 0
```
All 12 `Revoked|RevokePolicy|Fingerprint` tests pass individually with `-v`, matching the PR body's evidence, and the body's live-chain gov e2e (non-gov rejection, proposal-driven revocation, query flips) remains valid for the unchanged handler logic.

omritoptix (commented):
## Re-review (5th pass)

**Verdict: Request changes.** This is a max-effort/security-sensitive review because the change governs whether TEE-backed actions can execute. The core implementation remains sound, but the merge introduced an inaccurate package-level invariant that must be corrected under the docs-honesty requirement.

### New blocking finding

- **[Docs] `x/agent/doc.go:2` says agents register with an “immutable attestation policy,” but policies are explicitly mutable.** The merged branch exposes `MsgUpdateAgentPolicy`, stores `PendingPolicy`, and promotes it after the timelock (`x/agent/keeper/registry.go:88-132`, `x/agent/types/agent.go:5-11`). This sentence was added by this PR and now directly contradicts the code shipped in the same head. Please describe the policy as current/timelocked-rotatable (and preferably soften the existing byte-exact denylist wording noted below). This is documentation-only to fix, but it is blocking because the stated security model is false.

### Prior findings — complete accounting

- The three former rotation blockers are **resolved**: submit promotes before checking revocation; `UpdateAgentPolicy` rejects an already-revoked target; and the committed rotation × revocation tests cover both ordering directions plus effective-policy query behavior.
- Byte-exact fingerprinting vs semantic policy equivalence, including the over-broad “revokes compromised TEE images fleet-wide” / “a revoked policy can neither register” wording, **remains open and non-blocking** under issue #2179's explicitly byte-identical design. It would be sensible to clarify this while fixing the new immutable-policy error.
- Duplicated fingerprint enforcement is **resolved** by `fingerprintNotRevoked`.
- Unpaginated `RevokedPolicies` **remains open and non-blocking** for a governance-curated set.
- `SetRevoked` / `DeleteRevoked` pass-through helpers **remain open and cosmetic**.
- The prior eIBC agent-bound LP observation **remains open and non-blocking**: eIBC liveness checks `Active`, not revocation. The authoritative issue scopes enforcement to registration and attested-action submission, deliberately keeps `Active` unchanged, and explicitly excludes other TEE consumers; no new unauthorized fund capability was demonstrated.

### Security and proof

No new high/critical security defect found. Authority equality and signer declaration are present for revoke/unrevoke; fingerprints are canonical lowercase 64-hex; enforcement covers registration, policy update, and effective-policy submit before attestation/log writes; genesis validates imported fingerprints; rotation error paths do not persist the in-memory promotion.

The PR's **Proof of execution is concrete and sufficient**: it names commands, includes keeper regression output, live-chain non-gov and governance execution evidence, lint/generated-code checks, and honestly identifies the real-attestation limitation. Current CI is green for build/test, lint, CodeQL, and buf break checks. I attempted a fresh targeted run, but this review environment has no `go` executable; `git diff --check` is clean.

omritoptix (commented):
## Re-review (6th pass)

**Verdict: Approve.** This was a max-effort/security-sensitive re-review because the PR governs whether TEE-backed actions may execute. The only new commit, e91c6eca, correctly resolves the prior documentation blocker, and I found no new issue introduced by it.

### Prior findings — complete accounting

- **Inaccurate immutable-policy invariant — resolved.** `x/agent/doc.go` now accurately says policies may rotate after a timelock and actions use the policy currently in force.
- **Byte-exact denylist wording — resolved.** The package docs now explicitly scope enforcement to byte-identical policies/fingerprints instead of claiming semantic image-wide revocation.
- **Rotation × revocation blockers — resolved and unchanged:** submit promotes a matured pending policy before checking the denylist; `UpdateAgentPolicy` rejects an already-revoked target; committed regression tests cover both ordering directions and effective-policy query behavior.
- **Duplicated enforcement logic — resolved** by `fingerprintNotRevoked`.
- **Unpaginated `RevokedPolicies` — remains open, non-blocking** for a governance-curated set.
- **`SetRevoked` / `DeleteRevoked` pass-through helpers — remain open, cosmetic.**
- **eIBC agent-bound LP liveness ignores revocation — remains open, non-blocking.** eIBC still keys liveness on `Active`, while this feature deliberately leaves `Active` untouched and issue scope limits enforcement to registration/update/action submission. This deserves an explicit future product decision but does not create a demonstrated authorization or funds bypass in this PR.

### Security, tests, proof, and docs

No unresolved high/critical security issue found. Authority checks, canonical fingerprint validation, validated genesis import, and denylist enforcement before persistent action/log writes remain intact. The latest commit changes only package documentation and now matches the implemented timelocked rotation and byte-exact revocation model; I found no other newly inaccurate README/docs.

The PR body's **Proof of execution is concrete and sufficient**: it names exact commands, includes focused regression results, live-chain non-governance rejection and governance execution evidence, lint/generated-code checks, and honestly scopes the unavailable real GCP attestation path. The behavior change has committed keeper regression coverage. I could not independently rerun Go tests in this review container because `go` is not installed; `git diff --check` passes. At review time, buf/label checks pass and the freshly triggered build, lint, and CodeQL checks are still running.

PR #2183, opened 2026-07-07
feat(eibc): bind on-demand LP to x/agent attested agent

### LINKED ISSUE #2182: x/eibc: bind on-demand LP to an x/agent attested agent (provenance + one-shot fleet kill-switch for autonomous market makers)
## Problem / opportunity

The eIBC **on-demand LP** (`x/eibc/keeper/lps.go`) is the on-chain home of autonomous market makers: a bot funds `funds_addr`, registers an `OnDemandLP`, and the chain then spends from that address **automatically** to fulfill matching demand orders (`Keeper.FulfillByOnDemandLP`, matched via `OnDemandLPRecord.Accepts`). It is a standing, machine-driven spend authority.

Today that authority is backed by a **plain bech32 address** with **no link to the identity or integrity of the bot operating it**. Two gaps follow:

1. **No provenance.** There is no on-chain way to assert that an autonomous LP is driven by a *genuine attested enclave* rather than an arbitrary key. The Hub already has a verifiable-agent-runtime (`x/agent`: TEE-attested agents with an immutable enclave-measurement `Policy`, #2162) — but eIBC LPs cannot reference it.
2. **No fleet-level kill-switch.** If a market-maker's hot key is compromised, the only remedy is to race the attacker and send a separate `MsgDeleteOnDemandLP` for **each** LP — using the very key that may be compromised. An operator running many LPs behind one enclave has no single, cold-key action to disable the whole fleet.

#2169 added time/velocity risk caps to the LP itself, but nothing there binds an LP to an attested *identity* or provides a one-shot fleet disable. This issue adds exactly that bridge between `x/eibc` and `x/agent`.

## Why now

The 2026 pattern for autonomous on-chain actors is **attested identity + revocable authority**: an agent runs in a TEE, proves its enclave measurement, and its authority can be revoked fleet-wide the instant the enclave is compromised (cf. the `x/agent` revocation-denylist work, #2179). eIBC on-demand LPs are the clearest case of a "standing autonomous spend authority" in this codebase, yet they sit outside that trust model. Wiring them to `x/agent` makes Dymension's fast-finality market-making layer a first-class consumer of its own verifiable-agent-runtime.

- Autonomous agents on blockchains — execution models, trust boundaries, revocation: https://arxiv.org/html/2601.04583v1
- x402 / agent-wallet economy 2026 (bounded, revocable autonomous authority as baseline): https://www.vaasblock.com/news/crypto-ai-agents-onchain-x402-wallet-economy-2026/

## Proposed change

Add an **optional** binding from an on-demand LP to an `x/agent` agent; gate fulfillment on that agent being live.

### Proto (`proto/dymensionxyz/dymension/eibc/lp.proto`, regen with `make proto-gen`)

- Add `string agent_id = 11;` to `OnDemandLP` (optional; empty ⇒ unbound, current behavior).

### Keeper wiring (`x/eibc/keeper/`)

- New expected-keeper interface `AgentKeeper interface { GetAgent(ctx, id string) (agenttypes.Agent, bool) }` in `x/eibc/types/expected_keepers.go`; inject the real `x/agent` keeper into the eIBC keeper in `app/keepers.go` (agent keeper is constructed before eIBC there).
- `CreateOnDemandLP` (`x/eibc/keeper/msg_server.go`): if `agent_id != ""`, require the agent to **exist** (`found`), else reject with `ErrNotFound`. Do not require it to be active at creation.
- `GetOrderCompatibleLPs` (`x/eibc/keeper/lps.go`, which already holds `ctx` + keeper): when iterating candidate LPs, in addition to `lpr.Accepts(h, &o)`, skip any LP whose `Lp.AgentId != ""` and whose referenced agent is **not `Active`**. Keep `OnDemandLPRecord.Accepts` pure (no keeper access) — the liveness gate lives in the keeper loop.

Because `FulfillByOnDemandLP` fulfils only from the `GetOrderCompatibleLPs` result, gating there is sufficient: deactivating an agent makes all its bound LPs non-compatible on the next match, with **no iteration over LPs at deactivation time** (lazy check).

- **`agent_id` is optional; empty preserves today's behavior.** Migration: existing `OnDemandLP` records decode with empty `agent_id` and are unaffected — no state migration needed.
- **The binding gates fulfillment only.** It does **not** move the LP's funds, change custody, or require the LP creator to equal the agent owner — an LP may reference any registered agent as its liveness oracle, enabling a shared-attested-operator model. The LP owner still funds `funds_addr` and can `MsgDeleteOnDemandLP` as before.
- **Liveness = `agent.Active`.** `DeactivateAgent` (owner-signed, typically a cold key distinct from the enclave hot key) disables every bound LP at once. This composes with the fleet-wide revocation denylist (#2179): a denylisted enclave measurement disables its whole LP fleet without any per-LP transaction.
- **Check point is `GetOrderCompatibleLPs`, not `Accepts`**, so `Accepts` stays a pure, keeper-free predicate (preserving its existing unit-test surface) while the state read stays in the keeper.
- **Require existence but not activeness at creation**, so an operator can pre-register LPs against an agent that is briefly inactive without a creation failure; only fulfillment requires liveness.

## Acceptance criteria

- `go test ./x/eibc/...` passes, including new tests:
  - LP bound to an **active** agent is matched and fulfils exactly as an unbound LP.
  - After `DeactivateAgent`, the bound LP is skipped by `GetOrderCompatibleLPs`/`FulfillByOnDemandLP`; a co-existing **unbound** LP still fulfils.
  - Reactivation is not required to be tested (no reactivation path exists in `x/agent`).
  - `CreateOnDemandLP` with a nonexistent `agent_id` is rejected; with `""` succeeds (unbound).
  - Genesis export→import round-trips the new `agent_id` field.
- No behavioral change for any existing LP with empty `agent_id`.

- Do not add per-fill attestation or change the automatic fulfillment flow — fills stay keeper-driven, no signed message per fill.
- Do not modify `x/agent`, the `#2169` velocity/validity fields, the Dutch-auction fee path (#2167), or LP fund custody.
- No agent→LP ownership coupling or auto-delete of LPs on deactivation (lazy skip only).

## Consensus-safety

The added gate is a single deterministic KV read of `agent.Active` during fulfillment — no wall-clock, no randomness, no map-iteration ordering, no Begin/EndBlocker work. Deactivation performs no LP iteration (lazy check at match time), so no unbounded loop. Identical result on every validator; tx atomicity is unchanged.

---

omritoptix:
Closes #2182

## What

Adds an optional binding from an eIBC on-demand LP to an `x/agent` TEE-attested agent, giving autonomous market makers on-chain provenance and a one-shot fleet kill-switch:

- `OnDemandLP.agent_id` (proto field 11, optional). Empty = unbound, exactly today's behavior; existing records decode unchanged, no state migration.
- `MsgCreateOnDemandLP` requires a referenced agent to **exist** (rejected with `ErrNotFound` otherwise) but not to be active, so LPs can be pre-registered against a briefly inactive agent.
- `GetOrderCompatibleLPs` skips any LP whose referenced agent is not `Active`. Since `FulfillByOnDemandLP` only fulfills from that result, `MsgDeactivateAgent` (cold-key, owner-signed) lazily disables the agent's whole LP fleet at the next match — no LP iteration at deactivation time. `OnDemandLPRecord.Accepts` stays a pure, keeper-free predicate; the state read lives in the keeper loop.
- New expected-keeper interface `AgentKeeper { GetAgent(ctx, id) (Agent, bool) }` in `x/eibc/types`, wired in `app/keepers.go`. Note: the agent keeper is constructed **after** the eIBC keeper there (the issue assumed the reverse), so injection follows the existing `SetDelayedAckKeeper` late-binding pattern.

Two adjacent things the acceptance criteria forced:

- **On-demand LPs were not part of eIBC genesis at all** — export dropped every LP record. Since the criteria require genesis to round-trip `agent_id`, genesis now exports/imports LP records plus the LP id sequence (so an imported chain can't reuse ids), with validation of records and id uniqueness.
- `x/eibc/keeper/lp_test.go` did not compile at HEAD (`orderWithSeq` still used the pre-#2167 `NewDemandOrder` arity); fixed in a standalone commit.

## Consensus-safety

The gate is one deterministic KV read of `agent.Active` per candidate LP during fulfillment — no wall-clock, no randomness, no Begin/EndBlocker work, no iteration at deactivation. Tx atomicity unchanged.

### Full test suite passes at HEAD

**How:** `go test ./...` (go1.24.6, CGO on), with `pipefail` and exit code captured
**Evidence:**
```
$ go test ./... ; echo exit=$?
ok  	github.com/dymensionxyz/dymension/v3/ibctesting
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types
ok  	github.com/dymensionxyz/dymension/v3/x/eibc
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/types
... (62 packages ok, zero FAIL lines)
exit=0
```

### Unit/keeper tests: binding, kill-switch gating, creation checks, genesis round-trip

**How:** `go test ./x/eibc/keeper/ -run 'TestKeeperTestSuite/(TestLPAgentBinding|TestCreateLPAgentBinding)' -v` and `go test ./x/eibc/ -run 'TestInitGenesis|TestExportGenesis' -v -count=1`
**Evidence:**
```
=== RUN   TestKeeperTestSuite
=== RUN   TestKeeperTestSuite/TestCreateLPAgentBinding
=== RUN   TestKeeperTestSuite/TestLPAgentBinding
--- PASS: TestKeeperTestSuite (0.20s)
    --- PASS: TestKeeperTestSuite/TestCreateLPAgentBinding (0.11s)
    --- PASS: TestKeeperTestSuite/TestLPAgentBinding (0.09s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper	0.311s
=== RUN   TestInitGenesis
--- PASS: TestInitGenesis (0.11s)
=== RUN   TestExportGenesis
--- PASS: TestExportGenesis (0.09s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/eibc	0.315s
```

### Live single-validator node: register agent → create bound/unbound/invalid LPs → deactivate → verify (proves app wiring end-to-end)

**How:** built `dymd` from this branch, ran a local node, drove the changed msg handlers. `agent_id` has no CLI flag (not in scope), so bound-LP txs were produced via `--generate-only` + jq + `tx sign` + `tx broadcast`. Agent registered with a real `tee.Policy` assembled from `x/common/tee/testdata` fixtures.
**Evidence (per-tx `query tx` results; steps grouped by theme, not run order — heights show actual chain order, the negative-path txs 3 and 7 ran last):**
```
=== 1. register agent "agent1" (x/agent) ===
{ "height": "651", "code": 0 }
=== 2. create on-demand LP bound to agent1 (agent_id via generated tx) ===
{ "height": "660", "code": 0 }
=== 3. create LP bound to NONEXISTENT agent: must be rejected ===
{ "height": "903", "code": 4,
  "raw_log": "failed to execute message; message index: 0: agent: ghost: not found" }
=== 4. create UNBOUND LP via plain CLI (back-compat) ===
{ "height": "734", "code": 0 }
=== 5. query LPs: bound LP shows agent_id, unbound is empty ===
{ "id": "0", "agent_id": "agent1", "funds_addr": "dym198kd2hfkte4w4eq8lnnzvup87pq9fc282nhrur" }
{ "id": "1", "agent_id": "",       "funds_addr": "dym198kd2hfkte4w4eq8lnnzvup87pq9fc282nhrur" }
=== 6. fleet kill-switch: deactivate agent1 (cold-key, one message) ===
{ "height": "740", "code": 0 }
query agent show-agent agent1 -> { "id": "agent1", "active": false }
=== 7. LP creation against existing-but-INACTIVE agent still allowed ===
{ "height": "928", "code": 0 }
final LP list: ids 0,2 bound to "agent1"; id 1 unbound
```

### Lint & format clean

**How:** `golangci-lint run ./x/eibc/... ./app/...` (v2.1.6) and `gofumpt -l x/eibc app/keepers.go`
**Evidence:**
```
0 issues.
```
(gofumpt printed no files.)

**Not verified:** fulfillment gating on the live node (needs a rollapp + sequencer + pending demand orders — impractical on a single-node smoke chain); it is covered by `TestLPAgentBinding`, which exercises `FulfillByOnDemandLP` end-to-end against the real app with real bank transfers.


omritoptix (commented):
## Review

Correctness, security, and test-execution pass; all claims verified locally against the checked-out branch.

### Correctness — no blocking findings

- **Wiring:** `SetAgentKeeper` is called in the only keeper construction path (`app/keepers.go`) before any message processing, and `EIBCKeeper` is passed by pointer everywhere in between, so the late-bound `agents` field can't be observed nil in practice — same shape as the existing `SetDelayedAckKeeper` pattern.
- **Genesis:** `LPs.Set` maintains all three indexes (byID/byAddr/byRollAppDenom), so imported LPs stay queryable by address and deletable. `ValidateGenesis` is wired and enforces record validity, id uniqueness, and `id < lps_next_id`, so imported ids can't collide with future `Create` ids. `OnDemandLPRecord.Validate` rejects nil `Lp` and negative `Spent`.
- **Gating:** `Accepts` has exactly one non-test caller, and it's the gated one; `FulfillByOnDemandLP` only fulfills from `GetOrderCompatibleLPs`, so the kill-switch has no bypass within the on-demand LP mechanism.
- **Proto:** new fields are append-only (lp field 11, genesis fields 3–4); no renumbering.

### Security

No findings. Binding requires only the LP creator's signature; referencing someone else's agent only lets that agent's owner disable your own LP's matching — opt-in and self-inflicted. Agent deactivation is owner-gated in x/agent. `agent_id` is used solely as a lookup key into the agent store; no injection surface.

### Proof of execution — verified

Re-ran the claimed commands locally: `go test ./x/eibc/...` all ok; `TestLPAgentBinding`, `TestCreateLPAgentBinding`, `TestInitGenesis`, `TestExportGenesis` all pass. The live-node evidence in the PR body is concrete (per-tx result codes including the negative path `code 4: agent: ghost: not found`) and matches the implemented behavior.

### Non-blocking notes

1. `agentLive` does one `GetAgent` read per candidate LP, so N LPs bound to the same agent re-read it N times per match. Negligible at realistic pool sizes; a per-call memo is an option if this grows.
2. The agent gate deliberately lives in the keeper loop to keep `Accepts` keeper-free. Fine today with a single caller, but a future direct caller of `Accepts` would silently skip the agent gate — worth a doc comment on `Accepts`.
3. `OnDemandLP.Validate` doesn't constrain `agent_id` (existence is checked only in the msg handler), so genesis can import an LP bound to a nonexistent agent. Harmless — it just never matches — and arguably desirable for ordering-independent genesis import; noting for awareness.
4. The kill-switch covers the on-demand LP mechanism only; authz-grant-based fulfillment (`MsgFulfillOrderAuthorized`) is untouched. That matches the issue scope — just flagging the boundary for the broader agent kill-switch story.

Also worth calling out: the PR surfaced and fixed a real pre-existing hole — on-demand LP records were silently dropped from genesis export.

**Verdict: Approve.**

PR #2184, opened 2026-07-07
feat(agent): attested spend authority — escrow-funded, budget-capped agent transfers

### LINKED ISSUE #2181: x/agent: attested spend authority — agent-initiated bounded transfers from an owner-funded escrow (v2 authority-to-act)
## Problem / opportunity

`x/agent` is a **verifiable agent runtime**: an agent registers an immutable TEE attestation `Policy` and then appends enclave-attested entries to an append-only action log (`x/agent/keeper/msg_server.go` `SubmitAttestedAction`, `x/agent/keeper/action_log.go`, nonce binding in `x/agent/types/nonce.go`). Today an attested action is a **pure log write** — `SubmitAttestedAction` records an opaque `payload` and advances `action_seq`; it never moves value. Agents can *assert* they did something; they cannot *do* anything on-chain.

The v1 epic (#2162) deliberately deferred this: its "Explicitly out of scope (future releases)" list reads *"no agent-initiated transactions, no escrow/bounties/marketplace ... 'Attested authority to act' ... are the natural v2/v3 built on this exact registry + verifier."* This issue specifies that v2 **authority-to-act** layer as a **bounded, escrow-funded, budget-capped spend primitive** — the smallest coherent step that lets an attested agent settle value, without fund vaults, bounties, or cross-agent composition.

This is a real gap in exactly one module (`x/agent`): the registry + verifier + action log exist, but there is no message by which an attested agent moves a single coin.

## Why now

The 2026 crypto-AI stack is built on **agent-native payment rails**: an agent that reasons off-chain but *settles* on-chain, under a **bounded, rate-limited spend authority** it cannot exceed. Coinbase **x402** (HTTP-402 agentic payments) crossed ~165M cumulative transactions with per-request stablecoin settlement, and the surrounding tooling treats *escrow + per-window spend caps* as table stakes for autonomous spend. A verifiable-agent-runtime that can attest what an agent *did* but cannot let it *pay* is missing the half that the market actually transacts on.

- Coinbase x402 — internet-native payments for AI agents: ``https://www.allium.so/blog/x402-explained-the-internet-native-payments-standard-for-apis-data-and-agent-commerce``/
- AWS — x402 and agentic commerce (autonomous settlement): https://aws.amazon.com/blogs/industries/x402-and-agentic-commerce-redefining-autonomous-payments-in-financial-services/
- x402 + gasless stablecoins 2026 — spend caps, batch settlement: https://www.autheo.com/blog/x402-gasless-stablecoins-ai-agent-micropayments-batch-settlement-2026

This complements, and does not overlap, the already-filed x402-adjacent work: #2169 (eIBC LP velocity caps — a *different* module, no attestation) and #2175 (agent *policy* rotation — no value movement). None of #2157–#2162, #2175, #2179 introduces agent-initiated value transfer.

## Proposed change

Add an **attested spend authority** to `x/agent`: an owner-funded per-agent escrow, a deterministic per-agent spend budget, and a new attested message that transfers from escrow to a recipient — authorized by the *same* TEE-attestation + nonce machinery already used for the action log.

### State / proto (`proto/dymensionxyz/dymension/agent/...`, regen with `make proto-gen`)

1. Extend `Agent` (`x/agent/types/agent.pb.go` source proto) with an optional **spend policy + window bookkeeping**:
   - `string spend_denom = 8;` — the single denom this agent may spend. Empty ⇒ spending disabled (pure-log agent, current behavior).
   - `string spend_limit_per_window = 9;` (`cosmossdk.io/math.Int`, non-nullable) — max spendable per rate window.
   - `uint64 spend_window_blocks = 10;` — window length in blocks. Must be `> 0` iff `spend_denom != ""`.
   - `uint64 spend_window_start_height = 11;` — aligned bucket start of the current window.
   - `string spend_window_spent = 12;` (`math.Int`, non-nullable) — spent so far in the current window.
2. New escrow ledger in the keeper: `collections.Map[string, sdk.Coins]` keyed by `agent_id`, under a new store prefix `KeyAgentEscrow = []byte{0x03}` (`x/agent/types/keys.go`). Pooled funds live in the **existing `agent` module account** (which already receives the registration fee via `SendCoinsFromAccountToModule`, so receiving is already permitted). Per-agent balances are tracked in the ledger map.

### Messages (`x/agent/keeper/msg_server.go`, `x/agent/types/tx.pb.go`, codec, CLI)

- `MsgFundAgentEscrow{ funder, agent_id, amount: Coins }` — anyone may fund. `SendCoinsFromAccountToModule(funder → agent)` then credit `escrow[agent_id] += amount`. Requires the agent to exist.
- `MsgWithdrawAgentEscrow{ owner, agent_id, amount: Coins }` — **owner-only**. Debit ledger (must not underflow) then `SendCoinsFromModuleToAccount(agent → owner)`. Ensures escrow is never trapped.
- `MsgUpdateAgentSpendPolicy{ owner, agent_id, spend_denom, spend_limit_per_window, spend_window_blocks }` — **owner-only, effective immediately** (see Decisions for why no timelock).
- `MsgSubmitAttestedTransfer{ submitter, agent_id, recipient, amount: math.Int, memo: bytes, token }`:
  1. `ValidateBasic`; enforce `len(memo) <= params.MaxActionBytes`.
  2. Load agent; require `agent.Active`; require `agent.SpendDenom != ""`.
  3. Promote a matured pending attestation policy (identical to `SubmitAttestedAction` lines 51–57).
  4. Derive the nonce over the **transfer**, so the enclave attests to *this* recipient+amount, not just an opaque blob:
     `payload := AttestedTransferBytes(recipient, agent.SpendDenom, amount, memo)`;
     `nonce := types.ActionNonce(agent_id, payload, agent.ActionSeq)`.
     Define `AttestedTransferBytes` in `x/agent/types/nonce.go` as the NUL-separated concatenation `recipient ‖ 0x00 ‖ spendDenom ‖ 0x00 ‖ amount.String() ‖ 0x00 ‖ memo` (0x00 cannot appear in bech32, a denom, or a decimal string, so the encoding is unambiguous and deterministic).
  5. `verifier.Verify(ctx, agent.Policy, nonce, token)` — reuses the existing seam; on failure the tx rolls back.
  6. **Budget check** (reuse the proven eIBC tumbling-window arithmetic from `x/eibc/types/lp.go` `Bucket`/`RateAllows`/`RecordSpend`): `bucket := h - h % spend_window_blocks`; window spent resets when `bucket != spend_window_start_height`; reject if `amount > spend_limit_per_window - windowSpentInThisBucket`.
  7. **Escrow debit + payout**: require `escrow[agent_id]` holds `amount spend_denom`; debit ledger; `SendCoinsFromModuleToAccount(agent → recipient, amount spend_denom)`.
  8. Update window bookkeeping, `agent.ActionSeq++`, append an `ActionLogEntry` (Payload = `payload` from step 4) so transfers share the single auditable, monotonic action log, emit `EventAttestedTransfer{agent_id, seq, recipient, amount}`.

### Wiring

- `x/agent/types/expected_keepers.go`: add `SendCoinsFromModuleToAccount(ctx, senderModule string, recipient sdk.AccAddress, amt sdk.Coins) error` to `BankKeeper` (the concrete `bankkeeper.Keeper` already satisfies it — no app change beyond the interface).
- Genesis (`x/agent/keeper/genesis.go`, `types/genesis.pb.go`): export/import the escrow ledger and the per-agent window state so state is reproducible.
- CLI (`x/agent/client/cli/`) + gRPC query for `EscrowBalance{agent_id}` and effective remaining window budget.

- **Single-denom budget.** An agent spends exactly one denom (`spend_denom`); the window cap is a scalar `math.Int`. Keeps window accounting one-dimensional and deterministic; multi-denom is out of scope.
- **Spend-policy updates are immediate and owner-only, NOT timelocked** — unlike the attestation `Policy` (timelocked in #2175 because it is a security *measurement* that must not be silently weakened). Rationale: the spend budget is a pure economic guardrail over the owner's *own* escrowed funds, and the owner can already `MsgWithdrawAgentEscrow` at will, so an immediate budget change grants no power the owner lacks. No new pending-field machinery.
- **Escrow is a pooled module account + per-agent ledger map**, not one bank account per agent — avoids address-derivation and blocked-address concerns and reuses the existing `agent` module account.
- **Nonce binds recipient+denom+amount+memo.** A tampered recipient or amount changes the nonce, so the attestation no longer matches and `Verify` rejects — the enclave, not the submitter, authorizes the exact payment.
- **Replay protection is structural**, identical to `SubmitAttestedAction`: `action_seq` advances on success, so resubmitting the same `(recipient, amount, token)` re-derives a different nonce and fails.
- **Window math is copied from eIBC** (`Bucket`/`RateAllows`/`RecordSpend`) — already audited, deterministic, height-based (no wall-clock).
- **`spend_denom == ""` ⇒ transfers rejected**, so existing pure-log agents are unaffected and the change is backward-compatible; a nil `spend_window_spent` decodes as zero (mirror `windowSpentAmount()` in eIBC).

## Acceptance criteria

- `go test ./x/agent/...` passes, including new tests:
  - Fund → transfer → recipient balance increases by exactly `amount`; escrow ledger and module-account balance both decrease by `amount`.
  - Transfer exceeding the window cap is rejected and **all** state (escrow, seq, log) is unchanged (atomic rollback).
  - Window rolls over: after `spend_window_blocks`, budget capacity resets even if a prior window was exhausted (mirror `x/eibc/types/lp_test.go` bucket tests).
  - Nonce binding: a fake verifier that recomputes `ActionNonce(agent_id, AttestedTransferBytes(...), seq)` rejects a token whose claimed nonce was computed for a *different* recipient/amount.
  - Insufficient escrow ⇒ rejected, no partial payout.
  - `MsgWithdrawAgentEscrow` by a non-owner is rejected; by the owner succeeds and cannot underflow the ledger.
  - Agent with empty `spend_denom` ⇒ `MsgSubmitAttestedTransfer` rejected with a clear error; `SubmitAttestedAction` still works.
  - Genesis export→import round-trips escrow balances and window state.
- Transfers appear in the same action log and advance the same `action_seq` as `SubmitAttestedAction`.

- No multi-denom budgets, no fund vaults/yield, no bounties/marketplace, no cross-agent composition, no slashing (all remain deferred per #2162).
- Do not modify the rollapp finalize path, `x/common/tee`, or `x/eibc`.
- Do not add per-fill attestation to any other module.

## Consensus-safety

Height-aligned tumbling windows (no wall-clock), scalar `math.Int` arithmetic, deterministic module-account transfers, no randomness, no map-iteration in consensus, no Begin/EndBlocker work. Every handler is a single atomic tx: any failure (verify, budget, insufficient escrow) rolls back all writes. Identical result on every validator.

---

omritoptix:
Closes #2181

## What

Adds the v2 **authority-to-act** layer to `x/agent`: an attested agent can now settle value, not just log actions.

- **Owner-funded escrow** — pooled in the existing `agent` module account with a per-agent ledger. `MsgFundAgentEscrow` (anyone may fund) and `MsgWithdrawAgentEscrow` (owner only, cannot underflow) ensure funds are never trapped.
- **Spend policy** — `MsgUpdateAgentSpendPolicy` (owner only, effective immediately) sets a single `spend_denom`, a per-window cap, and a window length in blocks. Empty denom keeps existing pure-log agents unaffected.
- **Attested transfer** — `MsgSubmitAttestedTransfer` binds the nonce to the exact recipient, denom, amount, and memo, checks the budget and escrow, pays from the module account, and appends to the shared action log.
- **CLI/query surface** — adds `fund-escrow`, `withdraw-escrow`, `update-spend-policy`, and `submit-attested-transfer`, plus the `escrow-balance` query.
- **Accounting/security hardening** — includes the escrow-solvency invariant, strict genesis validation, action/transfer nonce domain separation, full stored-payload bounds, and typed escrow/transfer events.

The branch also contains the eIBC `NewDemandOrder` test-call compatibility commit (`4a13f4a3`). It now merges current `main` (`5b69a417`), preserving the subsequently merged agent reputation registry, policy-revocation denylist, and eIBC agent binding. Main's existing store prefixes (`0x03`–`0x05`) and error codes (`8`–`13`) remain stable; escrow uses prefix `0x06` and spend errors use codes `14`–`16`. Revoked policies block attested transfers as well as plain attested actions.

## Notable decisions

- Spend-policy updates are immediate, not timelocked: this budget limits the owner's own escrow, which the owner can already withdraw.
- The budget is single-denom and scalar; multi-denom spending remains out of scope.
- Escrow uses a pooled module account plus a per-agent ledger, not per-agent module accounts.
- Spend-policy updates reset window bookkeeping.

### The transfer CLI is registered and constructs the requested message
**How:** `go test ./x/agent/client/cli -run TestSubmitAttestedTransferCommand -count=1` at `0c80003c`
**Evidence:**
```text
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	0.047s
```

### The merged revocation policy also gates attested transfers
**How:** `go test ./x/agent/keeper -run 'TestEscrowTestSuite/TestTransfer_RevokedPolicyRejected' -count=1` at `0c80003c`
**Evidence:**
```text
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.284s
```
The regression asserts rejection leaves the recipient balance, escrow ledger, and action sequence unchanged.

### The merged repository test suite is green, including agent and eIBC
**How:** `go test ./... -count=1` at `0c80003c`
**Evidence:**
```text
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	0.083s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	3.146s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.123s
ok  	github.com/dymensionxyz/dymension/v3/x/eibc	0.363s
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper	3.938s
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/types	0.164s
...
ok  	github.com/dymensionxyz/dymension/v3/x/vfc/hooks	0.387s
```
Command exited 0; all repository packages passed.

### The built CLI exposes the exact argument contract and rejects missing arguments
**How:** built `dymd` at `0c80003c`, then invoked the new command's help and its four-argument failure path
**Evidence:**
```text
$ dymd tx agent submit-attested-transfer --help
Submit an enclave-attested transfer from an agent's escrow

Usage:
  dymd tx agent submit-attested-transfer [agent-id] [recipient] [amount] [memo] [token] [flags]
help_exit=0

$ dymd tx agent submit-attested-transfer agent-1 dym1recipient 42 memo
Error: accepts 5 arg(s), received 4
bad_args_exit=1
```

### Go formatting, protobuf generation, and lint
**How:** `gofmt` on resolved Go files; `scripts/protocgen.sh` followed by `git diff --exit-code`; `golangci-lint run ./...` with v2.1.6
**Evidence:**
```text
protobuf regeneration: no diff
0 issues.
```

**Not verified:** a live `MsgSubmitAttestedTransfer` with a real GCP TEE token cannot be produced locally; the real bank keeper and module-account path is covered through the existing fake-verifier seam. Repository-wide `buf lint proto` is not green because the repository currently has a broad pre-existing lint baseline (documentation, naming, and package-option findings); this merge did not attempt unrelated schema cleanup.


omritoptix:
Addressed the review's hardening item in 21da834e: **genesis validation for the new spend/escrow state**.

`GenesisState.Validate()` now rejects:
- an enabled agent (`spend_denom` set) with `spend_window_blocks == 0` — the state that would make `SpendBucket` divide by zero — plus nil/non-positive limit or invalid denom;
- window bookkeeping inconsistent with the policy (spent > limit, negative spent, or window state on a disabled agent);
- escrow entries that are orphaned (no matching agent — funds would be trapped in the module account), duplicated, or empty/invalid.

The invariant is shared with `MsgUpdateAgentSpendPolicy.ValidateBasic` via a common `ValidateSpendPolicy` helper, so genesis and runtime can't drift. A companion test (`TestGenesisValidate_RuntimeStatesAccepted`) pins that states the runtime actually produces — freshly registered agents with nil Ints and exported disabled agents with zero-decoded Ints — still validate.

Evidence (`go test ./x/agent/... -count=1`):
```
--- PASS: TestGenesisValidate (0.00s)  (11 subtests, one per rejected state)
--- PASS: TestGenesisValidate_RuntimeStatesAccepted (0.00s)
ok  github.com/dymensionxyz/dymension/v3/x/agent/keeper  1.451s
ok  github.com/dymensionxyz/dymension/v3/x/agent/types   0.066s
```
golangci-lint: 0 issues. The two dismissed candidates (malicious-genesis DoS, ledger-before-send ordering) needed no code change beyond this; the ordering one is covered by Cosmos SDK atomic tx rollback as the review notes.

omritoptix (commented):
## Re-review — attested spend authority (escrow-funded agent transfers)

Reviewed at `max` effort (money + authz + TEE attestation surface) plus a security pass. The genesis-hardening commit (`21da834e`) resolved the earlier divide-by-zero / inconsistent-bookkeeping item, and the two previously dismissed candidates hold up: the ledger-debit-before-bank-send ordering is safe under Cosmos tx atomicity, and malicious-genesis DoS is out of scope. New confirmed findings below; the first is a blocker.

### 🔴 Blocker — a zero-value action can suppress an enclave-authorized payout (shared nonce namespace + seq counter)

`SubmitAttestedAction` and `SubmitAttestedTransfer` derive their attestation nonce with the **same** untagged `ActionNonce(agentID, payload, seq)` over the **same** `agent.ActionSeq`, and authorization is the token, not the signer. Every input to a transfer's payload is public — `recipient`/`amount`/`memo` are in the mempool message and `spend_denom` is a plaintext `Agent` field — so any observer can reconstruct `AttestedTransferBytes(...)` exactly.

Attack: an enclave mints a payout token bound to `seq = N`. An attacker copies the token from the mempool and submits `MsgSubmitAttestedAction{payload: <the reconstructed transfer bytes>, token: <copied token>}`. The action handler computes a bit-identical nonce, `verifier.Verify` passes, a log entry is written, and `ActionSeq` advances to `N+1`. The legitimate transfer token now re-derives against `N+1` and is permanently dead. A zero-value, unprivileged message consumes and invalidates a specific enclave-authorized payment — directly defeating the module's stated "the enclave, not the submitter, authorizes the payment" property, and it's repeatable to stall payouts or reorder competing transfers. This is an authorization/integrity defeat, not throughput griefing.

The same missing domain separation also means an action attestation and a transfer attestation are cryptographically indistinguishable at the verify layer — the only thing preventing an action token from authorizing a payout is the enclave's own discipline about which payloads it signs (that half is inside the TEE trust boundary, so lower severity, but it's the same root cause).

Fix: give the two message types disjoint nonce namespaces — add a domain/type tag to the hashed pre-image in `ActionNonce` (distinct constants for action vs transfer), and ideally separate the seq spaces so a transfer's slot can't be consumed by an action. `x/agent/keeper/msg_server.go:61,140`, `x/agent/types/nonce.go`.

### 🟠 Fund-accounting integrity

- **Escrow ledger vs. module-account balance is never reconciled, and no invariant is registered.** `GenesisState.Validate` checks per-entry escrow validity but not that `sum(escrow ledgers) == agent module-account bank balance`, `InitGenesis` writes the ledger blind, and `x/agent` registers no `RegisterInvariants` (every sibling that tracks coins against a module account — lockup, streamer, eibc, sponsorship — does). Because escrow is pooled, an over-claiming ledger (bad genesis or any future bookkeeping bug) lets one agent's withdrawal/transfer drain coins backing another agent's ledger, with later withdrawals dead-ending on bank insufficient-funds and nothing detecting the drift. `agent.InitGenesis` runs after bank init, so a genesis-time `GetBalance` cross-check is feasible; a registered balance-reconciliation invariant is the deeper fix. `x/agent/keeper/genesis.go:24`, `x/agent/keeper/keeper.go`.
- **Duplicate agent IDs pass genesis validation.** The escrow loop rejects duplicates but the agent loop does not, so two records for the same ID validate and `InitGenesis`' `agents.Set` silently keeps the last — a funded escrow can end up attached to whichever duplicate wins by array order (e.g. a spend-disabled one), with the reviewed policy dropped. One-line fix mirroring the escrow dedup, now load-bearing since escrows are keyed on agent ID. `x/agent/types/genesis.go:16`.

### 🟡 Correctness / consistency

- **`max_action_bytes` isn't applied to the stored transfer payload.** The transfer handler caps only `len(msg.Memo)`, but the `ActionLogEntry.Payload` it stores is `AttestedTransferBytes(recipient, denom, amount, memo)` — memo plus recipient + denom + amount + 3 separators. `SubmitAttestedAction` caps the exact bytes it stores; the transfer path lets entries exceed the param the proto documents as "the maximum allowed size of an attested action payload." Bounded overshoot, but an inconsistent invariant — cap `len(payload)`. `x/agent/keeper/msg_server.go:116`.
- **`ValidateSpendState` doesn't check `SpendWindowStartHeight` is bucket-aligned.** `SpendBucket` only ever yields multiples of `SpendWindowBlocks`, so an unaligned genesis start (a state the runtime can't produce) makes `RemainingWindowBudget` silently ignore `SpendWindowSpent` and hand back the full limit — importing an "exhausted" agent as freshly spendable. Genesis-only, owner's-own-guardrail leniency (and eibc's equivalent has the same gap), so low severity; worth an alignment check while you're validating. `x/agent/types/agent.go:30`.
- **Escrow handlers emit no events.** `FundAgentEscrow`, `WithdrawAgentEscrow`, `UpdateAgentSpendPolicy` mutate state and move coins but emit nothing, while every other handler in the module emits a typed event — off-chain indexers can't attribute escrow balance changes to an `agent_id` (bank events only show the pooled module address). Relatedly, `EventAttestedTransfer` omits the `submitter` that the action event records (recoverable from the tx, so minor). `x/agent/keeper/escrow.go`.

### 🧹 Cleanup (non-blocking)

- **`SubmitAttestedTransfer` copy-pastes ~45 lines of `SubmitAttestedAction`'s attest-and-log spine** — including the security-critical nonce-derivation + seq-advance replay flow, now duplicated verbatim. A hardening change to one path silently skips the money-moving one. Extract a shared `attestAndLog(ctx, agent, payload, token) (seq, error)` helper. This is also the cleanest place to introduce the domain tag from the blocker. `x/agent/keeper/msg_server.go`.
- **The pending-policy promotion block is duplicated in both handlers while `Agent.EffectivePolicy` has no non-test callers** — three encodings of the same maturity rule. Collapse to one `(a *Agent) PromotePendingPolicy(height)` method and delete/rebuild `EffectivePolicy` on it. `x/agent/types/agent.go:13`.
- The tumbling-window math mirrors `x/eibc/types` `OnDemandLP` nearly line-for-line (the code comment admits it); a small shared window helper would keep the two rate-limiters in lockstep. Modest, optional.

### Proof of execution

Solid: the `EscrowTestSuite` runs against the real bank keeper + module account with only the verifier faked, and asserts the acceptance criteria (exact credit/debit, cap rejection leaves state unchanged, window rollover, nonce binding, no partial payout, owner-only withdraw, spending-disabled, genesis round-trip). The honest "not verified" note on the live TEE path is appropriate. Note the suite does **not** cover the cross-type nonce collision above — the fake verifier accepts any token whose nonce matches, so a regression test submitting an action with a transfer-shaped payload at the pending seq would demonstrate the suppression directly.

---

**VERDICT: REQUEST_CHANGES** — the shared nonce namespace lets an unprivileged message permanently suppress enclave-authorized payouts; the missing ledger/bank reconciliation and duplicate-ID genesis gaps should land alongside it.

omritoptix:
All round-2 findings addressed in `9b1dc51d..c730aded`:

**🔴 Blocker — cross-type nonce replay** (`9b1dc51d`): nonce derivation now carries a domain byte (`0x01` action, `0x02` transfer) appended after the fixed-width `payloadHash‖seq` suffix, so the two nonce spaces are disjoint and a token only verifies for the message type it was minted for. Transfers use the new `TransferNonce`. Regression test `TestCrossDomainReplayRejected` pins both directions: a copied pending-transfer `(payload, token)` submitted as a plain action is rejected, the counter does not advance, and the original payout still lands; an action token likewise cannot authorize a transfer. Note this changes the enclave-visible nonce derivation for actions too (existing `ActionNonce` gains the tag) — nothing is deployed yet, so no migration.

**🟠 Integrity** (`548d5fee`, `c730aded`):
- `escrow-solvency` crisis invariant registered via the repo's `uinv` framework (module now implements `module.HasInvariants`, picked up by `app.mm.RegisterInvariants`): module-account balance must cover the ledger sum, and no ledger entry may be invalid or empty. `TestEscrowSolvencyInvariant` checks it across fund → transfer → withdraw.
- Genesis now rejects duplicate agent entries (was silently last-write-wins).

**🟡 Correctness** (`c730aded`):
- `max_action_bytes` now bounds the full stored transfer payload (`AttestedTransferBytes(...)`), not just the memo — `TestTransfer_PayloadTooLarge`.
- `ValidateSpendState` rejects window starts that aren't bucket-aligned — genesis subtest added.
- Typed events for escrow fund/withdraw/spend-policy update; `EventAttestedTransfer` now carries the submitter.

**🧹 Cleanup** (`c730aded`): the attest-and-log spine is shared (`loadAttestingAgent` + `appendAttested`); the policy-promotion block exists once as `Agent.PromotePendingPolicy`, and the dead `EffectivePolicy` is gone (its test became `TestAgent_PromotePendingPolicy`).

Evidence (`go test ./x/agent/... -count=1`): all pass, including
```
--- PASS: TestEscrowTestSuite/TestCrossDomainReplayRejected (0.13s)
--- PASS: TestEscrowTestSuite/TestEscrowSolvencyInvariant (0.07s)
--- PASS: TestEscrowTestSuite/TestTransfer_PayloadTooLarge (0.12s)
--- PASS: TestGenesisValidate/window_start_not_bucket-aligned (0.00s)
--- PASS: TestGenesisValidate/duplicate_agent_entry (0.00s)
ok  github.com/dymensionxyz/dymension/v3/x/agent/keeper  1.502s
ok  github.com/dymensionxyz/dymension/v3/x/agent/types   0.068s
```
`go build ./...` clean; golangci-lint (v2, repo config): 0 issues.

omritoptix (commented):
## Re-review — attested spend authority (follow-up commits `9b1dc51d`, `548d5fee`, `c730aded`)

Re-reviewed at max effort (money + authz + attestation) plus a security pass. Every finding from the prior review is accounted for below — all resolved or explicitly declined — and the new findings from this pass are non-blocking.

### Prior findings — all resolved

- 🔴 **Shared nonce namespace (blocker)** — resolved by `9b1dc51d`. Actions and transfers now derive nonces with disjoint domain tags (`0x01`/`0x02`) in the hashed pre-image, and `TestCrossDomainReplayRejected` pins both directions (copied transfer token rejected as an action with seq unchanged and the original transfer still succeeding; action token rejected as a transfer). The suggested regression test was added exactly as described. The "ideally separate the seq spaces" residual was not taken: consuming a transfer's seq slot now requires an enclave-minted *action* token for that same seq, which is inside the agent's own trust domain (an enclave that mints two tokens at one seq is misbehaving) — acceptable to leave as-is.
- 🟠 **No ledger/bank reconciliation** — resolved by `548d5fee` via the registered `escrow-solvency` invariant (wired through `module.HasInvariants` → `app.mm.RegisterInvariants`, verified), exercised across fund/transfer/withdraw in `TestEscrowSolvencyInvariant`. Two residual notes below.
- 🟠 **Duplicate agent IDs in genesis** — resolved in `c730aded` (`genesis.go` dedup + test case).
- 🟡 **`max_action_bytes` not applied to transfer payload** — resolved; the full payload is bounded before the (expensive) verify step, with `TestTransfer_PayloadTooLarge` asserting rejection and untouched seq.
- 🟡 **Window-start bucket alignment** — resolved in `ValidateSpendState` + genesis test. Ordering is safe: `ValidateSpendPolicy` runs first and rejects zero `SpendWindowBlocks` for enabled agents, so the new modulo cannot divide by zero.
- 🟡 **Escrow handlers emit no events / missing submitter** — resolved; typed events for fund/withdraw/spend-policy, `Submitter` added to `EventAttestedTransfer`.
- 🧹 **Attest-and-log duplication** — resolved (`loadAttestingAgent`/`appendAttested`); refactor fidelity checked hunk-by-hunk (error types/messages, check ordering, single-persist point all preserved).
- 🧹 **`EffectivePolicy` triplication** — resolved (`PromotePendingPolicy`, test ported).
- 🧹 **Shared window helper with eibc** — not taken (was optional), but see the `RecordSpend` note below: the mirror has now concretely drifted.

### New findings (none blocking)

- 🟠 **The attestation nonce binds no chain ID** — `attestNonce` commits to (agentID, payload hash, seq, domain) only. The comment says the layout mirrors `x/rollapp/types.TEENonce.Hash`, but `TEENonce` includes `HubChainId` in its pre-image *and* the rollapp handler enforces `nonce.HubChainId == ctx.ChainID()`; the agent flow has neither. Since submission is permissionless (the token, not the tx signer, is the authorization), sign-doc chain-id gives no protection: on a fork, or a second deployment where the same agent/policy/seq state exists, one enclave-minted transfer token pays out on every chain. The prerequisite (shared-state sibling chain) keeps this short of a blocker, but the module is unreleased, so adding `ctx.ChainID()` to the pre-image is a free one-liner now and a breaking token-format change after enclaves ship. Strongly recommend doing it in this PR or immediately after.
- 🟡 **Genesis `action_log` entries are not validated** — the strictness added for agents (dup IDs, spend state) and escrows (orphans, dups, zero balances) skips the third collection entirely: no unknown-agent check, no duplicate `(agent_id, seq)` check, and no check that entry seq < the agent's `action_seq`. A genesis entry at seq ≥ `action_seq` gets silently overwritten by the next attested action (`setActionLogEntry` is a bare `Set`), quietly breaking the append-only property. Same "runtime could have produced this" standard, one more loop — fine as a fast-follow.
- 🟢 **`Agent.RecordSpend` dropped the disabled-guard its eibc template has** — `OnDemandLPRecord.RecordSpend` returns early when rate limiting is disabled, with a comment naming the divide-by-zero and a `require.NotPanics` test; the agent copy calls `SpendBucket` unconditionally. Latent today (the sole caller checks `SpendEnabled` first), but any future caller without that exact pre-check panics — in a Begin/EndBlocker that's a halt. Restore the guard or actually share the helper.
- 🟢 **Invariant nits** — (a) errors are aggregated with `fmt.Errorf("escrow ledger: %v", errs)` on a `[]error`; sibling invariants (sequencer, sponsorship) use `errors.Join`. (b) The check is `ledger ≤ module balance`, but the agent module account is in the blocked-address set, so unattributed surplus shouldn't arise through normal sends — equality would also catch drift in the surplus direction (coins credited without a ledger entry). Worth a comment either way stating whether surplus is tolerated deliberately.
- 🟢 **Event schema split** — `SubmitAttestedAction` still emits a legacy untyped event while transfers and all escrow ops now emit typed events; consumers reconstructing the shared action log must parse two schemas. Fine to defer, but converging on typed events for the action path would finish the job `c730aded` started.
- 🧹 The ledger-debit→bank-send sequence is written out twice (`WithdrawAgentEscrow`, `SubmitAttestedTransfer`) and the owner-auth gate four times across handlers; a `payFromEscrow`/`loadOwnedAgent` pair would make the exact relation the solvency invariant polices structurally single-sourced.

### CI

`build` and `golangci-lint` both fail on `x/eibc/keeper/lp_test.go:165` (`NewDemandOrder` signature) — the pre-existing base breakage the PR body discloses; this PR touches only `x/agent` and its protos. `go test ./x/agent/... -count=1` passes locally at HEAD and `gofumpt` is clean.

### Proof of execution

Verified: reran the x/agent suite at HEAD (passes, including the new cross-domain-replay, payload-bound, and invariant tests), and the live smoke transcript's ledger arithmetic is consistent. The honest "not verified" note on the live TEE path remains appropriate.

---

**VERDICT: APPROVE** — the blocker is fixed and regression-pinned, both fund-accounting findings landed, and every smaller item was addressed or deliberately declined. The chain-ID nonce binding is the one thing I'd land before enclaves ship against this token format.

omritoptix (commented):
## Re-review — CI-failure follow-up (delta since the approved round: local commit `4a13f4a3`)

Scope of this pass: the `/workflow ci failed` report and the one new commit since the APPROVE review. Reviewed at high effort (the delta is a one-line test fix; the full feature was already reviewed twice at max effort with a security pass). No security surface in the delta.

### CI failure — diagnosed stale, fix verified

The failing `build`/`golangci-lint` runs are from **2026-07-07 at `c730aded`**, when main still carried the #2168×#2170 semantic merge conflict (`x/eibc/keeper/lp_test.go` calling `NewDemandOrder` with the old arity). That break was fixed **on main** on 2026-07-10 by `bbdefe20` (#2183) — the identical `nil, nil` one-liner. Since CI checks out the PR merge ref, **a plain re-run against current main passes**; nothing in this PR is at fault.

The branch also carries a local commit `4a13f4a3` making the same one-line fix. Verified correct: `nil` lands in the `feeEscalation` slot matching all 14 other non-escalating call sites, no lp_test assertion depends on escalation, and both sides of the merge made the byte-identical change so it auto-merges cleanly. **Note it is not pushed yet** — GitHub still shows head `c730aded`. Either push it or drop it and merge main / re-run CI; all three converge to the same tree. (Once pushed, the PR body's "This PR does not touch x/eibc" line becomes stale — worth a one-word edit.)

### New-main interaction — verified sound

Main moved under the PR: #2183 binds eibc on-demand LPs to x/agent attested agents (`AgentKeeper` interface, `agentLive` gate in LP matching, `SetAgentKeeper` wiring). Simulated the CI merge locally: clean textual merge, `go build ./...` OK, and `go test ./x/agent/... ./x/eibc/... -count=1` all green — including main's `TestLPAgentBinding`/`TestCreateLPAgentBinding` running against this PR's extended `Agent` (spend policy fields, domain-separated nonces). No semantic conflict: eibc reads only agent existence/policy, which this PR doesn't change the meaning of. (Several finder-agent candidates claimed the PR "removes" #2183 code — refuted: the branch merely predates it; the merge preserves everything.)

### Prior findings — accounting

- All blockers and majors from rounds 1–2 (cross-domain nonce replay, ledger/bank solvency invariant, duplicate genesis agent IDs, payload bound, genesis spend-state validation, events) were resolved in `9b1dc51d`..`c730aded` and re-verified in the previous review; the x/agent suite including their regression tests passes at the current merged tree.
- The APPROVE-round non-blocking recommendations remain open (no new x/agent commits): **chain-ID in the attestation nonce pre-image** (still the one to land before enclaves ship against this token format — free now, breaking later), genesis `action_log` validation, the `RecordSpend` disabled-guard, invariant nits, typed event for the action path, and the `payFromEscrow`/`loadOwnedAgent` dedup. All remain non-blocking fast-follows.

### New findings

None blocking. The only new item is the redundancy note above: `4a13f4a3` duplicates a fix main already has — harmless (identical-change auto-merge), but merging main into the branch would be the cleaner mechanism and also brings the branch's CI base current.

---

**VERDICT: APPROVE** — the CI failure is stale (fixed on main; re-run passes), the one-line delta is correct, and the merge with post-#2183 main builds and tests green. Push `4a13f4a3` (or merge main) and re-run CI to turn the checks green; the chain-ID nonce binding remains the recommended pre-ship follow-up.

omritoptix:
CI (build + golangci-lint) was red on a failure outside this PR's scope: `x/eibc/keeper/lp_test.go:165` does not compile on main — a semantic merge conflict between #2168 (added the `FeeEscalation` param to `NewDemandOrder`) and #2170 (added `lp_test.go`). Verified pre-existing by running `go vet ./x/eibc/keeper/` in a clean worktree at the branch base (`ed58cd34`): same typecheck error.

Fixed at the root in `4a13f4a3` by passing `nil` for the escalation param, matching every other non-escalating call site.

Evidence:
```
$ go test ./x/eibc/... -count=1
ok  github.com/dymensionxyz/dymension/v3/x/eibc         0.495s
ok  github.com/dymensionxyz/dymension/v3/x/eibc/keeper  3.646s
ok  github.com/dymensionxyz/dymension/v3/x/eibc/types   0.155s
$ golangci-lint run ./...   # whole repo, CI config
0 issues.
$ go vet ./...
exit 0
```

omritoptix (commented):
## Re-review — pushed CI fix and specification completeness

Reviewed at max effort because this change authorizes attested movement of pooled escrow funds. I also re-ran the security/data-integrity review over the current head `4a13f4a3`. Current GitHub checks are green (build, lint, CodeQL, buf), and the proof remains concrete: the real-app keeper suite with a fake verifier exercises the payout, ledger/module balance changes, rollback paths, budget rollover, nonce binding, replay, ownership, and genesis round-trip.

### Prior findings — complete accounting

- The original blocker (cross-domain nonce replay), escrow solvency invariant, duplicate-agent genesis validation, full-payload size bound, bucket alignment, missing events, and attest/log refactors remain resolved and regression-covered in `9b1dc51d..c730aded`.
- The later CI finding is resolved: `4a13f4a3` is now pushed, is the same correct `NewDemandOrder(..., nil, nil)` fix already present on current main, and current CI passes.
- The previously recorded non-blocking hardening items remain open and remain non-blocking: chain-ID nonce binding, action-log genesis validation, the disabled guard in `RecordSpend`, invariant formatting/surplus semantics, typed action events, and escrow/owner helper deduplication. No new high/critical security issue was found.

### Blocking — the required transfer CLI is missing

Issue #2181 specifies the new messages under `msg_server.go`, codec, **and CLI**, but the branch only registers commands for `fund-escrow`, `withdraw-escrow`, and `update-spend-policy`. There is no command for `MsgSubmitAttestedTransfer`: `x/agent/client/cli/tx.go` does not register one, `tx_escrow.go` does not implement one, and `AppModule.AutoCLIOptions` defines only query commands. This leaves the feature's primary operation unavailable through the requested `dymd tx agent` surface, while the PR body explicitly lists only the other three commands.

Please add and register a `submit-attested-transfer` command accepting the agent ID, recipient, amount, memo, and token (with a focused CLI construction/argument regression test), consistent with the issue's CLI requirement.

Also refresh the PR body's proof/context: it currently says the full suite fails and “This PR does not touch `x/eibc`,” but the pushed head now includes the eIBC test commit and all current checks pass. The existing behavioral proof is substantive, not hollow; these statements are simply stale.

VERDICT: REQUEST_CHANGES

omritoptix:
/workforce Address the latest Codex review completely: add and register the required submit-attested-transfer CLI command with a focused argument/construction regression test, then refresh the PR proof/context so it reflects the current green checks and eIBC test commit. Preserve all existing resolved security fixes.

omriai1 (commented):
## Re-review — current head 0c80003c

Reviewed at max effort because this PR authorizes attested movement of pooled escrow funds, with a security/data-integrity pass. The new delta is the transfer CLI plus a merge of current main; no new blocking issue was found.

### Prior findings — complete accounting

- RESOLVED: the original cross-type nonce replay blocker remains fixed by disjoint action/transfer domain tags and the bidirectional replay regression.
- RESOLVED: escrow under-collateralization detection, duplicate-agent genesis validation, full stored-payload bounds, bucket alignment, typed escrow/transfer events, and the shared attest/log and policy-promotion refactors remain present and covered.
- RESOLVED: the eIBC NewDemandOrder compatibility/CI finding is incorporated, and the current-main merge preserves agent-policy revocation gating for transfers.
- RESOLVED: the latest blocker, the missing submit-attested-transfer CLI, is fixed in c9f7ec1b. The command is registered, accepts the specified five arguments, constructs every MsgSubmitAttestedTransfer field, and has focused tests for registration, exact arity, construction, and invalid amount.
- STILL OPEN, non-blocking: attestation nonces do not bind chain ID, so a token can replay on a fork/sibling chain with matching policy/state; genesis action-log entries are not validated for agent/key/sequence/hash consistency; RecordSpend relies on callers checking SpendEnabled before bucket division; invariant formatting/surplus semantics, legacy untyped action events, and escrow/owner helper dedup remain cleanup items. These were previously classified as fast-follows; no unresolved high/critical security finding was identified.

### Proof, tests, and docs

The Proof of execution is concrete and current at 0c80003c: it includes focused CLI and revoked-policy test commands, repository-wide test output with exit 0, built CLI help and a failing bad-arity invocation, protobuf regeneration with no diff, and lint with zero issues. The live GCP TEE-token limitation is disclosed honestly. There is no QA Brief, so no journey/screenshot gate applies. This backend/CLI behavior change has committed regression coverage, and x/agent/doc.go was updated; no stale README or module documentation was found.

I could not independently rerun Go tests in this review container because go is not installed (the command exits 127). git diff --check is clean. Current GitHub build, lint, and analysis jobs are pending, not failed; buf-break-check passes.

VERDICT: APPROVE

PR #2187, opened 2026-07-11
feat(agent): on-chain reputation registry grounded in the attested action log

omritoptix:
Closes #2186

## What

Adds the third ERC-8004 registry (reputation) to `x/agent`, alongside the existing identity and validation registries: per-`(agent, client)` feedback records plus an incremental per-agent aggregate.

- **`MsgSubmitFeedback{client, agent_id, score, tag1, tag2, evidence_seq}`**
  - Score is fixed-point with 2 decimals: `0..10000` = `0.00%..100.00%`.
  - `tag1` required, `tag2` optional; both capped by the new `feedback_tag_max_bytes` param (default 32), ASCII control chars rejected.
  - `evidence_seq < agent.action_seq` required — every rating is grounded in an already-attested action, and agents with zero action history cannot be rated.
  - Owner self-rating rejected; agent existence required but active status is not (a deactivated agent can still be rated on its record).
  - The new `feedback_fee` param (default 1 DYM) is charged from the client and burned.
  - Resubmission by the same client overwrites the record and adjusts the aggregate (count unchanged).
- **`MsgRevokeFeedback`** — free; deletes the record, decrements the aggregate, and removes the `Reputation` entry when count reaches 0.
- **Queries** — `AgentReputation` (aggregate + integer-division `average_score`, zero-valued when no feedback exists), `AgentFeedback`, `AgentFeedbacks` (paginated, deterministic key-order prefix iteration).
- **State** — two new collections: `feedback` keyed by `(agent_id, client)` at prefix `0x03`, `reputation` keyed by `agent_id` at prefix `0x04`.
- **Genesis** — exports feedbacks in key order; init rebuilds the `reputation` aggregates from the feedback set instead of importing them; `Validate()` rejects duplicate `(agent_id, client)` pairs and out-of-range scores.
- **CLI** — `tx agent submit-feedback` / `revoke-feedback`, plus AutoCLI commands for the three queries.

No consumer wiring (eIBC, sequencer, discovery) — registry + queries only, per the issue.

## Deviation from the issue spec

[...]

PR #2193, opened 2026-07-25
fix(eibc): honor agent policy revocation

### LINKED ISSUE #2192: x/eibc: honor x/agent policy revocation in agent-bound LP liveness gate (complete the autonomous market-maker kill-switch)
## Problem / opportunity

`x/eibc` lets an on-demand LP be bound to an `x/agent` attested agent (issue #2182), and gates fulfillment on the agent being **live**. The liveness gate is incomplete: it honors owner **deactivation** but ignores governance **policy revocation**, so the protocol's strongest emergency lever cannot stop a known-bad agent from moving money.

Concretely, the fulfillment-time gate at `x/eibc/keeper/lps.go:204` is:

```go
func (s LPs) agentLive(ctx sdk.Context, agentID string) bool {
	if agentID == "" {
		return true
	}
	agent, found := s.agents.GetAgent(ctx, agentID)
	return found && agent.Active
}
```

It checks only `found && agent.Active`. Meanwhile `x/agent`'s governance denylist (`MsgRevokePolicy` → `revokedPolicies`) is a **separate, stronger** signal that never mutates `agent.Active` — by design: `x/agent/doc.go` states "Unrevoking is fully reversible since agent state is never mutated by revocation."

The asymmetry is exact and confirmable in the code:

- `x/agent/keeper/msg_server.go:64` — `SubmitAttestedAction` calls `fingerprintNotRevoked(ctx, agent.Policy)`, so a **revoked-policy agent cannot append new attested actions**.
- `x/eibc/keeper/lps.go:204` — the same revoked-policy agent's on-demand LP fleet **keeps fulfilling demand orders**, because `agentLive` never consults the revocation denylist.

So an agent that governance has declared malicious/compromised — one the chain refuses to let log the very actions it is performing — can still autonomously settle eIBC orders with its capital. The kill-switch that #2182 advertises ("provenance + one-shot fleet kill-switch") is not reachable by governance's revocation path. This is a consistency/safety gap, not merely cosmetic: the whole value of an attested, revocable agent fleet is undermined if the revocation cannot actually stop on-chain activity.

## Why now

Autonomous AI agents acting as on-chain market makers are moving from thesis to production: the x402 agentic-payment standard (Coinbase/Cloudflare) passed to **Linux Foundation governance on 2 April 2026** and had processed ~165M agent transactions / ~$50M volume across ~69k active agents by April 2026, with V2 adding session tokens and multi-chain support. A recurring, load-bearing requirement in the 2026 agent-identity / agent-payments literature is **revocability**: the ability to instantly and verifiably neutralize a compromised autonomous agent. Dymension already built the attestation + denylist machinery (`x/agent`) and the agent-bound market-maker path (`x/eibc` #2182); this issue closes the one seam where the revocation signal fails to reach the money path.

Sources:
- x402 → Linux Foundation governance, adoption metrics, V2 session tokens: ``https://www.allium.so/blog/x402-explained-the-internet-native-payments-standard-for-apis-data-and-agent-commerce``/ and https://metamask.io/news/what-is-x402
- Agent identity verification / revocation as a 2026 requirement: ``https://eco.com/support/en/articles/15192005-agent-identity-verification-how-ai-agents-authenticate-purchases-in-2026``

## Proposed change

Make the eIBC **fulfillment-time** liveness gate honor policy revocation, by delegating the full "operational" predicate to `x/agent` (the module that owns the meaning of `Active` + revoked).

1. **`x/agent/keeper` — new method** (e.g. in `registry.go`):

   ```go
   // IsAgentLive reports whether an agent may act on-chain: it must exist, be
   // Active, and its policy must not be in the governance revocation denylist.
   // Fail-closed: any lookup/fingerprint error returns false.
   func (k Keeper) IsAgentLive(ctx sdk.Context, agentID string) bool {
       agent, found := k.GetAgent(ctx, agentID)
       if !found || !agent.Active {
           return false
       }
       fp, err := types.PolicyFingerprint(agent.Policy)
       if err != nil {
           return false
       }
       revoked, err := k.IsPolicyRevoked(ctx, fp)
       if err != nil {
           return false
       }
       return !revoked
   }
   ```

   This mirrors the fingerprint checked by `SubmitAttestedAction` exactly (`agent.Policy`, not `EffectivePolicy`), so the invariant "an agent that cannot log its actions cannot fulfill eIBC orders" holds by construction.

2. **`x/eibc/types/expected_keepers.go` — extend `AgentKeeper`** with `IsAgentLive(ctx sdk.Context, id string) bool` (keep `GetAgent`, still used for the existence-only bind-time check at `msg_server.go:304`).

3. **`x/eibc/keeper/lps.go` — `agentLive` delegates**:

   ```go
   func (s LPs) agentLive(ctx sdk.Context, agentID string) bool {
       if agentID == "" {
           return true // unbound LP
       }
       return s.agents.IsAgentLive(ctx, agentID)
   }
   ```

   The empty-`agentID` short-circuit (unbound LP) stays in eIBC, since "unbound" is an eIBC concept.

This wires cleanly into the existing flow: `GetOrderCompatibleLPs` (`lps.go:194`) already calls `agentLive` per compatible LP; a revoked agent's LPs are simply excluded from the candidate set, so `FulfillByOnDemandLP` skips them exactly as it already skips deactivated ones. Un-revoking (`MsgUnrevokePolicy`) restores fulfillment with no extra code, matching the reversible-by-design semantics.

- **Scope the fix to the fulfillment-time gate only.** The bind-time check (`msg_server.go:304`) stays existence-only, honoring its explicit comment ("the agent need not be active at creation, only at fulfillment"). Revocation, like deactivation, matters only at fulfillment.
- **Check `agent.Policy` fingerprint** (the base immutable policy), identical to `SubmitAttestedAction` at `msg_server.go:64` — not `EffectivePolicy` — to keep the "can't-log ⇒ can't-fulfill" invariant tight. If the action-path gate is ever changed to `EffectivePolicy`, this method should track it (leave a one-line comment saying so).
- **Fail-closed.** On any error (fingerprint computation, store read) `IsAgentLive` returns `false` and never panics — a kill-switch must default to "off," and this path is reachable from order matching, so it must not halt.
- **Encapsulate the predicate in `x/agent`** (`IsAgentLive`) rather than exposing `IsPolicyRevoked` + policy to eIBC, to keep the `AgentKeeper` interface thin and the semantics owned by `x/agent`.
- **No new params, no proto changes, no state migration.** Pure read-path logic over existing state.

## Acceptance criteria

- `x/agent/keeper.IsAgentLive` exists with the semantics above and unit tests covering: (a) unknown agent → false; (b) inactive agent → false; (c) active + policy not revoked → true; (d) active + policy revoked → false; (e) active, revoked, then un-revoked → true again.
- `x/eibc` `AgentKeeper` interface includes `IsAgentLive`; `agentLive` delegates to it; the empty-id (unbound) short-circuit is preserved.
- New eIBC keeper test (mirroring `escalation_test.go`/`lp_test.go` patterns, using `apptesting` real keepers): register an agent, create an on-demand LP bound to it, confirm the LP fulfills a compatible order; then revoke the agent's policy via the gov authority (`MsgRevokePolicy`) and confirm `GetOrderCompatibleLPs` excludes the LP and `FulfillByOnDemandLP` returns "no compatible lp"; then un-revoke and confirm fulfillment resumes.
- Existing eIBC and agent tests pass; `go test ./x/eibc/... ./x/agent/...` green.
- Deterministic and consensus-safe: no wall-clock, randomness, or map-iteration in the new path; not invoked from Begin/EndBlocker; fail-closed with no panics; transactions remain atomic.

## Out of scope

- Do **not** change the bind-time (LP creation) check — it stays existence-only.
- Do **not** alter `SubmitAttestedAction`, the deactivation path, or `EffectivePolicy`/pending-policy rotation semantics.
- Do **not** touch the authorized/granted eIBC fulfillment path or unbound-LP behavior.
- No new reputation consumption, no funds/custody/marketplace logic, no slashing.

---

omriai1:
## Summary

- add the x/agent-owned `IsAgentLive` predicate for existence, activation, and effective-policy revocation
- resolve matured pending rotations at the current block height, matching attested-action and query semantics
- fail closed on agent, fingerprint, or revocation-store read errors
- delegate agent-bound eIBC LP matching to the full operational predicate while preserving unbound LP behavior
- cover revoke/unrevoke and both policy-rotation directions through keeper and eIBC fulfillment tests
- document the eIBC effect of policy revocation
- synchronize protobuf compatibility contracts with the checked-in agent schema

Closes #2192

### Agent liveness follows revocation, unrevocation, and effective-policy rotation
**How:** `go test ./x/agent/keeper -run 'TestIsAgentLive' -count=1 -v`
**Evidence:**
```text
=== RUN   TestIsAgentLive
--- PASS: TestIsAgentLive (0.00s)
=== RUN   TestIsAgentLive_UsesEffectivePolicy
--- PASS: TestIsAgentLive_UsesEffectivePolicy (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.122s
```

### Revoked agent-bound LPs, including matured revoked rotations, cannot fulfill
**How:** `go test ./x/eibc/keeper -run '^TestKeeperTestSuite/TestLPAgentPolicyRevocation$' -count=1 -v`
**Evidence:**
```text
=== RUN   TestKeeperTestSuite
=== RUN   TestKeeperTestSuite/TestLPAgentPolicyRevocation
--- PASS: TestKeeperTestSuite (0.11s)
    --- PASS: TestKeeperTestSuite/TestLPAgentPolicyRevocation (0.11s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/eibc/keeper	0.244s
```

### Protobuf compatibility contracts cover every generated field
**How:** `go test ./x/agent/types -run 'TestAgentProto(Field|RoundTrip)Compatibility' -count=1 -v`
**Evidence:**
```text
--- PASS: TestAgentProtoFieldCompatibility (0.00s)
    --- PASS: TestAgentProtoFieldCompatibility/Params (0.00s)
    --- PASS: TestAgentProtoFieldCompatibility/EventRegisterAgent (0.00s)
--- PASS: TestAgentProtoRoundTripCompatibility (0.00s)
    --- PASS: TestAgentProtoRoundTripCompatibility/Params (0.00s)
    --- PASS: TestAgentProtoRoundTripCompatibility/EventRegisterAgent (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.082s
```

### Exact CI race-and-coverage command passes
**How:** `go-acc -o coverage.txt ./... -- -v --race`
**Evidence:**
```text
PASS
coverage: 8.4
ok  	github.com/dymensionxyz/dymension/v3/x/streamer/keeper	28.555s	coverage: 8.4
PASS
coverage: 0.7
ok  	github.com/dymensionxyz/dymension/v3/x/streamer/types	1.690s	coverage: 0.7
PASS
coverage: 3.5
ok  	github.com/dymensionxyz/dymension/v3/x/vfc/hooks	2.725s	coverage: 3.5
```

### Changed Go code passes lint and formatting checks
**How:** `golangci-lint run --new-from-rev=HEAD ./x/agent/... ./x/eibc/...` and `git diff --check`
**Evidence:**
```text
0 issues.
```

**Not verified:** nothing; all behaviors above are proven.

### DIFF
--- a/x/agent/doc.go
+++ b/x/agent/doc.go
@@ -2,6 +2,7 @@
 // Agents register with an attestation policy, may rotate it after a timelock,
 // and append actions verified against the policy currently in force. A
 // governance-gated denylist prevents byte-identical policy fingerprints from
-// registering new agents or appending attested actions. Unrevoking is fully
-// reversible since agent state is never mutated by revocation.
+// registering new agents, appending attested actions, or fulfilling eIBC orders
+// through agent-bound on-demand LPs. Unrevoking is fully reversible since agent
+// state is never mutated by revocation.
 package agent
--- a/x/agent/keeper/revoke.go
+++ b/x/agent/keeper/revoke.go
@@ -41,14 +41,14 @@ func (k Keeper) IsPolicyRevoked(ctx sdk.Context, fp string) (bool, error) {
 	return k.revokedPolicies.Has(ctx, fp)
 }
 
-// IsAgentLive reports whether an agent may act on-chain. Keep this check on
-// agent.Policy aligned with SubmitAttestedAction's revocation check.
+// IsAgentLive reports whether an agent may act on-chain. Keep this effective
+// policy check aligned with SubmitAttestedAction and the Agent query.
 func (k Keeper) IsAgentLive(ctx sdk.Context, agentID string) bool {
 	agent, err := k.agents.Get(ctx, agentID)
 	if err != nil || !agent.Active {
 		return false
 	}
-	fp, err := types.PolicyFingerprint(agent.Policy)
+	fp, err := types.PolicyFingerprint(agent.EffectivePolicy(ctx.BlockHeight()))
 	if err != nil {
 		return false
 	}
--- a/x/agent/keeper/revoke.go
+++ b/x/agent/keeper/revoke.go
@@ -41,6 +41,24 @@ func (k Keeper) IsPolicyRevoked(ctx sdk.Context, fp string) (bool, error) {
 	return k.revokedPolicies.Has(ctx, fp)
 }
 
+// IsAgentLive reports whether an agent may act on-chain. Keep this check on
+// agent.Policy aligned with SubmitAttestedAction's revocation check.
+func (k Keeper) IsAgentLive(ctx sdk.Context, agentID string) bool {
+	agent, err := k.agents.Get(ctx, agentID)
+	if err != nil || !agent.Active {
+		return false
+	}
+	fp, err := types.PolicyFingerprint(agent.Policy)
+	if err != nil {
+		return false
+	}
+	revoked, err := k.IsPolicyRevoked(ctx, fp)
+	if err != nil {
+		return false
+	}
+	return !revoked
+}
+
 // AllRevokedPolicies returns all revoked fingerprints in ascending order.
 func (k Keeper) AllRevokedPolicies(ctx sdk.Context) ([]string, error) {
 	var fps []string
--- a/x/agent/keeper/revoke_test.go
+++ b/x/agent/keeper/revoke_test.go
@@ -92,6 +92,32 @@ func TestRevokePolicy_AuthorityGated(t *testing.T) {
 	require.False(t, revoked)
 }
 
+func TestIsAgentLive(t *testing.T) {
+	ctx, k, _ := setup(t)
+	ms := keeper.NewMsgServerImpl(*k)
+
+	require.False(t, k.IsAgentLive(ctx, "unknown"))
+
+	require.NoError(t, k.SetAgent(ctx, types.Agent{
+		Id:     "inactive",
+		Policy: policyA(),
+		Active: false,
+	}))
+	require.False(t, k.IsAgentLive(ctx, "inactive"))
+
+	seedAgentWithPolicy(t, ctx, k, "active", policyA())
+	require.True(t, k.IsAgentLive(ctx, "active"))
+
+	fp := fingerprint(t, policyA())
+	_, err := ms.RevokePolicy(ctx, types.NewMsgRevokePolicy(govAuthority, fp, "bad image"))
+	require.NoError(t, err)
+	require.False(t, k.IsAgentLive(ctx, "active"))
+
+	_, err = ms.UnrevokePolicy(ctx, types.NewMsgUnrevokePolicy(govAuthority, fp))
+	require.NoError(t, err)
+	require.True(t, k.IsAgentLive(ctx, "active"))
+}
+
 func TestRevokePolicy_ValidateBasic(t *testing.T) {
 	ctx, k, _ := setup(t)
 	ms := keeper.NewMsgServerImpl(*k)
--- a/x/agent/keeper/revoke_test.go
+++ b/x/agent/keeper/revoke_test.go
@@ -118,6 +118,37 @@ func TestIsAgentLive(t *testing.T) {
 	require.True(t, k.IsAgentLive(ctx, "active"))
 }
 
+func TestIsAgentLive_UsesEffectivePolicy(t *testing.T) {
+	ctx, k, _ := setup(t)
+	ms := keeper.NewMsgServerImpl(*k)
+	fpA := fingerprint(t, policyA())
+	fpB := fingerprint(t, policyB())
+
+	_, err := ms.RevokePolicy(ctx, types.NewMsgRevokePolicy(govAuthority, fpB, "bad rotation"))
+	require.NoError(t, err)
+	seedAgentWithPolicy(t, ctx, k, "rotating-to-revoked", policyA())
+	agent, _ := k.GetAgent(ctx, "rotating-to-revoked")
+	pending := policyB()
+	agent.PendingPolicy = &pending
+	agent.PendingPolicyHeight = 100
+	require.NoError(t, k.SetAgent(ctx, agent))
+	require.True(t, k.IsAgentLive(ctx.WithBlockHeight(99), agent.Id))
+	require.False(t, k.IsAgentLive(ctx.WithBlockHeight(100), agent.Id))
+
+	_, err = ms.RevokePolicy(ctx, types.NewMsgRevokePolicy(govAuthority, fpA, "bad old policy"))
+	require.NoError(t, err)
+	_, err = ms.UnrevokePolicy(ctx, types.NewMsgUnrevokePolicy(govAuthority, fpB))
+	require.NoError(t, err)
+	seedAgentWithPolicy(t, ctx, k, "rotating-from-revoked", policyA())
+	agent, _ = k.GetAgent(ctx, "rotating-from-revoked")
+	pending = policyB()
+	agent.PendingPolicy = &pending
+	agent.PendingPolicyHeight = 100
+	require.NoError(t, k.SetAgent(ctx, agent))
+	require.False(t, k.IsAgentLive(ctx.WithBlockHeight(99), agent.Id))
+	require.True(t, k.IsAgentLive(ctx.WithBlockHeight(100), agent.Id))
+}
+
 func TestRevokePolicy_ValidateBasic(t *testing.T) {
 	ctx, k, _ := setup(t)
 	ms := keeper.NewMsgServerImpl(*k)
--- a/x/agent/types/proto_layout_test.go
+++ b/x/agent/types/proto_layout_test.go
@@ -37,6 +37,8 @@ func TestAgentProtoFieldCompatibility(t *testing.T) {
 			"MaxActionBytes":            {"varint,1,opt,name=max_action_bytes,json=maxActionBytes,proto3", reflect.TypeOf(uint64(0))},
 			"AgentRegistrationFee":      {"bytes,2,opt,name=agent_registration_fee,json=agentRegistrationFee,proto3", reflect.TypeOf(types.DefaultParams().AgentRegistrationFee)},
 			"PolicyRotationDelayBlocks": {"varint,3,opt,name=policy_rotation_delay_blocks,json=policyRotationDelayBlocks,proto3", reflect.TypeOf(uint64(0))},
+			"FeedbackFee":               {"bytes,4,opt,name=feedback_fee,json=feedbackFee,proto3", reflect.TypeOf(types.DefaultParams().FeedbackFee)},
+			"FeedbackTagMaxBytes":       {"varint,5,opt,name=feedback_tag_max_bytes,json=feedbackTagMaxBytes,proto3", reflect.TypeOf(uint64(0))},
 		}},
 		{types.ActionLogEntry{}, map[string]fieldContract{
 			"AgentId":     {"bytes,1,opt,name=agent_id,json=agentId,proto3", reflect.TypeOf("")},
@@ -47,8 +49,9 @@ func TestAgentProtoFieldCompatibility(t *testing.T) {
 			"Time":        {"bytes,6,opt,name=time,proto3,stdtime", reflect.TypeOf(time.Time{})},
 		}},
 		{types.EventRegisterAgent{}, map[string]fieldContract{
-			"AgentId": {"bytes,1,opt,name=agent_id,json=agentId,proto3", reflect.TypeOf("")},
-			"Owner":   {"bytes,2,opt,name=owner,proto3", reflect.TypeOf("")},
+			"AgentId":     {"bytes,1,opt,name=agent_id,json=agentId,proto3", reflect.TypeOf("")},
+			"Owner":       {"bytes,2,opt,name=owner,proto3", reflect.TypeOf("")},
+			"Fingerprint": {"bytes,3,opt,name=fingerprint,proto3", reflect.TypeOf("")},
 		}},
 		{types.EventDeactivateAgent{}, map[string]fieldContract{
 			"AgentId": {"bytes,1,opt,name=agent_id,json=agentId,proto3", reflect.TypeOf("")},
@@ -91,14 +94,17 @@ func TestAgentProtoRoundTripCompatibility(t *testing.T) {
 		}, &types.Agent{}},
 		{&types.Params{
 			MaxActionBytes: 1024, AgentRegistrationFee: sdk.NewInt64Coin("adym", 12),
-			PolicyRotationDelayBlocks: 99,
+			PolicyRotationDelayBlocks: 99, FeedbackFee: sdk.NewInt64Coin("adym", 34),
+			FeedbackTagMaxBytes: 56,
 		}, &types.Params{}},
 		{&types.ActionLogEntry{
 			AgentId: "agent-1", Seq: 7, Payload: []byte("payload"),
 			PayloadHash: []byte("hash"), Height: 42,
 			Time: time.Date(2026, time.July, 14, 12, 34, 56, 789, time.UTC),
 		}, &types.ActionLogEntry{}},
-		{&types.EventRegisterAgent{AgentId: "agent-1", Owner: "dym1owner"}, &types.EventRegisterAgent{}},
+		{&types.EventRegisterAgent{
+			AgentId: "agent-1", Owner: "dym1owner", Fingerprint: "fingerprint",
+		}, &types.EventRegisterAgent{}},
 		{&types.EventDeactivateAgent{AgentId: "agent-1", Owner: "dym1owner"}, &types.EventDeactivateAgent{}},
 		{&types.EventUpdateAgentPolicy{AgentId: "agent-1", ActivationHeight: 42}, &types.EventUpdateAgentPolicy{}},
 	}
--- a/x/eibc/keeper/lp_test.go
+++ b/x/eibc/keeper/lp_test.go
@@ -378,6 +378,30 @@ func (suite *KeeperTestSuite) TestLPAgentPolicyRevocation() {
 	suite.Require().NoError(err)
 	suite.Require().Len(compat, 1)
 	suite.Require().NoError(k.FulfillByOnDemandLP(suite.Ctx, id2, 0))
+
+	agent, found := suite.App.AgentKeeper.GetAgent(suite.Ctx, "agent1")
+	suite.Require().True(found)
+	revokedPendingPolicy := policy
+	revokedPendingPolicy.PolicyValues = `{"measurement":"revoked-rotation"}`
+	agent.PendingPolicy = &revokedPendingPolicy
+	agent.PendingPolicyHeight = suite.Ctx.BlockHeight() + 1
+	suite.Require().NoError(suite.App.AgentKeeper.SetAgent(suite.Ctx, agent))
+	pendingFP, err := agenttypes.PolicyFingerprint(revokedPendingPolicy)
+	suite.Require().NoError(err)
+	_, err = agentMsgServer.RevokePolicy(suite.Ctx, agenttypes.NewMsgRevokePolicy(
+		govAuthority, pendingFP, "bad rotation",
+	))
+	suite.Require().NoError(err)
+
+	suite.Ctx = suite.Ctx.WithBlockHeight(agent.PendingPolicyHeight)
+	id3 := suite.orderWithSeq(3, orderAddr.String(), math.NewInt(40), math.NewInt(10))
+	order, err = k.GetDemandOrder(suite.Ctx, commontypes.Status_PENDING, id3)
+	suite.Require().NoError(err)
+	compat, err = k.LPs.GetOrderCompatibleLPs(suite.Ctx, *order)
+	suite.Require().NoError(err)
+	suite.Require().Empty(compat)
+	err = k.FulfillByOnDemandLP(suite.Ctx, id3, 0)
+	suite.Require().True(errorsmod.IsOf(err, gerrc.ErrNotFound), "expected no compatible lp, got %v", err)
 }
 
 // CreateOnDemandLP requires a referenced agent to exist, but not to be active.
--- a/x/eibc/keeper/lp_test.go
+++ b/x/eibc/keeper/lp_test.go
@@ -4,9 +4,12 @@ import (
 	errorsmod "cosmossdk.io/errors"
 	"cosmossdk.io/math"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	"github.com/dymensionxyz/dymension/v3/app/apptesting"
 	agentkeeper "github.com/dymensionxyz/dymension/v3/x/agent/keeper"
 	agenttypes "github.com/dymensionxyz/dymension/v3/x/agent/types"
+	"github.com/dymensionxyz/dymension/v3/x/common/tee"
 	commontypes "github.com/dymensionxyz/dymension/v3/x/common/types"
 	"github.com/dymensionxyz/dymension/v3/x/eibc/types"
 	"github.com/dymensionxyz/gerr-cosmos/gerrc"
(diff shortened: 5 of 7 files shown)


omriai1 (commented):
## Request changes

The direction is right — encapsulating the predicate in `x/agent` and keeping the unbound-LP short-circuit in `x/eibc` is the correct split, and the eIBC wiring is clean. But the predicate reads the wrong policy, and the result is that the kill-switch this PR exists to complete still does not stop the money path.

### Blocking: the kill-switch is still bypassable, via policy rotation

`IsAgentLive` fingerprints the **stored** `agent.Policy`. Every other place the denylist is enforced resolves the **effective** policy:

- `SubmitAttestedAction` promotes a matured `PendingPolicy` into `agent.Policy` *before* running `fingerprintNotRevoked` — so its check is on the post-promotion (effective) policy.
- The gRPC `Agent` query fingerprints `agent.EffectivePolicy(ctx.BlockHeight())`, explicitly so the reported `revoked` status matches what submit-time enforcement would apply right now.

`IsAgentLive` is now the only reader that disagrees. So the invariant this PR is built on — "an agent that cannot log its actions cannot fulfill eIBC orders" — does not hold.

The scenario is the emergency case, not an exotic one. An owner schedules a rotation; governance later revokes the image being rotated into (the timelock backstop that `UpdateAgentPolicy` explicitly defers to). Once the rotation matures:

- `SubmitAttestedAction` refuses every action — the agent cannot log anything.
- The `Agent` query reports `revoked: true`, so operators and dashboards believe the kill-switch worked.
- The agent's on-demand LP fleet keeps matching and fulfilling demand orders, moving real funds.

This does not self-heal. Promotion of `PendingPolicy` into storage happens **only** as a side effect of a *successful* `SubmitAttestedAction`, and `x/agent` has no BeginBlocker or EndBlocker. Since the revocation blocks that message before it persists anything, `agent.Policy` can never be promoted while the effective policy is revoked. The bypass window is permanent and entirely attacker-controlled — a compromised agent simply stops submitting actions.

Verified end-to-end against this branch, driving the real `RegisterAgent` / `UpdateAgentPolicy` / `RevokePolicy` path and the real eIBC keeper:

```text
SubmitAttestedAction blocked as expected: policy revoked: 4044d0bd...: unmet precondition
compatible LPs after kill-switch = 1   (want 0)
FulfillByOnDemandLP err = <nil>        (want not-found)
--- FAIL: KILL-SWITCH BYPASS: revoked agent's LP still matched
```

The order was fulfilled and funds moved for an agent the chain considers revoked.

**The same staleness also fails closed**, which is worth fixing for its own sake: an agent whose *old* policy is revoked and which has matured a clean rotation is allowed to act by `x/agent` and reported clean by the query, yet `IsAgentLive` still sees the revoked stored policy and freezes its entire LP fleet — with no operator-visible cause. `x/agent/doc.go` promises unrevoking is "fully reversible ... with zero side effects"; here, unrevoking cannot restore the fleet.

Both directions are fixed by one line, using the accessor the codebase already has:

```go
fp, err := types.PolicyFingerprint(agent.EffectivePolicy(ctx.BlockHeight()))
```

Confirmed: with that applied, both divergences resolve, and `./x/agent/... ./x/eibc/...` pass (excluding the pre-existing failure noted below).

One behavioural consequence worth a comment when you make the change: the predicate becomes height-dependent, so an LP's liveness can flip at the activation height with no transaction in between. That is deterministic and consensus-safe, but non-obvious.

### The comment documents the bug as if it were the intent

```go
// Keep this check on agent.Policy aligned with SubmitAttestedAction's revocation check.
```

Those two are precisely *not* aligned — `SubmitAttestedAction` checks the post-promotion policy. As written, this comment will convince the next reader that the divergence was considered and is correct. It needs to be removed or rewritten to say why the effective policy is the right basis.

The linked issue asserted the same thing ("identical to `SubmitAttestedAction` at `msg_server.go:64`"), which is where this came from — but line 64 sits *after* the promotion at lines 53-57. The issue's premise was wrong; it shouldn't be carried into the code.

### Test gap that let this through

Neither new test constructs an agent with a `PendingPolicy`, so both pass against the buggy predicate — stored and effective policy coincide in every state they exercise. Please add coverage for:

- matured rotation into a **revoked** policy (the security case) — assert at the **eIBC** level, since that is where funds move;
- matured rotation into a **clean** policy while the old one is revoked (the freeze case);
- rotation scheduled but **not yet matured**.

Two smaller gaps: the revocation test only exercises a single LP, so it never asserts the co-existence property `TestLPAgentBinding` does assert for deactivation (revoked bound LP skipped *while an unbound LP still fulfills*); and an LP bound to a non-existent agent id is untested.

### Minor

- `IsAgentLive` uses `k.agents.Get` while neighbouring code uses `GetAgent`. This is the *better* choice here — `GetAgent` panics on non-`ErrNotFound` store errors, and this runs inside the order-matching loop — but it's the one genuinely non-obvious decision in the function and currently reads as inconsistency. Worth a line saying why.
- `x/agent/doc.go` enumerates what the denylist prevents ("registering new agents or appending attested actions"). Once this lands, it also disables agent-bound LP fulfillment; consider extending that sentence.

### On the proof of execution

The evidence is concrete and I reproduced both commands independently — they pass as claimed. The limitation is that they demonstrate the cases the tests cover, and the tests miss the state where the feature fails.

Your "Not verified" note is honest and I confirmed `TestAgentProtoFieldCompatibility` fails identically on `main`, so it is genuinely pre-existing and unrelated to this diff. The direction is stated backwards, though: the test *expects* 5 `Params` and 3 `EventRegisterAgent` fields while the generated types have 3 and 2 — not the other way around.

omriai1 (commented):
## Re-review

The blocking issue from the last round is genuinely fixed, and fixed properly. One explicitly-requested item was dropped, which is what holds this.

### Resolved

**The kill-switch bypass (blocker).** `IsAgentLive` now fingerprints `agent.EffectivePolicy(ctx.BlockHeight())`, so it agrees with `SubmitAttestedAction`'s post-promotion check and the `Agent` query in every state. Both directions close: a matured revoked rotation can no longer fulfill, and an agent that matures a clean rotation out of a revoked policy is no longer frozen. The stale comment that documented the divergence as intent is gone and now states the real invariant.

I did not take the tests on faith — I reverted the one-line fix and re-ran them:

```text
--- FAIL: TestIsAgentLive_UsesEffectivePolicy
    revoke_test.go:136: Should be false

--- FAIL: TestKeeperTestSuite/TestLPAgentPolicyRevocation
    lp_test.go:402: Should be empty, but was [{0 80 ... agent_id:"agent1" ...}]
```

Both fail against the pre-fix predicate and pass against the fix. The eIBC one fails at the money path — the revoked agent's LP appearing in the compatible set — which is exactly where the assertion needed to be. These are real regression tests, not tests written to pass.

**Test gaps.** The three requested rotation states are covered: not-yet-matured (height 99 vs 100), matured-into-revoked (asserted at the eIBC level, correctly), and matured-clean-out-of-revoked.

Two follow-ups I asked for that I'm now withdrawing rather than leaving open:

- *LP bound to a non-existent agent id* — moot. `x/agent` has no agent-deletion path and `CreateOnDemandLP` rejects unknown ids, so the state is unreachable. `IsAgentLive` fails closed on it regardless. Not worth a test.
- *Co-existence (revoked bound LP skipped while an unbound LP still fulfills)* — worth having, but deactivation and revocation now share one seam (`agentLive` → `IsAgentLive`), and `TestLPAgentBinding` already asserts the property across that seam. Structurally covered. Nit, not a gap.

Verified independently: `./x/eibc/... ./x/agent/keeper` all pass. `TestAgentProtoFieldCompatibility` fails identically on a clean checkout of `main` at `ab4d98a` — genuinely pre-existing, correctly disclosed.

### Blocking: `x/agent/doc.go` no longer describes what revocation does

```go
// A governance-gated denylist prevents byte-identical policy fingerprints from
// registering new agents or appending attested actions.
```

That enumeration is the package's own statement of the kill-switch's blast radius, and after this PR it is incomplete in the direction that matters: revoking a fingerprint now also freezes every agent-bound on-demand LP's ability to fulfill demand orders. An operator reading `x/agent` to decide whether to pull the switch in an incident would conclude it only stops registration and action-logging, and would not learn that it halts a fulfillment fleet moving funds.

This was raised last round and landed without a change or a note. It's one sentence.

Worth doing at the same time, both previously raised and both still open — neither blocking on its own:

- `IsAgentLive` uses `k.agents.Get` rather than the neighbouring `GetAgent`. Still the right call (`GetAgent` panics on non-`ErrNotFound`, and this runs inside the order-matching loop), still the one non-obvious decision in the function, still unexplained.
- The predicate is now height-dependent, so an LP's liveness can flip at the activation height with no transaction in between. Deterministic and consensus-safe, but surprising enough to name.

### On the security surface

I re-ran a security pass over the diff. Three things surfaced; I want to be explicit that **none of them is introduced, widened, or regressed by this PR**, so none blocks it:

- An operator can sidestep the switch by creating a fresh *unbound* LP from the same funds address. This is pre-existing — `agentLive` on `main` already short-circuits unbound LPs, LP creation is unrestricted, and the merged deactivation switch (#2183) is escapable the identical way. It reflects the binding's actual shape: the LP funds address is an ordinary key and the agent id is opt-in metadata, so revocation withdraws an attestation, not custody.
- A revoked agent can rotate to a fresh unrevoked policy and recover after the timelock. `UpdateAgentPolicy` already gates scheduling on the denylist and its comment states the design — governance denylists known-bad images; recovery via a genuinely new image is what the timelock is for.
- LP creation binds to already-revoked agents. Intentional and commented ("existence only: the agent need not be active at creation, only at fulfillment"); inert while revoked.

I flag them because the doc sentence above is where a reader forms their model of how much revocation actually buys — which is why getting that sentence right matters more than it looks.

### Proof of execution

Concrete and reproducible; I ran both commands and they behave as claimed. The mutation check above is the evidence the last round was missing — the tests now demonstrably discriminate.

The "Not verified" note still has the `TestAgentProtoFieldCompatibility` direction reversed: the test *expects* 5 `Params` and 3 `EventRegisterAgent` fields, and the generated types have 3 and 2.

---

Update `doc.go` and this is good to go. The substance of the fix is right and the tests now prove it.

omriai1 (commented):
## Re-review

The round-2 blocker is closed. `doc.go` now states the full blast radius, and I re-verified the round-1 fix rather than carrying it forward on the previous review's word. Approving.

### Resolved

**`x/agent/doc.go` blast radius (round-2 blocker).** The enumeration now reads "registering new agents, appending attested actions, or fulfilling eIBC orders through agent-bound on-demand LPs." That is the complete set — it matches the three `fingerprintNotRevoked` call sites plus the new eIBC seam. An operator reading `x/agent` before pulling the switch now learns it halts a fulfillment fleet.

The trailing claim — "unrevoking is fully reversible since agent state is never mutated by revocation" — now spans a module boundary, so I checked whether it survives the widening. It does: revocation writes only the denylist, the LP record is untouched, and compatibility is recomputed per order rather than persisted as a flag. The sentence is earned, not just extended.

**Round-1 blocker (kill-switch bypass via policy rotation).** Re-verified independently rather than taken on the last round's word. Reverting the predicate to fingerprint the stored `agent.Policy`:

```text
--- FAIL: TestIsAgentLive_UsesEffectivePolicy               revoke_test.go:136
--- FAIL: TestKeeperTestSuite/TestLPAgentPolicyRevocation   lp_test.go:402
```

Both fail pre-fix and pass post-fix, and the eIBC failure lands at the money path. I also confirmed the `FulfillByOnDemandLP` → `ErrNotFound` assertions carry weight independently of the `GetOrderCompatibleLPs` pre-assertions, so the test is not resting on a diagnostic.

Everything else from rounds 1 and 2 — the stale "aligned with `SubmitAttestedAction`" comment, the three rotation states, the two withdrawn follow-ups — stays resolved. `x/agent/keeper` and `x/eibc/...` pass; `gofumpt` clean; `golangci-lint` reports 0 issues.

### Security

No high or critical findings, and nothing introduced or widened here.

Two things I checked that the previous rounds did not, both clean:

- **Fulfillment paths that skip the seam.** `FulfillOrder` and the authz-granted `FulfillOrderAuthorized` do not consult agent state — correctly, since neither reads the LP registry. `FulfillByOnDemandLP` is reachable only from `TryFulfillOnDemand` and sources its candidates exclusively from `GetOrderCompatibleLPs`, so `agentLive` → `IsAgentLive` is a genuine single choke point rather than one gate among several. There are no eIBC BeginBlocker/EndBlocker fulfillment paths.
- **Whether the error swallow is fail-closed by construction or by luck.** Both collections are keyed by terminal `StringKey`, whose `Encode` cannot fail, and real store I/O faults panic rather than returning an error. The reachable error set is therefore "agent absent" and "stored bytes fail to decode" — `false` is the conservative answer to both, and there is no input by which an attacker induces an error to their benefit. The worst an attacker-chosen id achieves is excluding their own LP.

The three pre-existing items named last round (fresh unbound LP from the same funds address; rotation to a clean image after the timelock; creation binding to an already-revoked agent) are unchanged and remain out of scope.

### Still open, none blocking

Two of these were explicitly non-blocking last round and I am not escalating them; the third is new and small.

- **`k.agents.Get` vs `GetAgent` is still unexplained** (third round). Worth settling, because both readings are defensible and the code does not say which is meant: `GetAgent` panics on a decode error, which is the module's convention for state corruption; `Get` swallows it, which is safer inside the order-matching loop. Since the swallow is provably fail-closed here, either is fine — but a maintainer should not have to derive that. One clause, or switch to `GetAgent` and reuse `fingerprintNotRevoked`, which would also drop the duplicated fingerprint-then-revoked sequence.
- **The predicate is height-dependent**, so an LP's liveness can flip at the activation height with no transaction in between. Deterministic and consensus-safe; still unnamed.
- **`agentLive`'s comment lost the useful half.** Trimming it to "Unbound LPs (empty id) are always live" documents the trivial branch and drops the fleet-kill-switch framing the deleted clause carried — right as this PR *widens* that semantics from deactivation to governance revocation. That product knowledge was worth keeping.

### Proof of execution

Concrete, and I reproduced both commands verbatim — output matches the PR body exactly. The mutation check above is what makes them meaningful: the tests demonstrably discriminate rather than merely pass.

The "Not verified" note still has the direction reversed, third round running: the test *expects* 5 `Params` and 3 `EventRegisterAgent` fields, and the generated types have 3 and 2. The disclosure itself is sound — `x/agent/types/proto_layout_test.go` is byte-identical to `main` and fails there the same way, so it is genuinely pre-existing and unrelated to this diff.

---

Good to go. The fix is right, the tests prove it, and the docs now describe what the switch actually does.

omriai1 (commented):
## Re-review

The post-approval commit is a legitimate improvement, not a patch-over. Both blockers from rounds 1 and 2 remain closed. Approving.

### The new commit (`3df0935a`) checks out

`test(agent): sync proto compatibility contracts` touches one file and only adds contract entries: `Params.FeedbackFee` (field 4), `Params.FeedbackTagMaxBytes` (field 5), and `EventRegisterAgent.Fingerprint` (field 3).

The thing worth confirming for a test-only commit that turns a red build green is the *direction* — whether assertions were loosened to fit the code. They were not, and the test's own structure makes that checkable: it asserts `require.Len(t, test.fields, typ.NumField())`, so the contract map must exactly equal the generated struct's field count. Adding entries is the only way to satisfy it; deleting them would fail just as loudly. The commit restores the contract to the generated types.

I verified the three tags match the generated structs character-for-character, and that the PR changes no `.proto` and no `.pb.go` — `git diff --stat main...HEAD -- x/agent/types/ proto/` returns only the test file. So there is no hand-edited-generated-code risk. The fields landed in #2187 and #2180 without the contract test following; this is that debt being paid.

This also settles the "Not verified" note that ran through three rounds. The note was stale in the opposite direction from how it was flagged: on `main` the contract map holds 3 `Params` and 2 `EventRegisterAgent` entries while the generated types have 5 and 3. Baseline failure on a clean `main` worktree (line 66 pre-fix):

```text
proto_layout_test.go:66: "map[AgentRegistrationFee:... MaxActionBytes:... PolicyRotationDelayBlocks:...]"
    should have 5 item(s), but has 3
proto_layout_test.go:66: "map[AgentId:... Owner:...]" should have 3 item(s), but has 2
```

The test lagged the code. With the commit, `./x/agent/types` passes in full, and "Not verified: nothing" is now accurate.

### Prior findings — all accounted for

**Resolved.** Reverting the predicate to fingerprint the stored `agent.Policy`:

```text
--- FAIL: TestIsAgentLive_UsesEffectivePolicy               revoke_test.go:136
--- FAIL: TestKeeperTestSuite/TestLPAgentPolicyRevocation   lp_test.go:402
```

Both fail pre-fix, both pass post-fix, and the eIBC failure lands on the money path — the revoked agent's LP reappearing in the compatible set. These discriminate rather than merely pass.

The round-2 blocker is closed: `doc.go` enumerates all three effects, including eIBC fulfillment through agent-bound LPs. The three rotation states and the stale "aligned with `SubmitAttestedAction`" comment stay resolved.

**Still open, none blocking, unchanged from round 3:**

- `IsAgentLive` uses `k.agents.Get` rather than the module's `GetAgent`. Reframing this from previous rounds: the choice is *correct* and slightly undersold. `GetAgent` panics on non-`ErrNotFound`, so this PR removes a panic-capable call from the order-matching loop. That is a real hardening, and it deserves the one clause that would say so — a docs nit, not a defect.
- The predicate is height-dependent, so an LP's liveness can flip at a rotation activation height with no transaction in between. Deterministic and consensus-safe; still unnamed. Mitigating context: `GetOrderCompatibleLPs` was already height-dependent via `Accepts`, so this is not a new class of behavior.
- `agentLive`'s comment still documents only the trivial unbound branch, having dropped the fleet-kill-switch framing right as that semantics widened.

One addition in the same family: reuse `fingerprintNotRevoked`, already in this file and what the other three revocation call sites use. `_, err := k.fingerprintNotRevoked(ctx, agent.EffectivePolicy(ctx.BlockHeight())); return err == nil` is behavior-preserving — I ran it green — and makes the alignment the comment *asks* for structural rather than a standing manual obligation. Non-blocking, but it retires the first bullet above at the same time.

Separately, worth a follow-up rather than this PR: `proto_layout_test.go` pins 6 of the 8 messages in `agent.pb.go` — `EventPolicyRevoked` and `EventPolicyUnrevoked` are uncovered, as are all messages in the other six generated files. The suite guards a curated subset, so the exact drift this commit repairs can recur silently. Deriving the message list from the file descriptor would make new messages fail closed.

### Security

No high or critical findings, and nothing introduced or widened.

The choke point holds: `GetOrderCompatibleLPs` and `FulfillByOnDemandLP` each have exactly one non-test caller, x/eibc registers no BeginBlocker or EndBlocker, and `FulfillOrder` / `FulfillOrderAuthorized` reach `fulfillBasic` without touching the LP store — correctly, since those spend the fulfiller's own funds under their own signature and carry no agent binding. `agentLive` is a genuine single gate on the automated fleet.

The error swallow is fail-closed by construction: every reachable error means "liveness cannot be established," and no attacker-chosen input turns an error into `true` — the worst it achieves is excluding the attacker's own LP. Determinism holds: `EffectivePolicy` is a pure height comparison over consensus state, with no map iteration, clock, or randomness on the new path.

The three pre-existing items from round 2 (fresh unbound LP from the same funds address; rotation to a clean image after the timelock; creation binding to an already-revoked agent) are unchanged and remain out of scope.

### Proof of execution

All three test commands reproduce verbatim against the PR body. `-race` clean across `./x/agent/... ./x/eibc/...`, `golangci-lint` v2.1.6 reports 0 issues, `gofumpt -l` and `go vet` clean.

One scoping note on the evidence, not a defect: the CI-command section shows a tail of three unrelated packages, which is a plausible excerpt of a long `go-acc` sweep but does not by itself establish that all of `./...` is green. I verified the race and lint claims on the touched packages rather than re-running the full sweep.

PR #2195, opened 2026-08-01
fix(eibc): derive on-demand LP shuffle seed

### LINKED ISSUE #2194: x/eibc: chain-derive the on-demand LP fulfillment shuffle seed (caller-supplied rng lets a fulfiller front-run/starve competing autonomous MMs)
## Problem / opportunity

The eIBC **on-demand LP** subsystem (`x/eibc`) is the on-chain home of autonomous market makers: an MM funds an address, registers an `OnDemandLP`, and the chain then auto-fulfills matching demand orders from that address. When several LPs are all compatible with one order, the chain must pick **one** winner, and it does so by a *shuffle* of the compatible set:

`x/eibc/keeper/lps.go` `FulfillByOnDemandLP`:
```go
lps, err := k.LPs.GetOrderCompatibleLPs(ctx, *o) // deterministic, id-ascending
o.ApplyEffectiveFee(uint64(ctx.BlockHeight()))
r := rand.New(rand.NewPCG(rng, 0))              // rng is a caller-supplied uint64
r.Shuffle(len(lps), func(i, j int) { lps[i], lps[j] = lps[j], lps[i] })
for _, lp := range lps { /* first one that can pay wins, then return */ }
```

The shuffle seed is **entirely attacker-controlled**. It comes straight from the transaction:

`x/eibc/keeper/msg_server.go` `TryFulfillOnDemand`:
```go
shuffleSeed := uint64(msg.Rng)
err = m.FulfillByOnDemandLP(ctx, msg.OrderId, shuffleSeed)
```

`MsgTryFulfillOnDemand` (`proto/.../eibc/tx.proto`) is **permissionless** — `option (cosmos.msg.v1.signer) = "signer"`, any account — and `rng` is an unconstrained `int64` ("rng to choose fulfiller from eligible randomly"). `ValidateBasic` (`x/eibc/types/tx.go:173`) checks only that the signer is a valid address and `order_id` is non-empty; `rng` is never constrained. The CLI (`x/eibc/client/cli/tx.go:208`) passes it as a plain positional integer that **defaults to `0`**.

Two concrete failures follow, and both defeat the "choose fulfiller randomly" fairness property the code is trying to provide:

1. **Grindable selection (front-running).** Any actor who owns — or colludes with — one of the compatible LPs can compute the PCG shuffle offline for candidate seeds until it finds a `rng` that places **their** LP first in the shuffled slice, then submit `MsgTryFulfillOnDemand{order_id, rng}`. Because the first compatible LP that can pay wins and the function returns, that actor wins **every** order its LP is compatible with, front-running and **starving all competing autonomous MMs**. The "randomness" is chosen by the beneficiary.
2. **Constant permutation on the default seed.** With the CLI default `rng = 0`, `rand.NewPCG(0, 0)` is a *fixed* generator, so the shuffle is a **constant permutation** of the id-ascending compatible set. For a stable set of compatible LPs the **same LP wins every time**. Honest relayers who don't bother to vary the seed systematically favor whichever LP the fixed permutation happens to front-load.

The selection of *which* MM captures a fee is a value-bearing, consensus-executed decision. Sourcing its entropy from the tx submitter turns a fairness mechanism into a submitter-chosen outcome. This is a genuine gap: there is no existing field, param, or code path that makes fill allocation ungameable — the recent eIBC autonomous-MM issues (#2167 Dutch-auction fee escalation, #2169 velocity caps, #2182 agent binding, #2189 min-fee floor, #2192 revocation gate) are all orthogonal and none touches selection-seed integrity.

## Why now

2026's crypto-AI flow is autonomous market makers and intent solvers competing to fill orders. Fair, **ungameable** allocation among competing solvers is the property these networks are built on — "sophisticated AI solvers compete ruthlessly … intent-solver systems accounted for $4.1B in cross-chain volume over a recent 90-day period," and the x402 micropayment wave (~165M transactions / ~69k active agents by April 2026) multiplies the number of tiny orders many agent-LPs contend for. A selection seed the fulfiller picks means the biggest griefer, not the market, decides who earns; that directly undermines the multi-agent MM market eIBC is positioning for.

- Intent-based solver competition / AI agents as market makers: https://1inch.com/blog/post/how-ai-agents-execute-on-chain-trades · https://dev.to/barissozen/why-intent-based-trading-is-the-natural-protocol-shape-for-ai-agents-2a4f
- x402 agent-payment micropayment volume: ``https://www.allium.so/blog/x402-explained-the-internet-native-payments-standard-for-apis-data-and-agent-commerce``/
- Liquidity-exhaustion / adversarial ordering in intent-based cross-chain systems: https://arxiv.org/pdf/2602.17805

## Proposed change

Replace the caller-supplied shuffle seed with a **chain-derived deterministic seed** that neither the fulfiller (at submit time) nor the order creator (at order-creation time) can control, while keeping the existing shuffle algorithm and everything downstream unchanged.

1. **New seed helper** in `x/eibc/keeper/lps.go` (unexported):
   ```go
   // deterministicFulfillSeed derives the LP-shuffle seed from data the tx
   // submitter cannot choose: the order id (per-order variation) mixed with the
   // current block hash (per-block entropy neither the fulfiller nor the order
   // creator controls at their respective action times). Consensus-safe: the
   // header hash is consensus data identical across validators, and sha256 with
   // fixed byte order is deterministic.
   func deterministicFulfillSeed(ctx sdk.Context, orderID string) uint64 {
       h := sha256.New()
       h.Write([]byte(orderID))
       h.Write(ctx.HeaderHash()) // []byte; empty in some init/test contexts — still deterministic
       sum := h.Sum(nil)
       return binary.BigEndian.Uint64(sum[:8])
   }
   ```
   (`sha256` is already used deterministically in this package — `x/eibc/types/demand_order.go:189`.)

2. **Wire it in** `x/eibc/keeper/msg_server.go` `TryFulfillOnDemand`: stop reading `msg.Rng`; pass the derived seed:
   ```go
   seed := deterministicFulfillSeed(ctx, msg.OrderId)
   err = m.FulfillByOnDemandLP(ctx, msg.OrderId, seed)
   ```
   `FulfillByOnDemandLP(ctx, order, seed uint64)` keeps its current signature and body; only the seed source changes. The pre-shuffle compatible set from `GetOrderCompatibleLPs` is already deterministic (KeySet range iteration in `rollapp,denom,id` ascending order), so `(orderID, headerHash)` fully determines the winner.

3. **Deprecate the `rng` field** without a wire-breaking removal. In `proto/.../eibc/tx.proto` keep `int64 rng = 3;` (field number preserved for backward compatibility) and change its comment to `// Deprecated: ignored. The fulfiller is chosen by a chain-derived deterministic seed.`; run `make proto-gen`. Update the CLI (`x/eibc/client/cli/tx.go` `NewCmdTryFulfillOnDemand`): keep accepting the optional positional `[rng]` arg so existing scripts don't error, but set `Rng: 0` unconditionally, and change the `Long` help to state the argument is ignored.

## Decisions (all made — implement exactly this)

- **Entropy inputs:** `orderID` bytes concatenated with `ctx.HeaderHash()`, in that order. Order id gives per-order variation within a block; header hash gives per-block entropy that is unknown/uncontrollable to the fulfiller when they build the tx and to the order creator when the order was created (a prior block). This closes both the fulfiller-grinding and the (more exotic) order-id-crafting vector.
- **Hash + reduction:** `sha256`, take the first 8 bytes big-endian as the `uint64` seed. Matches the existing `sha256` usage in the module; no new deps.
- **Empty header hash:** acceptable and handled — if `ctx.HeaderHash()` returns empty (some init/genesis/test contexts), the seed derives from the order id alone. Still fully deterministic; **must not panic**.
- **SDK accessor:** use `ctx.HeaderHash()` (returns `[]byte`). If this fork's `sdk.Context` does not expose it, use `ctx.HeaderInfo().Hash` — both are consensus-deterministic and equivalent for this purpose.
- **Keep the shuffle algorithm** (`rand.NewPCG(seed, 0)` + `r.Shuffle`) exactly as-is; only the seed value changes.
- **Do not remove** the `rng` proto field (avoids a breaking wire change); ignore its value.
- **Module label:** this is a self-contained `x/eibc` change; no `x/agent` or proto message-shape changes beyond the deprecation comment.

## Acceptance criteria

- New unit tests in `x/eibc/keeper/lps_test.go`:
  - **Seed independence from `msg.Rng`:** with ≥2 compatible, fundable LPs for one order at a fixed block height, calling the fulfill path with two *different* `msg.Rng` values selects the **same** LP (verify via the `EventMatchedOnDemandLP` / resulting `Spent` bookkeeping). The winner must not depend on `msg.Rng`.
  - **Determinism:** same `(orderID, headerHash)` → same winner across repeated runs; distinct order ids over a stable compatible set produce a varied (not constant) winner distribution — i.e. the default-`0` constant-permutation behavior is gone.
  - **Empty header hash:** running with an empty `ctx.HeaderHash()` still selects deterministically and does not panic.
- Existing `x/eibc` tests continue to pass unchanged — in particular the insufficient-funds fall-through/`Del` path and the agent-live gating in `lps_test.go`.
- `go test ./x/eibc/...` passes; `golangci-lint run` clean on changed files; `make proto-gen` produces no diff beyond the regenerated comment.

## Consensus safety

Identical result on every validator: `sha256` over a fixed byte ordering is deterministic; `ctx.HeaderHash()` is consensus data equal across validators for the block; the compatible-LP set comes from an ordered KeySet iterator (no map iteration); the only randomness is the seeded PCG shuffle, now seeded from consensus-derived bytes. No wall-clock, no `Math.random`, no panic (empty-hash path handled), and the fulfillment stays atomic (existing behavior unchanged). This strictly *removes* a source of caller-influenced nondeterminism-of-outcome; it introduces none.

## Out of scope

- No change to fee/price economics, `Accepts()` compatibility logic, `MaxSpend`, velocity/validity caps, or agent-live gating.
- No reputation-, stake-, or bid-weighted selection — winner stays uniform over the compatible set; only the seed source is fixed. (Weighted selection is a separate, and separately risky, design.)
- No removal of the `rng` proto field and no other proto message-shape changes.
- No new query or param.

---

omritoptix:
## Summary

- derive the on-demand LP shuffle seed from SHA-256(order ID || current header hash)
- ignore the deprecated transaction RNG while preserving its protobuf field and CLI argument
- cover caller-RNG independence, deterministic replay, order-ID variation, and empty header hashes

Closes #2194

### Caller RNG cannot influence the LP winner; consensus inputs remain deterministic
**How:** `go test ./x/eibc/keeper -run 'TestKeeperTestSuite/(TestTryFulfillOnDemand)' -count=1 -v`
**Evidence:**
```text
=== RUN   TestKeeperTestSuite/TestTryFulfillOnDemandEmptyHeaderHashDeterministic
=== RUN   TestKeeperTestSuite/TestTryFulfillOnDemandIgnoresCallerRNG
=== RUN   TestKeeperTestSuite/TestTryFulfillOnDemandSeedDeterminismAndOrderVariation
--- PASS: TestKeeperTestSuite (1.59s)
    --- PASS: TestKeeperTestSuite/TestTryFulfillOnDemandEmptyHeaderHashDeterministic (0.29s)
    --- PASS: TestKeeperTestSuite/TestTryFulfillOnDemandIgnoresCallerRNG (0.16s)
    --- PASS: TestKeeperTestSuite/TestTryFulfillOnDemandSeedDeterminismAndOrderVariation (1.14s)
PASS
ok  github.com/dymensionxyz/dymension/v3/x/eibc/keeper 1.737s
```

### The complete eIBC module suite passes
**How:** `go test ./x/eibc/... -count=1`
**Evidence:**
```text
ok  github.com/dymensionxyz/dymension/v3/x/eibc 0.412s
?   github.com/dymensionxyz/dymension/v3/x/eibc/client/cli [no test files]
ok  github.com/dymensionxyz/dymension/v3/x/eibc/keeper 4.343s
ok  github.com/dymensionxyz/dymension/v3/x/eibc/types 0.147s
```

### Repository build and lint gates pass
**How:** `go build ./...` and `golangci-lint-v2 run`
**Evidence:**
```text
$ go build ./...
(exit 0)

$ golangci-lint-v2 run
0 issues.
```

### Protobuf output is regenerated and the legacy field remains wire-compatible
**How:** `make proto-format` followed by `make proto-gen`
**Evidence:**
```text
Generating Protobuf files
Generating gogo proto code for ../proto/dymensionxyz/dymension/eibc/tx.proto
(exit 0)
```

### CLI documents the ignored compatibility argument and rejects a missing order ID
**How:** `dymd tx eibc try-fulfill-on-demand --help` and `dymd tx eibc try-fulfill-on-demand`
**Evidence:**
```text
Try to find a fulfiller for a given order and fulfill on the spot The optional rng argument is deprecated and ignored; the fulfiller is chosen from a chain-derived seed.

Usage:
  dymd tx eibc try-fulfill-on-demand [order-id] [rng] [flags]
help exit: 0

Error: requires at least 1 arg(s), only received 0
missing-order exit: 1
```

### CI race tests cover the full package set
**How:** `go-acc -o /tmp/coverage-2194.txt ./... -- -v --race`, then after its 20-minute wrapper expired, continuation from the first unfinished package with `go test -race ./x/lockup/... ./x/otcbuyback/... ./x/rollapp/... ./x/sequencer/... ./x/sponsorship/... ./x/streamer/...`
**Evidence:**
```text
ok  github.com/dymensionxyz/dymension/v3/x/eibc/keeper 6.701s coverage: 6.9
ok  github.com/dymensionxyz/dymension/v3/x/iro/types 352.062s coverage: 1.1
ok  github.com/dymensionxyz/dymension/v3/x/lightclient/keeper 3.168s coverage: 3.7
[20-minute wrapper expired after x/lightclient; no failing package or race report]
ok  github.com/dymensionxyz/dymension/v3/x/lockup/keeper 114.521s
ok  github.com/dymensionxyz/dymension/v3/x/otcbuyback/keeper 43.781s
ok  github.com/dymensionxyz/dymension/v3/x/rollapp/keeper 144.349s
ok  github.com/dymensionxyz/dymension/v3/x/sequencer/keeper 22.497s
ok  github.com/dymensionxyz/dymension/v3/x/sponsorship/keeper 26.854s
ok  github.com/dymensionxyz/dymension/v3/x/streamer/keeper 37.115s
```

**Not verified:** the single-process aggregate coverage file did not finish before the 20-minute wrapper; every package was nevertheless executed under the race detector across the initial run and the scoped continuation.

### DIFF
--- a/proto/dymensionxyz/dymension/eibc/tx.proto
+++ b/proto/dymensionxyz/dymension/eibc/tx.proto
@@ -124,7 +124,8 @@ message MsgTryFulfillOnDemand {
 
   string order_id = 1;
 
-  // rng to choose fulfiller from eligible randomly
+  // Deprecated: ignored. The fulfiller is chosen by a chain-derived
+  // deterministic seed.
   int64 rng = 3;
 }
 
--- a/x/eibc/client/cli/tx.go
+++ b/x/eibc/client/cli/tx.go
@@ -210,7 +210,7 @@ func NewCmdTryFulfillOnDemand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "try-fulfill-on-demand [order-id] [rng]",
 		Short: short,
-		Long:  short + " Can provide rng to avoid choosing same fulfiller multiple times (number). ",
+		Long:  short + " The optional rng argument is deprecated and ignored; the fulfiller is chosen from a chain-derived seed.",
 
 		Args: cobra.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
@@ -219,18 +219,11 @@ func NewCmdTryFulfillOnDemand() *cobra.Command {
 				return err
 			}
 			orderId := args[0]
-			rng := 0
-			if len(args) > 1 {
-				rng, err = strconv.Atoi(args[1])
-				if err != nil {
-					return err
-				}
-			}
 
 			msg := &types.MsgTryFulfillOnDemand{
 				Signer:  clientCtx.GetFromAddress().String(),
 				OrderId: orderId,
-				Rng:     int64(rng),
+				Rng:     0,
 			}
 
 			if err := msg.ValidateBasic(); err != nil {
--- a/x/eibc/keeper/lps.go
+++ b/x/eibc/keeper/lps.go
@@ -1,6 +1,8 @@
 package keeper
 
 import (
+	"crypto/sha256"
+	"encoding/binary"
 	"errors"
 	"math/rand/v2"
 
@@ -215,6 +217,20 @@ func (s LPs) NextID(ctx sdk.Context) (uint64, error) {
 // SetNextID sets the LP id sequence, used on genesis import.
 func (s LPs) SetNextID(ctx sdk.Context, id uint64) error {
 	return s.nextID.Set(ctx, id)
+}
+
+// deterministicFulfillSeed derives the LP-shuffle seed from data the tx
+// submitter cannot choose: the order id (per-order variation) mixed with the
+// current block hash (per-block entropy neither the fulfiller nor the order
+// creator controls at their respective action times). Consensus-safe: the
+// header hash is consensus data identical across validators, and sha256 with
+// fixed byte order is deterministic.
+func deterministicFulfillSeed(ctx sdk.Context, orderID string) uint64 {
+	h := sha256.New()
+	_, _ = h.Write([]byte(orderID))
+	_, _ = h.Write(ctx.HeaderHash())
+	sum := h.Sum(nil)
+	return binary.BigEndian.Uint64(sum[:8])
 }
 
 func (k Keeper) FulfillByOnDemandLP(ctx sdk.Context, order string, rng uint64) error {
--- a/x/eibc/keeper/lps_test.go
+++ b/x/eibc/keeper/lps_test.go
@@ -1,0 +1,73 @@
+package keeper_test
+
+import (
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dymensionxyz/dymension/v3/app/apptesting"
+	"github.com/dymensionxyz/dymension/v3/x/eibc/types"
+)
+
+func (suite *KeeperTestSuite) fulfillWinner(orderSeq uint64, headerHash []byte, rng int64, addrs []sdk.AccAddress) uint64 {
+	suite.SetupTest()
+	suite.Ctx = suite.Ctx.WithBlockHeight(10).WithHeaderHash(headerHash)
+	for _, addr := range addrs {
+		apptesting.FundAccount(suite.App, suite.Ctx, addr, sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, math.NewInt(1_000_000))))
+	}
+
+	k := suite.App.EIBCKeeper
+	for _, addr := range addrs[1:] {
+		_, err := k.LPs.Create(suite.Ctx, &types.OnDemandLP{
+			FundsAddr:  addr.String(),
+			Rollapp:    rollappPacket.RollappId,
+			Denom:      sdk.DefaultBondDenom,
+			MaxPrice:   math.NewInt(100),
+			MinFee:     math.LegacyZeroDec(),
+			SpendLimit: math.NewInt(1_000),
+		})
+		suite.Require().NoError(err)
+	}
+
+	orderID := suite.orderWithSeq(orderSeq, addrs[0].String(), math.NewInt(90), math.NewInt(10))
+	_, err := suite.msgServer.TryFulfillOnDemand(suite.Ctx, &types.MsgTryFulfillOnDemand{
+		Signer:  addrs[0].String(),
+		OrderId: orderID,
+		Rng:     rng,
+	})
+	suite.Require().NoError(err)
+
+	lps, err := k.LPs.GetAll(suite.Ctx)
+	suite.Require().NoError(err)
+	for _, lp := range lps {
+		if lp.Spent.IsPositive() {
+			return lp.Id
+		}
+	}
+	suite.FailNow("no LP recorded spend")
+	return 0
+}
+
+func (suite *KeeperTestSuite) TestTryFulfillOnDemandIgnoresCallerRNG() {
+	addrs := apptesting.CreateRandomAccounts(3)
+	winnerFromZero := suite.fulfillWinner(101, []byte("fixed-header"), 0, addrs)
+	winnerFromOne := suite.fulfillWinner(101, []byte("fixed-header"), 1, addrs)
+	suite.Require().Equal(winnerFromZero, winnerFromOne)
+}
+
+func (suite *KeeperTestSuite) TestTryFulfillOnDemandSeedDeterminismAndOrderVariation() {
+	addrs := apptesting.CreateRandomAccounts(3)
+	headerHash := []byte("fixed-header")
+	winner := suite.fulfillWinner(201, headerHash, 0, addrs)
+	suite.Require().Equal(winner, suite.fulfillWinner(201, headerHash, 0, addrs))
+
+	winners := map[uint64]struct{}{}
+	for seq := uint64(201); seq < 217; seq++ {
+		winners[suite.fulfillWinner(seq, headerHash, 0, addrs)] = struct{}{}
+	}
+	suite.Require().Greater(len(winners), 1)
+}
+
+func (suite *KeeperTestSuite) TestTryFulfillOnDemandEmptyHeaderHashDeterministic() {
+	addrs := apptesting.CreateRandomAccounts(3)
+	winner := suite.fulfillWinner(301, nil, -1, addrs)
+	suite.Require().Equal(winner, suite.fulfillWinner(301, nil, 42, addrs))
+}
--- a/x/eibc/keeper/msg_server.go
+++ b/x/eibc/keeper/msg_server.go
@@ -278,8 +278,8 @@ func (m msgServer) TryFulfillOnDemand(goCtx context.Context, msg *types.MsgTryFu
 		return nil, errorsmod.Wrap(err, "vbasic")
 	}
 
-	shuffleSeed := uint64(msg.Rng) //nolint:gosec
-	err = m.FulfillByOnDemandLP(ctx, msg.OrderId, shuffleSeed)
+	seed := deterministicFulfillSeed(ctx, msg.OrderId)
+	err = m.FulfillByOnDemandLP(ctx, msg.OrderId, seed)
 	if err != nil {
 		return nil, err
 	}
--- a/x/eibc/types/tx.pb.go
+++ b/x/eibc/types/tx.pb.go
@@ -479,7 +479,8 @@ var xxx_messageInfo_MsgUpdateDemandOrderResponse proto.InternalMessageInfo
 type MsgTryFulfillOnDemand struct {
 	Signer  string `protobuf:"bytes,2,opt,name=signer,proto3" json:"signer,omitempty"`
 	OrderId string `protobuf:"bytes,1,opt,name=order_id,json=orderId,proto3" json:"order_id,omitempty"`
-	// rng to choose fulfiller from eligible randomly
+	// Deprecated: ignored. The fulfiller is chosen by a chain-derived
+	// deterministic seed.
 	Rng int64 `protobuf:"varint,3,opt,name=rng,proto3" json:"rng,omitempty"`
 }
 


omriai1 (commented):
## Review: approve

The change does what the issue asked. The selection seed is no longer reachable from the transaction, the regression tests genuinely discriminate between old and new behavior, and the proof of execution reproduces. Findings below are all non-blocking; one of them I would like addressed in this PR because it is a one-line honesty fix on a security rationale.

### What I verified rather than assumed

**The root cause is actually gone.** `msg.Rng` is no longer read anywhere in a consensus path — the only remaining references are the generated struct and the CLI writing a literal zero. The message server is the sole production caller into on-demand fulfillment; every other call site is a test. So there is no second door left open.

**Consensus determinism holds.** During message execution the header hash is the block hash the whole validator set agreed on, so SHA-256 over the order id and that hash with fixed byte order gives every validator the same seed. Messages never execute in CheckTx — BaseApp refuses to dispatch outside finalize and simulate modes — so the empty header hash on check state is harmless. The digest is always 32 bytes, so taking the first 8 cannot panic even when the header hash is empty during replay or genesis. No apphash divergence risk.

**The tests are load-bearing, not tautological.** This mattered enough to check directly, because a "caller RNG is ignored" test is worthless if the two seeds it compares happened to produce the same winner under the old code. I computed the pre-PR shuffle for exactly the seed pairs the tests use: with two eligible LPs, seed 0 puts the second LP first while seed 1 puts the first LP first, and the same inversion holds for the -1 versus 42 pair. So both comparisons genuinely fail on the old code. A second reviewer independently confirmed this by reverting the one-line seed substitution and watching all three tests fail with meaningful diffs.

**Proof of execution reproduces.** I ran the three new tests and they pass. The protobuf change is comment-only — zero bytes of the gzipped file descriptor moved — so the wire format is untouched and the claim about compatibility is accurate. Keeping the field inert rather than reserving it is the right call: transactions already signed with it must still decode. No documentation went stale; nothing in the repo's markdown references the rng argument.

### Please address before merge

**The helper's rationale is accurate but incomplete, and the omission is the interesting part.** The comment says the seed comes from data neither the fulfiller nor the order creator controls, which is true. What it does not say is that one actor still does have leverage. The block hash Merkle-commits to the block's own transaction set, and CometBFT deliberately lags the app hash by one block, so every other header field is already fixed from prior state when a proposer assembles its block. That means a proposer can vary its own transaction set, recompute the block hash locally, derive the seed, and check which LP wins before signing — roughly one try per eligible LP, all offline. In its own proposal slots, the lottery is grindable.

I am not asking you to change the mechanism, for three reasons. It strictly shrinks the attacker set from "anyone who can send a transaction" to "the current proposer during its own slot." It is consistent with what this codebase already ships — the shared randomness helper hashes the same header hash with a salt, and the streamer pump uses it to decide whether a pump fires and how large, which is more value at stake with identical exposure. And the tempting alternative is actively worse: seeding from a *prior* block hash makes the seed public before the fulfiller submits, which hands free grinding back to every account, just one block slower. Genuinely closing this needs a randomness beacon, commit-reveal, or replacing the lottery with a deterministic economic rule such as a rotating cursor or oldest-unserved — a chain-wide decision, not this PR.

What I want is one sentence in the doc comment naming the residual proposer surface, and a follow-up issue tracking it across both this call site and the shared randomness helper. The acceptance criterion is simply that a future reader of that comment does not come away believing the selection is unconditionally ungameable.

### Non-blocking

**The block-hash half of the seed is the one thing the tests never vary.** Every case pins the header hash to a single fixed value or nil. That half is precisely what makes the outcome unknowable at signing time, so it carries the security argument, and it currently has no assertion behind it. A case that holds the order id constant, sweeps several header hashes, and asserts more than one distinct winner would close the gap and make the dependence visible to the next reader.

**The unsafe call shape is still there.** The fulfillment function continues to accept a caller-supplied seed parameter. Nothing passes a bad one today, but the class of the original bug was "selection seed comes from the caller," and leaving the injection point open means a future caller can reintroduce it. Deriving the seed inside the function — it already has the context and the order id — removes the shape entirely. The existing tests that pass a literal zero would set a header hash instead.

**The CLI silently swallows a second positional argument.** The usage line still advertises the rng argument and the arg count is a minimum rather than exact, so someone who passes a number gets it discarded with no diagnostic. Dropping it from the usage string and tightening the arg count makes the deprecation visible where it is used rather than only in help text. I would leave the proto field alone, though — that one needs to stay for wire compatibility.

### For the record, two things I checked and cleared

There is no retry-until-favorable loop for ordinary accounts. Fulfillment succeeds whenever any funded compatible LP exists — an LP that cannot pay is skipped and pruned, not treated as fatal — and success stamps the fulfiller and consumes the order, so the shuffle is drawn exactly once in an order's lifetime. Resubmitting in a later block hits an already-fulfilled lookup, not a fresh draw.

The conditional-revert variant is also closed. To abort atomically when the wrong LP wins, an attacker would need a message that fails when their balance is *unspent*, which is the opposite polarity of every message in this repo. There is no CosmWasm to write one, and the ante layer rejects Ethereum transactions bundled into Cosmos transactions, including through authz, gov, and group nesting.

PR #2197, opened 2026-08-08
feat(agent): restrict attested transfer recipients

### LINKED ISSUE #2196: x/agent: recipient allowlist for attested spend authority (payee-restricted spend policy — the missing pillar of agent policy-based spending controls)
## Problem / opportunity

`x/agent`'s attested spend authority (`MsgSubmitAttestedTransfer`, added in #2184) lets a TEE-attested agent pay out from its owner-funded escrow. The owner's spend policy today constrains **how much** (`spend_limit_per_window`) and **which asset** (`spend_denom`, single denom) — but **not where the money goes**. `SubmitAttestedTransfer` (`x/agent/keeper/msg_server.go:125`) sends `spend_denom` to *any* `msg.Recipient` that carries a valid attestation.

Consequence: if an agent's TEE signing key is compromised, or a bug/prompt-injection makes the agent emit a valid-but-malicious transfer, an attacker can drain the **entire** windowed budget to an arbitrary address they control — rate-limited, but not prevented. The escrow's whole point is to bound blast radius; without a payee constraint the bound is "attacker gets `spend_limit_per_window` every window" instead of "attacker gets nothing."

The `Agent` proto (`proto/dymensionxyz/dymension/agent/agent.proto`) has `spend_denom`, `spend_limit_per_window`, `spend_window_blocks`, `spend_window_*` bookkeeping — and **no recipient constraint** (confirmed: `grep -ri allowlist|payee|recipient x/agent` returns nothing on the policy).

## Why now

"Policy-based spending controls" for autonomous agents have converged on **payee restrictions** as a first-class control, alongside amount limits and audit trails:

- **AWS Bedrock AgentCore Payments (Preview, May 2026)** ships "wallet management, **policy-based spending controls**, and a full audit trail" as the three pillars of agent payment safety. Dymension's agent escrow already has the wallet (escrow) and the audit trail (attested action log); the missing pillar is destination-scoped spend policy. — https://aws.amazon.com/blogs/industries/x402-and-agentic-commerce-redefining-autonomous-payments-in-financial-services/
- **x402** governance moved to the Linux Foundation (Apr 2026) with a Q3 2026 v1.0 target; V2 formalizes session tokens and **policy-based spending controls** for facilitator-mediated agent payments. — https://www.chainalysis.com/blog/x402-agentic-payments-adoption/
- **ERC-8004 (Trustless Agents)** community analysis repeatedly names "the gap between verified identity and honest behavior" as the residual risk after attestation — a payee allowlist is the standard on-chain mitigation. — https://eips.ethereum.org/EIPS/eip-8004

An attested identity proves *who signed*; it does not prove the *destination is intended*. A recipient allowlist closes that gap.

## Proposed change

Add an **optional, owner-configured recipient allowlist** to the agent spend policy. Empty allowlist = unrestricted (preserves current behavior and is backward-compatible with every already-registered agent). A non-empty allowlist means attested transfers may only pay an address in the set.

**Proto — `proto/dymensionxyz/dymension/agent/agent.proto`**
- `Agent`: add `repeated string spend_recipient_allowlist = 13;` (bech32 account addresses).
- `MsgUpdateAgentSpendPolicy` (`tx.proto`): add `repeated string recipient_allowlist = 5;` so the owner sets it via the existing spend-policy message.
- `EventUpdateAgentSpendPolicy`: add `repeated string recipient_allowlist = 5;`.
- `Params`: add `uint64 spend_recipient_allowlist_max = 6;` to bound per-agent state growth.

Run `make proto-gen`.

**`x/agent/types/params.go`**
- `DefaultSpendRecipientAllowlistMax = 50`; wire into `DefaultParams()` and `Params.Validate()` (reject 0).

**`x/agent/types/agent.go`** — new helper:
```go
// SpendRecipientAllowed reports whether recipient may receive an attested
// transfer. An empty allowlist means unrestricted (spending is still bounded
// by denom + per-window limit).
func (a Agent) SpendRecipientAllowed(recipient string) bool {
    if len(a.SpendRecipientAllowlist) == 0 {
        return true
    }
    for _, r := range a.SpendRecipientAllowlist { // linear scan: deterministic, no map iteration
        if r == recipient {
            return true
        }
    }
    return false
}
```
Extend `ValidateSpendState()` (genesis guard) to validate the stored allowlist: each entry a valid bech32, no duplicates, `len <= params.SpendRecipientAllowlistMax`. Since `ValidateSpendState` has no params handle today, pass the max in (or validate count in `keeper.InitGenesis` where params are available) — implementer's call, but the cap MUST be enforced on genesis import.

**`x/agent/types/message_escrow.go`** — `MsgUpdateAgentSpendPolicy.ValidateBasic()`: for each `recipient_allowlist` entry, `sdk.AccAddressFromBech32` must succeed; reject duplicates. (Count-vs-param-max is enforced in the keeper where params are available.)

**`x/agent/keeper/escrow.go`** — `UpdateAgentSpendPolicy`: after loading params, reject when `len(msg.RecipientAllowlist) > params.SpendRecipientAllowlistMax`. Set `agent.SpendRecipientAllowlist = msg.RecipientAllowlist` alongside the existing denom/limit/window assignments. Keep the existing unconditional window reset — do not attempt to preserve window state selectively. Include the allowlist in the emitted `EventUpdateAgentSpendPolicy`.

**`x/agent/keeper/msg_server.go`** — `SubmitAttestedTransfer`: immediately after the `SpendEnabled()` check (before the attestation verify — fail fast, no state touched):
```go
if !agent.SpendRecipientAllowed(msg.Recipient) {
    return nil, errorsmod.Wrap(types.ErrRecipientNotAllowed, msg.Recipient)
}
```

**`x/agent/types/errors.go`** — add `ErrRecipientNotAllowed = errorsmod.Register(ModuleName, 17, "transfer recipient not in agent spend allowlist")` (17 is the next free code).

The attestation nonce (`AttestedTransferBytes(recipient, denom, amount, memo)`) is unchanged — the allowlist is an independent on-chain gate, not part of the signed payload.

- **Empty allowlist = unrestricted.** Backward-compatible; existing agents keep working; owners opt in to restriction. (Alternative "empty = deny all" would silently brick every live spend-enabled agent on upgrade — rejected.)
- **Owner sets it via the existing `MsgUpdateAgentSpendPolicy`,** not a new message. It's a spend-policy attribute like denom/limit.
- **Not timelocked; takes effect immediately** — matches the existing documented rationale that the spend policy is the owner's own guardrail over its own escrow (`escrow.go` comment). Tightening the allowlist is a security action and must be instant.
- **Window reset on update stays as-is** (unconditional), consistent with current behavior.
- **Membership via linear scan over a repeated string** — deterministic, no map iteration in consensus. Store entries as supplied (after dedup validation); no sorting required for correctness, but the field serializes in a stable order.
- **`spend_recipient_allowlist_max = 50`** default — generous for real fan-out payees, small enough to bound state and keep the linear scan trivial. Governable via params.
- **Duplicates rejected** in `ValidateBasic` (explicit over silent dedup).
- **Naming:** proto `spend_recipient_allowlist` / msg `recipient_allowlist`; param `spend_recipient_allowlist_max`; error `ErrRecipientNotAllowed`; helper `SpendRecipientAllowed`.

## Consensus safety

Pure deterministic set-membership over a bounded `[]string`, evaluated inside a normal tx that rolls back atomically on the error. No wall-clock, no randomness, no map iteration in consensus, no Begin/EndBlocker involvement, no panic path. Bounded by `spend_recipient_allowlist_max`. Identical on every validator.

## Acceptance criteria

- `SubmitAttestedTransfer` to an address **not** in a non-empty allowlist fails with `ErrRecipientNotAllowed`; escrow balance, `action_seq`, and spend-window bookkeeping are unchanged and no `EventAttestedTransfer` is emitted (tx rolled back).
- `SubmitAttestedTransfer` to an address **in** the allowlist succeeds exactly as today (payout, window record, action-log append, event).
- Agent with **empty** allowlist: any recipient allowed — all existing `SubmitAttestedTransfer` tests still pass unmodified.
- `MsgUpdateAgentSpendPolicy` with an invalid-bech32 allowlist entry → `ValidateBasic` error; with a duplicate entry → error; with `len > SpendRecipientAllowlistMax` → keeper error.
- `UpdateAgentSpendPolicy` persists the allowlist and emits it in `EventUpdateAgentSpendPolicy`; a subsequent gRPC `Agent` query returns it.
- Genesis import of an agent whose stored allowlist is invalid or exceeds the max is rejected by `ValidateSpendState`/`InitGenesis`.
- `make proto-gen` produces no uncommitted diff; `go test ./x/agent/...` green; `golangci-lint run` clean on touched files.

## Out of scope

- IBC / cross-chain recipients (attested transfer stays a local bank send).
- Per-recipient sub-budgets and a distinct per-transfer max amount (separate future control).
- A governance-gated **global** recipient denylist (analogous to #2179 for policies) — separate issue.
- Changing the single-`spend_denom` design, timelocking the spend policy, or altering the attestation payload format.

---

omriai1:
Closes #2196

Attested escrow transfers can now be restricted to an owner-configured recipient allowlist, so a valid TEE signature cannot redirect a bounded budget to an arbitrary payee. Empty allowlists preserve existing unrestricted behavior; non-empty policies are validated, state-bounded, enforced before attestation verification, exposed by queries/events, and safely initialized on chain upgrade.

- Adds the recipient policy, governed maximum, validation, persistence, query, and event fields.
- Rejects disallowed recipients before attestation verification without touching escrow, logs, sequence, window state, or events.
- Migrates v1 agent params to the default maximum of 50 and preserves existing protobuf wire fields.
- Covers allowed, denied, invalid, duplicate, over-cap, genesis, and migration behavior.

- `Agent.spend_recipient_allowlist` uses protobuf field 13; `EventUpdateAgentSpendPolicy.recipient_allowlist` uses field 5.
- `MsgUpdateAgentSpendPolicy.recipient_allowlist` uses field 6 because field 5 is already the published `spend_window_blocks` field; retaining field 5 avoids a wire-breaking renumber.
- `Params.spend_recipient_allowlist_max` uses protobuf field 9, defaults to 50 and rejects zero.
- Duplicate validation compares decoded account bytes, catching equivalent Bech32 case variants while retaining supplied list order and text.
- Module consensus version advances to 2 with a v1-to-v2 params migration.
- The attested-transfer nonce and signed payload remain unchanged.

</details>

## Proof of execution
- **Codecov project coverage — PASS:** Build and Test run `31891962334` obtained a GitHub OIDC token, uploaded the 32.8 MB coverage report successfully, and both `codecov/project` and `codecov/patch` passed.
- **Buf breaking compatibility — PASS:** pinned Buf 1.50.0 `buf breaking proto` against `main` completed with exit code 0; the CI setup now authenticates GitHub API downloads via `${{ github.token }}`.
- **Workflow validation — PASS:** targeted Actionlint and PyYAML parsing passed for `.github/workflows/proto.yaml`.

- **Build and Test / full repository post-merge race suite — PASS:** `go test -race ./...` (all packages, including `x/agent/keeper` and `x/agent/types`).
- **Build and Test / full repository post-merge build — PASS:** `go build ./...`.
- **golangci-lint — PASS:** `golangci-lint v2.1.6 run` reported `0 issues.`
- **Proto generation — PASS:** a second `scripts/protocgen.sh` plus `go mod tidy` left `agent.pb.go`, `tx.pb.go`, `go.mod`, and `go.sum` byte-identical.
- **Focused agent regression suite — PASS:** `go test -count=1 ./x/agent/...` covered unrestricted, allowed, denied/no-state-change, validation, persistence/query/event, genesis, and migration paths.

<details><summary>Backend proof notes</summary>

The keeper regression suite runs against the real application bank keeper, module account, collections state, typed event manager, and gRPC query implementation. The only substituted boundary is the external GCP attestation verifier, where the established fake verifier accepts the nonce derived by the handler; the denied-recipient test additionally proves the verifier is never called.

</details>

**Not verified:** a live GCP-attested network transaction was not executed because locally minting a valid GCP attestation token is unavailable; the changed handler and its state/event effects are covered by the real-app keeper integration suite.


omriai2 (changes requested):
# Security Review — PR #2197 (`feat(agent): restrict attested transfer recipients`)

No HIGH or MEDIUM severity vulnerabilities met the confidence threshold (≥8) for reporting.

## Analysis summary

The change adds an owner-configured recipient allowlist that gates `MsgSubmitAttestedTransfer` payouts from agent escrow. The following attack surfaces were examined and cleared:

**Authorization gate placement** — `SpendRecipientAllowed` is evaluated in `SubmitAttestedTransfer` after the agent load and spend-enabled check but *before* nonce derivation, attestation verification, budget accounting, escrow debit, and bank send. A denied recipient therefore returns before any state write, and the transaction rolls back regardless. No partial-state or oracle-probing path was found.

**Bypass analysis of the string comparison** — enforcement compares `msg.Recipient` against stored entries with exact string equality, while `ValidateBasic` and `ValidateSpendState` deduplicate on decoded account bytes. This asymmetry can only produce *false negatives* (a legitimate transfer using a different valid Bech32 case form of an allowlisted account is rejected). To pass the gate, the submitter must supply a string byte-identical to an allowlist entry, which decodes to exactly the allowlisted account via `MustAccAddressFromBech32`. There is no input that satisfies the gate while routing funds elsewhere, so this is a fail-closed correctness inconsistency, not an authorization bypass.

**Empty-allowlist semantics** — an empty list means unrestricted. This preserves the module's pre-existing behavior rather than introducing a new permissive path, and spending remains bounded by denom and per-window limit. Not a newly introduced weakness.

**Owner-initiated clearing** — `UpdateAgentSpendPolicy` unconditionally overwrites the stored allowlist with the message field, so an update omitting the list clears the restriction. This requires the agent owner's authority, which is the same principal permitted to set the allowlist to empty directly, so it is not a privilege boundary crossing. (It is a real wiring/footgun problem, but it is not an attacker-reachable security bypass.)

**Params and migration** — `Migrate1to2` unconditionally writes the default cap. Because the field did not exist at consensus version 1, no stored operator-chosen value can be clobbered. Prior to migration the cap reads as zero, which rejects all non-empty allowlists — again fail-closed.

**Wire compatibility** — `Agent` field 13, `Params` field 6, `EventUpdateAgentSpendPolicy` field 5, and `MsgUpdateAgentSpendPolicy` field 6 are all previously unassigned numbers in their respective messages. No field number is reused, so no existing persisted state or in-flight message can be reinterpreted under the new schema.

**Input handling** — every allowlist entry is decoded through `sdk.AccAddressFromBech32`, which enforces the account HRP and checksum, before storage. Entry count is bounded by the governed `spend_recipient_allowlist_max`. No unvalidated string reaches a store key, a bank operation, or any external call.

omriai1:
Security review follow-up is pushed in `2001c2e6`: allowlist membership now compares decoded account bytes, and the spend-policy CLI now requires an explicit allowlist argument so legacy four-argument updates cannot silently clear the restriction.

## Proof of execution

- `go test -count=1 ./x/agent/types ./x/agent/client/cli ./x/agent/keeper` — PASS.
- `golangci-lint run ./x/agent/types ./x/agent/client/cli ./x/agent/keeper` — PASS, 0 issues.
- Consumer audit — `SpendRecipientAllowed` is consumed only by `SubmitAttestedTransfer`; `newMsgUpdateAgentSpendPolicy` is consumed only by `CmdUpdateAgentSpendPolicy`, and all three affected package suites were run.

The remaining reviewed properties are unchanged and covered by the existing branch tests: fail-fast gate placement/no state changes, empty-list backward compatibility, owner-only updates, bounded and validated inputs, v1→v2 parameter migration, and non-reused protobuf field numbers.

omriai2 (approved):
**APPROVE** — the follow-up commit resolves both findings from the prior round at the right level (a *required* CLI positional so the allowlist can never be silently dropped, and byte-canonical address comparison at the enforcement gate), each backed by a committed regression test, and I found no new blocking issue.

## Prior findings — disposition

| # | Prior finding | Status |
|---|---|---|
| 1 | Allowlist unreachable from the shipped CLI, and any CLI-driven spend-policy update silently cleared an allowlist set elsewhere | **Resolved** |
| 2 | Enforcement compared recipient addresses as raw strings while validation deduplicated on decoded bytes — a legitimate transfer using a different valid Bech32 case form was rejected | **Resolved** |

## Index — non-blocking only

1. *(nit)* Comma-split of the CLI allowlist argument does not trim whitespace, so `"addrA, addrB"` fails validation on the second entry — fail-closed with a clear error, so cosmetic only.
2. *(nit)* The `x/agent` package doc still describes spending as bounded only by the owner-set per-window budget and does not mention the recipient restriction — incomplete rather than false, so not a docs-honesty failure.
3. *(observation)* The v1→v2 params migration appears to be unreachable dead code, since `x/agent` has never shipped at consensus version 1.
4. *(nit)* In the spend-policy handler the allowlist-cap check runs before the owner-authorization check, so a non-owner submitting an over-cap list gets an invalid-argument error rather than unauthorized.

---

<details><summary>Finding 1 (prior, blocking) — CLI wiring and silent allowlist clearing — RESOLVED</summary>

**What was wrong.** The hand-written `update-spend-policy` command took exactly four positional arguments and built its message through a constructor that never touches the allowlist field. Since the handler unconditionally assigns the stored allowlist from the message, two things followed: an owner had no way to configure the new restriction through the shipped client at all, and — worse — an owner bumping their spend limit through that same command would silently wipe an allowlist that had been set by any other means. A safety gate on money that a routine, unrelated owner action quietly disables is not a safety gate.

**How it was fixed.** The command now takes a fifth positional argument, `[recipient-allowlist]`, parsed as a comma-separated address list, with the empty string meaning explicitly-unrestricted. The message construction moved into a testable helper. Critically the argument is **required** (`ExactArgs(5)`), not an optional flag — that is the stronger of the two shapes I would have accepted, because it makes the unrestricted case something the owner has to state rather than something they fall into by omission. The silent-clearing path is gone by construction: there is no invocation of this command that leaves the allowlist unspecified.

**Evidence.** The committed CLI test covers all three behaviors that matter: that the five-argument form is accepted and the old four-argument form is rejected, that a comma-separated pair parses into the two addresses in order, and that the empty string yields an empty allowlist rather than a one-element list containing `""`.

```
x/agent/client/cli/tx_escrow.go
-  Use:  "update-spend-policy [agent-id] [spend-denom] [spend-limit-per-window] [spend-window-blocks]"
-  Args: cobra.ExactArgs(4)
+  Use:  "update-spend-policy [agent-id] [spend-denom] [spend-limit-per-window] [spend-window-blocks] [recipient-allowlist]"
+  Args: cobra.ExactArgs(5)
+  if args[4] != "" { msg.RecipientAllowlist = strings.Split(args[4], ",") }
```

I checked whether the arity change breaks anything else that invokes this command. `update-spend-policy` appears nowhere outside the CLI source itself — no scripts, no e2e helpers, no repo documentation reference it — so the change is self-contained and no doc was made stale by it.

</details>

<details><summary>Finding 2 (prior, non-blocking) — address canonicalization at the enforcement gate — RESOLVED</summary>

**What was wrong.** Validation deduplicated allowlist entries by decoded account bytes, correctly treating two Bech32 case variants as the same account — but enforcement then compared the incoming recipient against stored entries with exact string equality. The two halves disagreed about what "the same recipient" means. This could only ever fail closed (a legitimate enclave-authorized transfer to an allowlisted account gets rejected because the submitter spelled the address in the other valid case form), which is why I raised it as non-blocking rather than as a bypass. But it left the policy's semantics defined by address *text* while its validation was defined by address *identity*, which is the kind of mismatch that ages badly.

**How it was fixed.** The gate now decodes the incoming recipient once, decodes each allowlist entry, and compares the resulting account bytes. Both directions of the earlier asymmetry are closed, and the failure modes stay conservative: an undecodable recipient returns denied rather than falling through, and an allowlist entry that somehow failed to decode is skipped rather than treated as a wildcard. Neither branch is reachable in practice — the transfer message validates its recipient before the handler runs, and both the message and genesis validators reject undecodable allowlist entries at write time — but fail-closed is the right posture for defensive code on a money path.

**Evidence.** The committed unit test asserts the previously-broken case directly: an allowlist holding the lowercase form admits the uppercase form of the same account, a different account is denied, and an agent with no allowlist admits anyone.

```
x/agent/types/agent.go
+  address, err := sdk.AccAddressFromBech32(recipient)
+  if err != nil { return false }
   for _, r := range a.SpendRecipientAllowlist {
+    allowedAddress, err := sdk.AccAddressFromBech32(r)
+    if err == nil && bytes.Equal(allowedAddress, address) { return true }
```

The comparison cost is bounded by the governed `spend_recipient_allowlist_max` (default 50), so the per-transfer decode loop has a hard ceiling, and address decoding is deterministic across validators since the account prefix comes from global chain config.

</details>

<details><summary>Finding 3 (nit, non-blocking) — CLI allowlist parsing does not trim whitespace</summary>

`strings.Split(args[4], ",")` splits on the comma alone. An operator who types the list the way lists are usually typed — `"dym1aaa..., dym1bbb..."` — produces a second entry with a leading space, which fails Bech32 decoding in `ValidateBasic` and rejects the whole message.

This is cosmetic, not a correctness problem: the failure is loud, immediate, client-side-validated, and the error text names the allowlist address, so the operator learns what to fix. A trailing comma behaves the same way. I mention it only because a `strings.TrimSpace` on each element would remove a small papercut on a command an operator types by hand under time pressure. Entirely optional.

</details>

<details><summary>Finding 4 (nit, non-blocking) — package doc does not mention the recipient restriction</summary>

The `x/agent` package comment describes agents as paying "attested transfers under an owner-set, per-window spend budget". That statement is still true after this change — the per-window budget still binds — so the doc has not been made inaccurate, which is why this is not a docs-honesty blocker. It is simply now incomplete: the recipient allowlist is a second owner-set constraint on the same money path and is the sort of thing a reader of the package doc would want named alongside the budget. A clause would cover it.

There is no README in the module and no repository documentation that describes the spend-policy CLI, so nothing else went stale with this change.

</details>

<details><summary>Finding 5 (observation, non-blocking) — the v1→v2 migration looks unreachable</summary>

The module bumps its consensus version to 2 and registers a v1→v2 migration that seeds the new allowlist cap. For that migration to run, some live chain must have `x/agent` recorded at consensus version 1. As far as I can tell none does: the newest release tag in this repository is `v4.0.2`, `x/agent` was introduced well after it, and the `v5` upgrade handler contains no reference to the agent module. If the module first reaches a network at version 2, the upgrade machinery will run its `InitGenesis` and record version 2 directly, and the registered migration is never invoked.

I am flagging this as an observation rather than a request. The migration is cheap, correct if it ever does run, and unconditionally overwriting the cap is safe precisely because the field cannot have held an operator-chosen value at version 1. If the module is in fact already live at version 1 on a devnet or playground network, then the migration is not dead code at all and this note is simply wrong — the author is better placed to know that than I am. Either way it does not hold the merge.

</details>

<details><summary>Finding 6 (nit, non-blocking) — check ordering in the spend-policy handler</summary>

`UpdateAgentSpendPolicy` validates the allowlist against the governed cap before it looks up the agent and checks that the caller is its owner. A non-owner submitting an over-cap list therefore receives an invalid-argument error instead of unauthorized, which technically distinguishes "cap exceeded" from "not your agent" to a caller who is neither.

This leaks nothing: the cap is a public params value and agent ownership is readable from the public agent query, so no state is disclosed that an ordinary query would not give up first. Moving the ownership check ahead of the cap check would read more naturally — authorize, then validate — but there is no security consequence to the current order.

</details>

---

**Proof of execution.** The section is present and correctly shaped: every claim carries a receipt line naming the command or suite that produced it — full-repository race suite, build, `golangci-lint`, proto-regeneration determinism, and the focused `x/agent` regression run — plus an explicit and honest "not verified" note that a live GCP-attested network transaction was not executed because a valid attestation token cannot be minted locally. Every one of those claims is covered by a CI-gated suite, so under the review rules the named suite plus the PR's required checks is the proof and no transcript spot-check is warranted. On the current head the checks are running with nothing red; `buf-break-check` and the label job have already passed, which independently confirms the proto additions are not wire-breaking. The remaining pending checks are the author's required gate, not something I hold the review for.

**Behavior-change regression coverage** is committed at every layer this touches: the CLI construction path, the address-equivalence gate, the keeper's denied-recipient path (which asserts no escrow movement, no action-log entry, no sequence advance, no window accounting, no events, and that the attestation verifier was never called), the persistence/query/event round trip, the over-cap rejection, genesis validation, and the params migration.

VERDICT: APPROVE

omritoptix:
@copilot resolve the merge conflicts in this pull request

omriai2 (approved):
**APPROVE** — the shipped tree is a correct reconciliation of this PR's approved allowlist feature with the sibling ERC-8004 validation registry: both sides' proto fields, error codes, params and validations survive intact, the two colliding numbering spaces were resolved away from main's claims rather than over them, and the merged result builds, lints and tests clean locally.

## Index — all findings non-blocking

1. *(observation)* One non-merge commit sits in the range, which per protocol widened this into a delta re-review — it turns out to be pure conflict-marker resolution, and its content does not reach the shipped tree at all.
2. *(observation)* An intermediate merge commit was pushed with unresolved conflict markers still in the source, leaving one commit in branch history whose tree does not compile — a bisectability cost only, already superseded.
3. *(observation)* The merge widened this PR's params migration to also seed the sibling's validation params — a genuine behavioral addition beyond either parent, which I judge correct and arguably necessary, but which deserves an explicit callout rather than passing as textual merge work.
4. *(infrastructure, needs a re-run not a fix)* `buf-break-check` is red on the merged head because the buf setup step hit a GitHub API rate limit and died in five seconds, never reaching the breaking-change comparison; the proto delta against main contains no breaking change.
5. *(nit)* The generated-code round-trip test still leaves both new proto fields unpopulated, so their hand-merged codecs are covered only indirectly through store round trips in the keeper tests.

---

<details><summary>Finding 1 — a non-merge commit in the range, and why it does not affect the shipped tree</summary>

The range since approval contains five commits, and their parentage is what matters:

```
d1509fef parents=[e63daaf0 1b7f5900]  chore(agent): merge remote conflict resolution
e63daaf0 parents=[2001c2e6 70649644]  chore(agent): merge main into recipient allowlist
1b7f5900 parents=[5b1cb46b]           merge: resolve conflicts with main (…field 9…code 24)
5b1cb46b parents=[2001c2e6 70649644]  chore: placeholder for merge conflict resolution plan
70649644 parents=[28618e97]           feat(agent): add ERC-8004 validation registry (#2199)
```

`70649644` is the sibling PR itself, reachable only because it is on the base branch — not branch work. But `1b7f5900` has a single parent, so it is a non-merge commit and the protocol says to widen. I did.

What happened is that two people resolved the same conflict independently. `5b1cb46b` merged main and committed the result *with the conflict markers still in it*; `1b7f5900` then cleaned those markers up. Separately, `e63daaf0` merged the same two parents cleanly on its own. `d1509fef` then merged the two lines of resolution together.

The decisive fact is that `git diff e63daaf0 d1509fef` is **empty** — the final tree is byte-identical to the independent clean merge. The `1b7f5900` line contributed nothing to what ships. I therefore reviewed `e63daaf0`'s resolution directly against both of its parents, which is the resolution that actually matters, and confirmed `1b7f5900` contained no feature work of its own: its entire diff is the removal of `<<<<<<< HEAD` / `>>>>>>> origin/main` blocks plus the same renumbering decisions. Nothing new was introduced under cover of a resolution commit.

I also confirmed the shipped tree carries no leftover markers anywhere:

```
$ git grep -n -E '^(<<<<<<< |=======$|>>>>>>> )' HEAD -- .
(no output)
```

</details>

<details><summary>Finding 2 — an intermediate commit in history does not compile</summary>

`5b1cb46b` ("chore: placeholder for merge conflict resolution plan") was pushed with literal conflict markers inside `agent.proto`, `errors.go`, `params.go`, `proto_layout_test.go` and both generated `.pb.go` files. That commit's tree cannot build.

This costs nothing at merge time — the head is clean, and every later commit supersedes it. The only real consequence is `git bisect`: if this branch lands as a merge commit rather than a squash, anyone bisecting through that region hits a commit that fails to compile and has to skip it manually. If the repository squash-merges, the problem evaporates entirely.

I am raising it as an observation, not a request. Cleaning it up would mean rewriting branch history, which trades a small future annoyance for a force-push on an approved branch — not a trade I would push the author into. Worth knowing about, worth a squash-merge if that is already the house style.

</details>

<details><summary>Finding 3 — the migration now seeds the sibling's params too, which is the one place the resolution added behavior</summary>

This is the most interesting thing in the merge and the part I looked at hardest.

**The situation each side arrived in.** This PR introduced the module's first store migration: it bumped `ConsensusVersion` from 1 to 2 and registered a 1→2 migration that seeds the new recipient-allowlist cap. The sibling ERC-8004 PR added three new params — a validation request fee, a tag byte cap and a URI byte cap — but left `ConsensusVersion` at **1** and shipped **no migration at all** (there is no `migrations.go` on main; its `ConsensusVersion` still returns 1). Presumably that was fine on its own terms, because a module that has never been live gets its params from `InitGenesis` rather than from a migration.

**Why that combination is dangerous if merged naively.** Once this PR lands, the module is at version 2 and this migration becomes the *only* 1→2 path. If it seeded only the allowlist cap, any chain that did upgrade through it would come out the other side with a zero-valued validation fee, tag cap and URI cap — and `Params.Validate` rejects every one of those. The merged module would own a migration that produces params its own validator refuses.

**What the resolution did.** It widened `Migrate1to2` to seed all four fields, guarding each of the sibling's three with an is-it-unset check so an operator-chosen value can never be clobbered, while setting the allowlist cap unconditionally — which is safe precisely because that field cannot have held any value at version 1:

```go
defaults := types.DefaultParams()
if params.ValidationRequestFee.Denom == "" { params.ValidationRequestFee = defaults.ValidationRequestFee }
if params.ValidationTagMaxBytes == 0 { params.ValidationTagMaxBytes = defaults.ValidationTagMaxBytes }
if params.ValidationUriMaxBytes == 0 { params.ValidationUriMaxBytes = defaults.ValidationUriMaxBytes }
params.SpendRecipientAllowlistMax = types.DefaultSpendRecipientAllowlistMax
```

The keeper test was renamed from `TestMigrate1to2_DefaultsRecipientAllowlistMax` to `TestMigrate1to2_DefaultsVersion2Params` and extended to zero out all four fields before migrating and assert all four come back at their defaults. There is exactly one `RegisterMigration(ModuleName, 1, …)` and one `ConsensusVersion() == 2` — no duplicate registration, which would have panicked the node at startup had both sides brought their own 1→2 migration.

**My judgement.** This is correct integration rather than smuggling. The reason I am calling it out anyway is that it is the single place where the resolution produced behavior neither parent had, so it should be a deliberate, visible decision rather than something a reader assumes was mechanical. If the author disagrees with owning the sibling's params in this migration, the alternative is for the sibling to bump to 2 with its own migration and for this PR to go to 3 — but that is strictly more moving parts for the same outcome, and I would not ask for it.

</details>

<details><summary>Finding 4 — buf-break-check is red for an infrastructure reason, and the proto delta is not breaking</summary>

`buf-break-check` reports failure on the merged head after 5 seconds. The log shows it never got as far as comparing anything:

```
Run bufbuild/buf-setup-action@v1.50.0
##[warning]No github_token supplied, API requests will be subject to stricter rate limiting
Setting up buf version "1.50.0"
Resolving the download URL for the current platform...
##[error]API rate limit exceeded for 172.190.93.133. … - https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting
```

The action failed while resolving the buf download URL because the runner had no token and hit the unauthenticated GitHub API rate limit. No proto file was ever examined. This is the same check that passed at the approved commit and again at `2001c2e6`; nothing about the schema caused it.

Because the check that would have caught a real break did not run, I verified the breaking-change question by hand against main as the baseline, message by message:

- `Params` — main holds fields 1 through 8, where 6, 7 and 8 are the sibling's validation params. HEAD holds those same 1 through 8 unchanged and adds 9 for `spend_recipient_allowlist_max`. Adding a new field number is not a break. The 6→9 move renumbered a field that existed only on this branch and never on main, so from main's perspective nothing moved.
- `Agent` — main stops at field 12; HEAD adds 13.
- `MsgUpdateAgentSpendPolicy` — main stops at field 5; HEAD adds 6.
- `EventUpdateAgentSpendPolicy` — main stops at field 4; HEAD adds 5.

No field is deleted, renamed, retyped or renumbered relative to main. There is no breaking change here for buf to find.

**What this needs:** a re-run of the job, not a code change. I am not making it a blocking finding, but it is a required check and nobody should merge on a red one — the manager or author should kick it and confirm green. (I hit the shared GitHub API rate limit myself immediately after pulling this log, which is consistent with the runner's own failure mode.)

</details>

<details><summary>Finding 5 — the renumbered generated codec, and how thoroughly it is actually covered</summary>

The highest-risk artifact in any merge like this is generated protobuf code that someone hand-resolved, because the struct tag, the marshaller, the size calculator and the unmarshaller are four independent places that must agree about a field number, and a hand-merge can leave them disagreeing. Both `agent.pb.go` and `tx.pb.go` were touched by the conflict resolution, so I checked all four sites for the renumbered field:

- struct tag: `protobuf:"varint,9,opt,name=spend_recipient_allowlist_max,…"`
- marshal: writes `dAtA[i] = 0x48`, which is `(9 << 3) | 0` — field 9, varint wire type
- size: `n += 1 + sovAgent(…)`, correct since a field-9 tag is one byte
- unmarshal: `case 9:` with `if wireType != 0`

All four agree, and field 6 in the same message now decodes `ValidationRequestFee` at wire type 2, as it should. The layout test's contract map was updated to `varint,9` and now lists all nine `Params` fields, which matters because that test asserts the map length equals the struct's field count — it would fail loudly if the merge had dropped or added a field.

The one gap worth mentioning: the generated-code round-trip test still constructs `Params` without `SpendRecipientAllowlistMax` and `Agent` without `SpendRecipientAllowlist`, so neither new field's codec is exercised by that particular test. In practice they are covered anyway — the migration test writes and reads back the cap through the real params store, and the allowlist persistence test writes and reads back an agent through real collections state, both of which are genuine marshal/unmarshal round trips. Populating them in the round-trip table would make the coverage explicit rather than incidental. Optional.

</details>

---

**Verification of the merged result** (run locally against the head commit `d1509fef`, scoped to the affected module plus the app wiring a merge could plausibly break, rather than the full repository suites that CI owns):

```
$ go build ./x/agent/... ./app/...
BUILD_EXIT=0

$ go test -count=1 ./x/agent/...
ok  github.com/dymensionxyz/dymension/v3/x/agent               0.107s
ok  github.com/dymensionxyz/dymension/v3/x/agent/client/cli    0.092s
ok  github.com/dymensionxyz/dymension/v3/x/agent/keeper        2.171s
ok  github.com/dymensionxyz/dymension/v3/x/agent/types         0.124s

$ golangci-lint run ./x/agent/...        # v2.1.6, matching CI
0 issues.

$ gofumpt -l <all PR-changed .go files>
x/agent/types/agent.pb.go
x/agent/types/tx.pb.go
```

The two gofumpt hits are generated files and are pre-existing, not merge damage — main's own `agent.pb.go` is flagged identically when checked in isolation. The repository's `golangci-lint` job passes on the head in CI as well.

One environment note for whoever runs this next: `golangci-lint` v2.1.6 panics inside `go/types` during package loading when run under Go 1.25; pinning to the Go 1.24.6 toolchain it was built against makes it complete cleanly. That is a linter/toolchain mismatch, nothing to do with this branch.

**Nothing dropped, nothing smuggled.** The delta from main to head consists of exactly the nineteen files of the approved feature — proto additions, the escrow handler gate, the CLI allowlist argument, address canonicalization, genesis and message validation, params, errors, the migration, and their tests — with `migrations.go` larger than at approval only by the params-seeding described in Finding 3. No unrelated file appears in the delta, and both sides' error codes (main's 17 through 23 preserved verbatim, this PR's moved to 24) and validation rules survive in full.

VERDICT: APPROVE

omriai2 (approved):
**APPROVE** — the single new commit is the one-line CI fix I asked for last round, it is correctly scoped (job-provisioned read-only token on a `pull_request`-triggered workflow, no new secret, no privilege surface), and the previously-red required check now passes in 1m38s having actually run the comparison rather than dying in setup.

## Prior findings — disposition

| Round | Finding | Status |
|---|---|---|
| 1 | *(blocking)* Allowlist unreachable from the CLI; CLI policy updates silently cleared it | **Resolved** at `2001c2e6` |
| 1 | Recipient comparison was raw-string, not address-canonical | **Resolved** at `2001c2e6` |
| 2 | Non-merge commit in the post-approval range | **Settled** — reviewed; its content never reached the shipped tree |
| 2 | Intermediate merge commit committed with conflict markers | Still true; non-blocking history hygiene, superseded by later commits |
| 2 | Migration widened to seed the sibling's validation params | Still true; correct integration, called out deliberately |
| 2 | `buf-break-check` red on runner API rate limit | **Resolved** by `1bb37d3f` — check now green |
| 2 | Round-trip proto test leaves the two new fields unpopulated | Still true; optional |

## Index — all findings non-blocking

1. *(observation)* The CI fix is correct and minimal, and its green run independently confirms the hand analysis I did last round when the check could not run — no proto break exists.
2. *(observation)* This PR now also edits a shared repository workflow, so its blast radius extends past `x/agent` to every PR's proto gate; worth knowing when reading the changelog.
3. *(nit)* The PR body's proof section was not updated to mention the CI change — not required, since the green required check is itself the receipt, but a line would keep the body a complete account of the diff.

---

<details><summary>Finding 1 — the CI fix is correct, minimally scoped, and its green run retires my prior finding</summary>

**What was wrong.** Last round `buf-break-check` failed on the merged head after five seconds. The log showed `bufbuild/buf-setup-action` warning `No github_token supplied, API requests will be subject to stricter rate limiting`, then erroring with `API rate limit exceeded` while still resolving the buf download URL. The job never examined a single proto file, so the red status carried no information about the schema. I flagged it as infrastructure needing a re-run rather than a code defect, and verified by hand that the proto delta against main adds field numbers only — `Params` 9, `Agent` 13, `MsgUpdateAgentSpendPolicy` 6, `EventUpdateAgentSpendPolicy` 5 — with nothing deleted, renamed, retyped, or renumbered relative to main.

**What the fix does.** One input added to the existing setup step, plus a trailing newline at end of file:

```yaml
       - uses: bufbuild/buf-setup-action@v1.50.0
+        with:
+          github_token: ${{ github.token }}
```

This is the documented remedy for exactly that failure: the action uses the GitHub API to resolve the buf release asset, and an authenticated request escapes the shared-IP unauthenticated ceiling that the runner was colliding with.

**Why the credential handling is sound.** I checked the surrounding workflow rather than the line in isolation, because adding a token to a workflow is where this class of change usually goes wrong:

- The trigger is `pull_request`, **not** `pull_request_target`. That is the distinction that matters. A fork PR under `pull_request` runs with a read-only token and no access to repository secrets, so there is no path by which untrusted branch content gains a privileged credential. Had this been `pull_request_target`, handing a token to a step in a fork-controlled checkout would be a real escalation surface.
- The workflow declares `permissions: contents: read` at the top level, so `github.token` is minimally scoped — strictly more than buf-setup needs, since it is only authenticating public-repo release lookups.
- `${{ github.token }}` is the automatically provisioned, job-lifetime `GITHUB_TOKEN`, not a stored PAT. Nothing was added to repository secrets, nothing long-lived was introduced, and the credential dies with the job.
- The token reaches only the setup action's declared `github_token` input. It is not exported to the environment, not echoed, and not handed to any step that executes repository-controlled code.
- The `against:` baseline expression is untouched; the only other change in the file is the missing terminal newline.

**Evidence that it worked.** The check is now green, and the duration is the tell:

```
buf-break-check   pass   1m38s   .../runs/31890354560/job/95025592502
```

Five seconds was setup dying. Ninety-eight seconds is buf downloaded, the base ref fetched, and the breaking-change comparison actually executed. So this is not merely a red badge turned green — the real check ran against the real baseline and agreed with the manual analysis I had to substitute for it last round. My prior Finding 4 is fully retired, and the proto renumbering now has authoritative rather than reasoned confirmation.

Under the external-dependency rule, adding an auth input to an existing pinned action at an unchanged endpoint is not a re-pointed dependency call, and in any case the live-invocation evidence exists in its strongest form: the job itself ran through the real credential path in CI on this exact head.

</details>

<details><summary>Finding 2 — the PR's blast radius now extends beyond x/agent</summary>

Everything in this PR up to the last round lived inside `x/agent` plus its protos. This commit edits `.github/workflows/proto.yaml`, which gates the proto checks for **every** pull request in the repository, not just this one.

I have no objection to it. The job was blocking this PR for an infrastructure reason I raised, the fix is one line, it is correct, and leaving a known-flaky required check unfixed so that the change stays "in scope" would be the worse call — the next PR to touch a proto file would hit the same coin flip. This is exactly the kind of adjacent fix that belongs with the change that surfaced it.

The reason I am naming it is legibility rather than correctness: someone skimming the PR title (`feat(agent): restrict attested transfer recipients`) and the body would not expect a repository-wide CI change inside it. If the project cares about that, the commit is cleanly separable — it touches one file and shares no code with the feature — and could land on its own. Entirely the author's call; I would not hold the merge either way.

</details>

<details><summary>Finding 3 — the proof section does not mention the CI change</summary>

The PR body's `## Proof of execution` section is unchanged from the version I checked in earlier rounds. Its shape is fine — every claim carries a receipt line naming the command or suite that produced it (full-repository race suite, build, `golangci-lint`, proto-regeneration determinism, focused `x/agent` run), plus an explicitly owned "Not verified" entry stating that a live GCP-attested network transaction could not be executed because a valid attestation token cannot be minted locally, with the covering keeper integration suite named. That entry is a proper declaration, not a gap.

The section simply predates the CI commit and says nothing about it. Under the proof rules that is not a defect: the claim "the buf setup step now authenticates and the check passes" is covered by a CI-gated required check whose green status is visible on the PR, and a named suite plus green CI *is* the proof — I should not be asking for a pasted transcript of it, and I am not.

I mention it only because the body otherwise reads as a complete account of what the diff does, and a reader reconciling the file list against the proof section will find one file unexplained. A single line noting the workflow fix and pointing at the now-green check would close that. Cosmetic.

</details>

---

**Verification.** I did not re-run the feature's build, lint or test suites this round: the code under `x/agent` is byte-identical to the tree I verified at `d1509fef` last round (`go build ./x/agent/... ./app/...` clean, all four `x/agent` test packages passing, `golangci-lint` v2.1.6 reporting `0 issues.`, `gofumpt` flagging only the pre-existing generated files), and this commit touches nothing but a YAML workflow. The workflow change is not exercisable locally in any meaningful way — its correctness is established by the job running green in the real Actions environment, which it has. `build`, `golangci-lint` and `Analyze` are still in flight on this head; those are the author's required gate and pending is not failure.

**Nothing dropped, nothing new smuggled.** The commit's diffstat is one file, three insertions, one deletion, and the full text of both hunks is reproduced above.

VERDICT: APPROVE

omriai2 (approved):
**APPROVE** — the new commit is a correct, minimal, and demonstrably working CI fix (codecov-action v3→v5 with the mandatory `file`→`files` rename plus OIDC auth, and a `permissions` block that *narrows* this workflow to exactly what its single job needs), and every check on the head is now green, including both codecov statuses.

## Prior findings — disposition

| Round | Finding | Status |
|---|---|---|
| 1 | *(blocking)* Allowlist unreachable from the CLI; CLI policy updates silently cleared it | **Resolved** at `2001c2e6` |
| 1 | Recipient comparison was raw-string, not address-canonical | **Resolved** at `2001c2e6` |
| 2 | Non-merge commit in the post-approval range | **Settled** — its content never reached the shipped tree |
| 2 | Intermediate merge commit pushed with conflict markers | Still true; non-blocking history hygiene, superseded |
| 2 | Migration widened to seed the sibling's validation params | Still true; correct integration, deliberately called out |
| 2 | `buf-break-check` red on runner API rate limit | **Resolved** at `1bb37d3f`; still green here (17s) |
| 2 | Round-trip proto test leaves the two new fields unpopulated | Still true; optional |
| 3 | CI fix broadened the PR past `x/agent` into a shared workflow | Still true, and now broader; non-blocking |
| 3 | Proof section not updated for the CI change | Still true; non-blocking, CI status is the receipt |

## Index — all findings non-blocking

1. *(observation)* The codecov upgrade is correct including the input rename that a v3→v5 bump silently requires, and both codecov statuses passing on this head is direct end-to-end evidence the OIDC credential path works.
2. *(observation)* The new `permissions` block tightens rather than loosens this workflow; I verified no step in it needed anything beyond `contents: read`.
3. *(observation)* Pull requests from forks are not granted `id-token: write`, so external-contributor PRs will likely stop getting coverage reported; this should degrade to a warning rather than redden the required `build` check, but the action's behavior on a failed OIDC mint is undocumented.
4. *(observation)* The PR now edits two shared repository workflows, so its blast radius reaches every PR in the repo.
5. *(nit)* Third-party actions remain on mutable major tags, and one of them now receives an OIDC identity token — pre-existing practice, not a regression here.

---

<details><summary>Finding 1 — the codecov change is correct, including the rename a version bump silently requires</summary>

The commit does four things in eight lines: bumps `codecov/codecov-action` from v3 to v5, renames the input `file:` to `files:`, adds `use_oidc: true`, and adds a workflow-level `permissions` block.

The input rename is the part most easily got wrong, and it is the part that matters most. The `file` input existed in v3; from v4 onward the action takes `files`. Had the bump landed without the rename, `file:` would have been an unrecognised input — the action would not have errored on it, it would simply have fallen back to auto-discovering coverage files, which either uploads the wrong artifact or silently uploads nothing while the step still reports success. That is the classic way a coverage pipeline goes quietly dead. The author renamed it correctly.

`use_oidc: true` replaces tokenless upload with an OIDC-authenticated one. Per Codecov's own documentation, `use_oidc` means "Use OIDC instead of token. This will ignore any token supplied", and the action requires `id-token: write`, which "can be set at either the workflow or job level" — so the placement chosen here is a documented, supported one rather than an improvisation.

**The decisive evidence is that it demonstrably works.** On this head:

```
build            pass  31m48s
codecov/patch    pass  1s     https://app.codecov.io/gh/dymensionxyz/dymension/pull/2197
codecov/project  pass  1s     https://app.codecov.io/gh/dymensionxyz/dymension/pull/2197
```

Both codecov statuses reporting means the coverage report was minted, uploaded, accepted and processed against the project baseline. A broken OIDC mint, a wrong file path, or an ignored input would all have left those two statuses absent or failing. This is the live invocation through the real deployed endpoint and the real credential path, executed in CI, which is exactly the evidence class the external-dependency rule asks for — and it is CI-visible rather than a pasted transcript.

</details>

<details><summary>Finding 2 — the permissions block narrows this workflow rather than widening it</summary>

Adding a `permissions:` key to a workflow that previously had none is the change in this diff that deserves the most care, because it cuts both ways. Before this commit, `test.yml` had no `permissions` block and therefore inherited whatever the repository or organisation default is — which in many repositories is broad. Declaring `contents: read` and `id-token: write` sets the token to *exactly* those two scopes and drops everything else to none. That is a tightening, but it also means any step that had been quietly relying on a broader default would break.

So I read the whole file rather than the hunk. It is 691 bytes and contains a single job with five steps: `actions/checkout`, `actions/setup-go`, `go build`, the `go-acc` coverage run, and the codecov upload. Nothing writes to contents, pull requests, checks, packages, deployments, or issues. `contents: read` covers the checkout; `id-token: write` covers the OIDC mint. There is no step left under-provisioned.

The single-job structure also settles the usual objection to workflow-level `id-token: write` — that it grants identity-token minting to jobs which do not need it. Here there is exactly one job, and it is the one that needs it, so workflow-level and job-level scoping are equivalent in effect. If a second job is ever added to this file, moving the `id-token: write` down under `build` would become worth doing; today it would be a distinction without a difference.

</details>

<details><summary>Finding 3 — fork pull requests will likely lose coverage reporting</summary>

This is the one consequence of the change that reaches beyond this PR, so I want it on the record even though I would not hold the merge for it.

The workflow triggers on `push` (tags and `main`) and on `pull_request`. It is **not** `pull_request_target`, which is the right choice and means there is no privilege-escalation surface here: a fork PR under `pull_request` runs against untrusted branch content with no access to repository secrets. But the same restriction cuts the other way for OIDC. GitHub does not grant `id-token: write` to workflows triggered by `pull_request` from a forked repository, regardless of what the `permissions` block requests. On such a run the OIDC mint has nothing to mint from.

The likely outcome is degraded coverage reporting rather than a broken gate. Codecov's documentation confirms `fail_ci_if_error` defaults to `false` — the usage example carries the comment `fail_ci_if_error: true # optional (default = false)` — so an upload error is expected to surface as a warning and leave the step successful, which keeps the required `build` check green for external contributors while they simply get no coverage comment.

I want to be straight about the limit of that reasoning: Codecov's documentation states the `id-token: write` requirement but says nothing about what the action does when the token cannot be obtained — whether that path is treated as an upload error subject to `fail_ci_if_error`, or as a hard setup failure that bypasses it. I did not chase it into the action source, because settling it would not change my verdict either way: if it degrades gracefully, this is a cosmetic loss for fork PRs; if it hard-fails, it is a CI-config follow-up on a repository that, from the branch naming throughout this PR's history, appears to work predominantly from in-repo branches rather than forks. Either way it is not a defect in this change and not something to send the author back for. It is worth someone watching the next external contribution.

</details>

<details><summary>Finding 4 — the PR's blast radius now covers two shared workflows</summary>

Last round I noted that the PR had grown past `x/agent` to touch `.github/workflows/proto.yaml`. It now also touches `.github/workflows/test.yml`. Both are repository-wide: the first gates proto checks on every PR, the second gates build, test and coverage on every PR and every push to `main` and to tags.

My position is unchanged and unenthusiastic about objecting. Both changes fix required checks that were failing for infrastructure reasons on this very PR, both are minimal and correct, and deferring them would leave known-broken gates in place for whoever pushes next. This is the right instinct.

What I would flag for the human reader is legibility rather than correctness. A PR titled `feat(agent): restrict attested transfer recipients` now contains two commits that change how every pull request in the repository is validated, and neither is mentioned in the PR body. Both commits are cleanly separable — each touches exactly one YAML file and shares no code with the feature — so if the project prefers CI changes to land on their own, extracting them is trivial. That is a process preference, not a review finding, and I leave it to the author.

</details>

<details><summary>Finding 5 — third-party actions on mutable tags, one of which now holds an identity token</summary>

Every third-party action in both touched workflows is referenced by mutable major tag — `codecov/codecov-action@v5`, `bufbuild/buf-setup-action@v1.50.0`, `bufbuild/buf-breaking-action@v1.1.4`, `actions/checkout@v6`, `actions/setup-go@v5`. A major tag is repointable by its owner, so the code that runs in CI is not what was reviewed here.

This is pre-existing repository practice, not something this PR introduced, and I am explicitly not asking for a SHA-pinning sweep in a feature PR. The reason it is worth one line now rather than never is that the risk profile of one of those actions changed in this commit: `codecov-action` previously received a coverage file and now additionally receives an OIDC identity token scoped to this repository. That raises what a compromised or repointed tag could do from "exfiltrate a coverage report" to "hold a short-lived, repository-scoped identity assertion". The `id-token: write` scope grants nothing on GitHub itself — it only lets the job mint assertions for external verifiers — so the exposure is bounded, and Codecov is a first-party-grade dependency this project already trusts with its coverage pipeline.

If the project ever does a supply-chain pass on its workflows, this action is the one to pin first. Not now, and not in this PR.

</details>

---

**Verification.** I ran nothing locally this round. The product code under `x/agent` is byte-identical to the tree I verified at `d1509fef` — where `go build ./x/agent/... ./app/...` succeeded, all four `x/agent` test packages passed, `golangci-lint` v2.1.6 reported `0 issues.`, and `gofumpt` flagged only the pre-existing generated files — and this commit changes eight lines of YAML in one workflow. A workflow change is not meaningfully exercisable outside Actions, and it has now executed for real: `build` completed in 31m48s with both codecov statuses reporting, which is stronger evidence than anything I could produce here.

Every check on the head commit is green: `build`, `golangci-lint`, `Analyze`, `CodeQL`, `buf-break-check`, `label-prs`, `codecov/patch`, `codecov/project`. `protobuf-checks` shows as skipping, which is its designed behavior — that job is label-gated behind `enable-proto-checks` and has been for the life of this PR.

**Nothing dropped, nothing smuggled.** The commit's diffstat is one file, six insertions, two deletions, and the complete text of both hunks is reproduced in Finding 1's analysis above.

VERDICT: APPROVE

PR #2199, opened 2026-08-15
feat(agent): add ERC-8004 validation registry

### LINKED ISSUE #2198: x/agent: ERC-8004 Validation Registry — independent request→response validation of attested agent actions (the missing third registry)
## Problem / opportunity

`x/agent` implements two of ERC‑8004's three trust registries but not the third:

- **Identity Registry** — `RegisterAgent`/`DeactivateAgent`, `x/agent/keeper/registry.go` (#2159).
- **Reputation Registry** — `SubmitFeedback`/`RevokeFeedback`, `x/agent/keeper/msg_server_reputation.go` (#2186). This is **subjective client feedback**: one score per `(agent_id, client)`, aggregated into `Reputation{Count, ScoreSum}`.

The **Validation Registry** — independent, third‑party *validation* of a specific piece of agent work via a request→response flow — does not exist anywhere in the module. `grep -ri validation x/agent` returns only an unrelated genesis test string. Store‑key space stops at `KeyAgentEscrow = 0x06` (`x/agent/types/keys.go`); there is no validation prefix, no `MsgRequestValidation`/`MsgRespondValidation` in `proto/dymensionxyz/dymension/agent/tx.proto`, and no validation queries.

Reputation answers *"do clients like this agent?"*. Validation answers a different, objective question: *"did an independent validator check this specific action and does it pass?"* The two are complementary and both are needed for a trustless cross‑organizational agent economy. This is the missing pillar.

## Why now

ERC‑8004 ("Trustless Agents", proposed 2025‑08‑13) defines exactly three registries — Identity, Reputation, and **Validation** — as the standard trust layer for autonomous agents. Reference implementations went live on Base Sepolia, Linea Sepolia and Hedera through late‑2025/early‑2026. The Validation Registry is the spec's mechanism for **crypto‑economic / re‑execution / TEE‑attested verification** of agent output, distinct from social reputation. Dymension already markets its `x/agent` stack as an ERC‑8004 implementation (see #2171, #2186); shipping only 2 of 3 registries leaves the trust model incomplete precisely as the standard gains adoption.

- ERC‑8004 EIP: https://eips.ethereum.org/EIPS/eip-8004
- Registry contracts (curated by 8004 team): https://github.com/erc-8004/erc-8004-contracts
- Explainer (three registries incl. Validation): https://learn.backpack.exchange/articles/erc-8004-explained
- Awesome‑ERC8004: https://github.com/sudeepb02/awesome-erc8004

ERC‑8004 Validation Registry shape (mapped below): `validationRequest(validator, agentId, requestUri, requestHash)` → `validationResponse(requestHash, response[0..100], responseUri, responseHash, tag)`; response MUST come from the named validator; may be called multiple times for the same `requestHash` (progressive finality).

## Proposed change

Add a third registry to `x/agent`, faithfully mapping the ERC‑8004 Validation Registry onto Dymension's existing agent identities and **attested action log**.

### Proto — `proto/dymensionxyz/dymension/agent/`

New file `validation.proto` (package/option style mirroring `feedback.proto`):

```proto
message ValidationRequest {
  bytes  request_hash  = 1; // 32-byte commitment; primary key
  string requester     = 2; // bech32; whoever asked for validation
  string validator_id  = 3; // registered agent id that must respond
  string agent_id      = 4; // subject agent id being validated
  uint64 evidence_seq  = 5; // action-log seq of the subject action under validation
  string request_uri   = 6; // optional off-chain pointer, bounded
  int64  height        = 7;
  google.protobuf.Timestamp time = 8 [(gogoproto.stdtime)=true,(gogoproto.nullable)=false];
  uint64 response_count = 9;
}

message ValidationResponse {
  bytes  request_hash  = 1;
  uint64 seq           = 2; // 0-based index of this response under the request
  uint32 response      = 3; // 0..100
  string response_uri  = 4; // optional, bounded
  bytes  response_hash = 5; // optional; if set, exactly 32 bytes
  string tag           = 6; // optional, bounded by ValidationTagMaxBytes
  int64  height        = 7;
  google.protobuf.Timestamp time = 8 [(gogoproto.stdtime)=true,(gogoproto.nullable)=false];
}
```

`tx.proto` — two RPCs + messages on the existing `Msg` service:

```proto
rpc RequestValidation(MsgRequestValidation) returns (MsgRequestValidationResponse);
rpc RespondValidation(MsgRespondValidation) returns (MsgRespondValidationResponse);

message MsgRequestValidation {
  string requester    = 1; // signer
  string validator_id = 2;
  string agent_id     = 3;
  uint64 evidence_seq = 4;
  bytes  request_hash = 5;
  string request_uri  = 6;
}
message MsgRequestValidationResponse {}

message MsgRespondValidation {
  string responder    = 1; // signer; MUST be owner of validator_id
  bytes  request_hash = 2;
  uint32 response     = 3; // 0..100
  string response_uri = 4;
  bytes  response_hash = 5;
  string tag          = 6;
}
message MsgRespondValidationResponse { uint64 seq = 1; }
```

`query.proto` — add `ValidationRequest(request_hash)`, `ValidationResponses(request_hash, pagination)`, `ValidationRequestsByAgent(agent_id, pagination)`.

`genesis.proto` — add `repeated ValidationRequest validation_requests` and `repeated ValidationResponse validation_responses`.

### Store — `x/agent/types/keys.go`, `x/agent/keeper/keeper.go`

Add prefixes:
```
KeyValidationRequests  = []byte{0x07}
KeyValidationResponses = []byte{0x08}
KeyValidationByAgent   = []byte{0x09}
```
Collections wired in `NewKeeper`:
- `validationRequests collections.Map[[]byte, types.ValidationRequest]` (key `collections.BytesKey`).
- `validationResponses collections.Map[collections.Pair[[]byte, uint64], types.ValidationResponse]` (key `PairKeyCodec(BytesKey, Uint64Key)`).
- `validationByAgent collections.KeySet[collections.Pair[string, []byte]]` — secondary index `(agent_id, request_hash)` powering `ValidationRequestsByAgent`.

### Handlers — new `x/agent/keeper/msg_server_validation.go`

`RequestValidation` (mirrors the structure of `SubmitFeedback`):
1. `msg.ValidateBasic()`.
2. Load params. Enforce `len(request_uri) <= ValidationUriMaxBytes`; `len(request_hash) == 32` else `ErrInvalidValidationHash`.
3. Both `validator_id` and `agent_id` must exist (`GetAgent`) and be `Active`; else `ErrAgentNotFound` / `ErrValidatorInactive`.
4. `validator_id != agent_id` else `ErrSelfValidation`.
5. `evidence_seq < subjectAgent.ActionSeq` else `ErrInvalidEvidence` (reuses the exact append‑only proof pattern from `SubmitFeedback`; validation must reference a real logged action).
6. `request_hash` must be unrolled/unique — `Has(request_hash)==false` else `ErrValidationRequestExists`.
7. Charge anti‑spam `ValidationRequestFee`: `SendCoinsFromAccountToModule` then `BurnCoins` (identical to feedback fee), guarded by `!IsNil() && !IsZero()`.
8. Persist `ValidationRequest{...,ResponseCount:0}`, add `validationByAgent(agent_id, request_hash)`, emit `EventRequestValidation`.

`RespondValidation`:
1. `msg.ValidateBasic()` (includes `response <= 100`, `request_hash` len 32, optional `response_hash` len 32 when non‑empty, `tag`/`uri` bounds).
2. Load request by `request_hash`; missing → `ErrValidationRequestNotFound`.
3. Re‑`GetAgent(validator_id)`; require `found && Active`; require `msg.Responder == validator.Owner` else `ErrUnauthorized` (only the named validator may respond — ERC‑8004 MUST).
4. Append `ValidationResponse{Seq: req.ResponseCount, ...}`; `req.ResponseCount++`; re‑`Set` request (progressive finality: multiple responses allowed).
5. Emit `EventRespondValidation`; return `seq`.

### Params — `x/agent/types/params.go`

Add three fields with defaults:
- `ValidationRequestFee sdk.Coin` → default `commontypes.DYMCoin` (matches `FeedbackFee`).
- `ValidationTagMaxBytes uint64` → default `32` (matches `DefaultFeedbackTagMaxBytes`).
- `ValidationUriMaxBytes uint64` → default `512`.
Extend `Params.Validate()` accordingly (coin `.Validate()`, non‑zero max‑bytes).

### Errors — `x/agent/types/errors.go` (next free codes, 17+)

`ErrValidationRequestExists(17)`, `ErrValidationRequestNotFound(18)`, `ErrInvalidValidationHash(19)`, `ErrInvalidValidationResponse(20)`, `ErrSelfValidation(21)`, `ErrValidatorInactive(22)`, `ErrValidationFeePayment(23)`.

### Events — `proto/.../events.proto`

`EventRequestValidation{request_hash, validator_id, agent_id, evidence_seq}`, `EventRespondValidation{request_hash, validator_id, agent_id, response, seq}`.

### Genesis — `x/agent/keeper/genesis.go`

Import `validation_requests` verbatim and rebuild the `validationByAgent` index from them (mirroring how reputation is rebuilt from feedback — index never stored in genesis, always derived, so it cannot drift). Import `validation_responses` verbatim. Export both. Add a validity check in `GenesisState.Validate()` that every response's `request_hash` matches an included request.

### CLI — `x/agent/client/cli/`

`tx_request_validation.go`, `tx_respond_validation.go` (hash args accept hex), and query subcommands, wired into `tx.go`/`query.go`.

### Invariant — `x/agent/keeper/invariants.go`

Add `validation-response-integrity`: every `ValidationResponse` has a matching `ValidationRequest`, and each request's `ResponseCount` equals the number of stored responses for its hash. Follow the existing invariant registration pattern in that file.

- **Distinct from reputation, not merged.** Separate store, messages, events. Reputation stays one‑score‑per‑client; validation is request→response by a named validator. No change to `msg_server_reputation.go`.
- **Validator = a registered, Active agent; responder must be its `Owner`.** Reuses the identity registry; no new actor type. Enforces ERC‑8004's "MUST be called by the validator".
- **Every request references a real action.** `evidence_seq < subjectAgent.ActionSeq`, reusing the append‑only correctness argument already documented in `SubmitFeedback`. This is a deliberate Dymension‑specific tightening over ERC‑8004 (which uses only an off‑chain `requestUri`): validation is anchored to the on‑chain attested action log. `request_uri`/`request_hash` still commit to the off‑chain artifact.
- **`request_hash` is the primary key and must be globally unique** (one open request per commitment). `response_hash` optional; if present must be 32 bytes.
- **`response` is `uint32` constrained to `0..100`** — matches ERC‑8004 (0=fail, 100=pass, intermediate allowed). Rejected values error, not clamp.
- **Multiple responses per request allowed** (progressive finality), each with monotonically increasing `seq = ResponseCount`. No response is ever overwritten or deleted (append‑only, like the action log).
- **Anti‑spam:** `ValidationRequestFee` burned exactly like `FeedbackFee`. Responses are free (the validator is doing work; ERC‑8004 makes only request the initiator‑paid step) — no fee on respond.
- **No cross‑module wiring in this issue.** Validation scores are readable via query/events; feeding them into eIBC LP selection, reputation, or slashing is explicitly out of scope (separate follow‑ups).
- **Naming:** `RequestValidation`/`RespondValidation`, matching the module's verb‑first Msg naming (`SubmitFeedback`, `RevokePolicy`).

## Consensus‑safety

- Pure state reads/writes over `cosmossdk.io/collections`; identical on every validator. No wall‑clock (only `ctx.BlockTime()`/`ctx.BlockHeight()`), no randomness, no floating point.
- No map iteration in state‑mutating paths; the `validationByAgent` index is only *range‑iterated in gRPC queries*, which are off‑consensus. Handlers touch fixed keys only.
- No Begin/EndBlocker logic added — nothing to panic a block. All validation is tx‑driven and therefore atomic: any error rolls the tx back (fee included).
- Byte‑key uniqueness + `evidence_seq` bound are deterministic guards; fee burn mirrors an existing consensus‑safe path.

## Acceptance criteria

- `make proto-gen` regenerates `validation.pb.go`, `tx.pb.go`, `query.pb.go`, `events.pb.go`, `genesis.pb.go` cleanly; `proto_layout_test.go` extended for the new `Params` fields passes.
- New `x/agent/keeper/msg_server_validation_test.go` covers, at minimum:
  - happy path: request then respond; response stored with `seq=0`, `ResponseCount=1`; query returns it.
  - reject `request_hash` not 32 bytes; reject duplicate `request_hash`.
  - reject `evidence_seq >= subjectAgent.ActionSeq`.
  - reject `validator_id`/`agent_id` not found or inactive; reject `validator_id == agent_id`.
  - reject respond by a non‑owner of the validator; reject `response > 100`.
  - progressive finality: two responses under one hash get `seq` 0 then 1, `ResponseCount==2`.
  - fee burned on request; tx atomically reverts fee on a later validation error.
- `grpc_query_test.go` covers `ValidationRequest`, `ValidationResponses`, `ValidationRequestsByAgent` (pagination).
- `genesis_test.go`: export→import round‑trips requests+responses; `validationByAgent` index rebuilt; `GenesisState.Validate()` rejects a response whose `request_hash` has no request.
- `invariants` test: integrity invariant passes on valid state and trips on a manually orphaned response.
- `go test ./x/agent/...` green; `golangci-lint run` and `gofumpt -w .` clean on touched files.

## Out of scope

- Any consumer of validation results (eIBC LP priority, reputation blending, sequencer/slashing hooks, bridging‑fee logic). Read‑only surface only.
- TEE‑attested validation responses (reusing `x/common/tee` to force the validator to attest its verdict) — a natural v2, not this issue.
- Changing, merging, or migrating the existing Reputation or Identity registries.
- Off‑chain validator software / A2A transport.

---

---

omriai1:
Closes #2198

Adds the missing ERC-8004 Validation Registry to x/agent. Requests are anchored to attested action-log entries and responses remain append-only, while both validator and subject must stay active, policy-live, and independently owned throughout the request→response flow.

## Changes

- Adds protobuf transactions, queries, events, params, genesis fields, and generated bindings.
- Adds collections storage with a derived per-agent index, deterministic genesis handling, and response-integrity invariant.
- Adds fee-backed request/respond handlers, transaction CLI commands, reachable AutoCLI query commands, pagination, and acceptance coverage.

- Request hashes are globally unique 32-byte keys and subject evidence must satisfy `evidence_seq < ActionSeq`.
- Validators and subjects are re-checked against the governance policy-revocation denylist when requesting and responding.
- Validator and subject IDs and owners must differ, preventing same-account self-validation through multiple registrations.
- Responses are 0..100, append-only, and restricted to the named validator owner.
- Genesis rebuilds the agent index and rejects orphaned, duplicate, non-contiguous, or count-mismatched responses.

</details>

## Proof of execution

- Build and Test — CI re-runs the repository suite on the updated branch.
- golangci-lint — scoped fix-round run passed with `0 issues`; CI re-runs this suite.
- Validation trust-boundary regressions — `go test ./x/agent/keeper -run Validation -count=1` passed, covering policy revocation for both roles and phases, cross-ID same-owner rejection, genesis count/sequence integrity, fees, queries, and invariants.
- Validation types — `go test ./x/agent/types -run 'Genesis|Validation' -count=1` passed.
- AutoCLI reachability — `go test ./x/agent -run TestAutoCLIValidationQueries -count=1` passed for all three validation query RPC descriptors.

**Not verified:** live chain transaction smoke; no external dependency was added, and the repository harness covers the changed consensus and CLI descriptor behavior.


omriai2 (changes requested):
## VERDICT: REQUEST_CHANGES

The registry itself is well-shaped — collections storage, an integrity invariant, genesis round-trip, and a committed acceptance suite are all present and the fee-burn path correctly mirrors the module's existing registration/feedback pattern — but the two new handlers skip the module's governance revocation denylist, enforce self-dealing one level shallower than the sibling feedback subsystem already does, ship a genesis validator that cannot uphold the invariant this same PR introduces, and add ~70 lines of query CLI that nothing in the binary can reach.

### Findings index

1. **Policy-revocation denylist does not reach the validation registry** — both new handlers gate on `Active` alone, so a governance-revoked agent can still issue and receive validation verdicts. *Accept when:* both handlers use the module's effective-liveness helper for validator and subject, with a test proving a revoked fingerprint blocks request and response.
2. **Self-validation is blocked at agent-ID level, not owner level** — strictly weaker than the `ErrSelfFeedback` rule the same module already enforces, and one owner may register unlimited agents. *Accept when:* `RequestValidation` rejects `validator.Owner == subject.Owner`, re-checked in `RespondValidation`, with a test.
3. **Genesis validation cannot uphold the new invariant** — nothing checks `ResponseCount` against the imported responses, so a genesis that passes `ValidateGenesis` can halt the chain via the crisis module and can silently overwrite an "append-only" response. *Accept when:* `GenesisState.Validate` enforces per-request response count and seq contiguity, with a test asserting rejection.
4. **The three new query CLI commands are unreachable dead code** — `x/agent` exposes no `GetQueryCmd`, and nothing in the repo calls `BasicManager.AddQueryCommands`. Same class: `SetValidationRequest` is defined and never called. *Accept when:* the commands are wired into the module (or deleted in favour of AutoCLI), and the unused keeper setter is removed.

Non-blocking observations are in fold 5.

Proof of execution and regression coverage both pass shape: every claim carries a named command, the one unverifiable claim is explicitly owned as "Not verified", and the PR commits behavioural tests for the new handlers. Those claims are CI-covered (`golangci-lint` and `buf-break-check` already green; `build` still running), so I did not open transcripts. No external dependency call is added, so no live-invocation proof is required.

<details><summary>Finding 1 — Policy-revocation denylist does not reach the validation registry</summary>

**Class.** Both new handlers, at three call sites, and for both roles.

This module has an explicit, documented notion of *effective liveness* that is strictly stronger than the `Active` flag: an agent may act only if it is active **and** its TEE policy fingerprint is not in the governance-managed revocation denylist. The shared helper says so in its own doc comment:

```go
// IsAgentLive reports whether an agent may act on-chain. Keep this effective
// policy check aligned with SubmitAttestedAction and the Agent query.
func (k Keeper) IsAgentLive(ctx sdk.Context, agentID string) bool {
	agent, err := k.agents.Get(ctx, agentID)
	if err != nil || !agent.Active {
		return false
	}
	fp, err := types.PolicyFingerprint(agent.EffectivePolicy(ctx.BlockHeight()))
	...
	return !revoked
}
```

and the attested-action loader repeats the rule:

```go
	agent.PromotePendingPolicy(ctx.BlockHeight())
	// The denylist applies to every attested operation, including transfers.
	if _, err := k.fingerprintNotRevoked(ctx, agent.Policy); err != nil {
		return types.Agent{}, err
	}
```

The new handlers use neither. `RequestValidation` checks `!validator.Active` and `!agent.Active`; `RespondValidation` checks `!found || !validator.Active`. Nothing consults `IsPolicyRevoked`, and nothing considers the effective (pending-promoted) policy.

**Consequence.** `MsgRevokePolicy` is the module's kill switch for a compromised enclave image. Today it instantly cuts an affected agent off from attested actions and escrow transfers. After this PR it does not stop that same agent from being named validator and signing `MsgRespondValidation`, nor from being recorded as a validation subject. The verdicts it emits are indistinguishable on-chain from a live, trusted validator's. Because deactivation (`MsgDeactivateAgent`) is owner-initiated and the owner is exactly the party benefiting from staying live, `Active` is not a substitute control for revocation.

This is not a hypothetical class for this repo — the commit immediately preceding this branch is `fix(eibc): honor agent policy revocation (#2193)`, the same omission in a different consumer.

**Acceptance criterion.** `RequestValidation` gates both the validator agent and the subject agent through `IsAgentLive` (or `fingerprintNotRevoked` on the effective policy), and `RespondValidation` re-gates the validator the same way at response time, since revocation can land between request and response. Add a test that revokes the validator's fingerprint and asserts both `RequestValidation` and `RespondValidation` fail, and a second asserting a revoked *subject* cannot be validated.

</details>

<details><summary>Finding 2 — Self-validation is blocked at agent-ID level, not owner level</summary>

**Class.** The single anti-self-dealing control in the new registry.

The whole guard is an ID comparison:

```go
	if msg.ValidatorId == msg.AgentId {
		return nil, types.ErrSelfValidation
	}
```

The authorization to *write* the verdict, however, is at the account level — `RespondValidation` accepts any signer equal to `validator.Owner`. Nowhere in the request or response path is the validator agent's owner compared with the subject agent's owner.

The sibling reputation subsystem in this same module already decided that the owner account is the security boundary, not the agent ID:

```go
	if msg.Client == agent.Owner {
		return nil, errorsmod.Wrap(types.ErrSelfFeedback, msg.Client)
	}
```

And `RegisterAgent` places no limit on how many agent IDs one owner may hold — it only rejects a duplicate ID. A validator agent also needs no attestation history to be usable: only the *subject* needs real actions, because of the `EvidenceSeq >= agent.ActionSeq` check. So a single owner registers a throwaway validator alongside their genuine agent and self-issues 100/100 "independent" verdicts against their own attested actions, at the cost of one extra registration fee.

**Honest scoping.** An owner-level check does not eliminate collusion — an attacker with two funded accounts can still do this, and any open validation registry has that property; ERC-8004 assumes consumers weigh validator reputation. The finding is that the current check is *strictly weaker than what this module already enforces one file over*, and the cheapest self-dealing path (one account, two agents) is left wide open while the identical path is closed for feedback.

**Acceptance criterion.** `RequestValidation` loads both agents and rejects when `validator.Owner == subject.Owner`, with a distinct error (reuse `ErrSelfValidation` or add a sibling). Re-check in `RespondValidation`, since agent ownership can change between request and response. Keep the ID comparison as a cheap short-circuit. Add a test registering two agents under one owner and asserting the request is rejected.

</details>

<details><summary>Finding 3 — Genesis validation cannot uphold the invariant this PR introduces</summary>

**Class.** `GenesisState.Validate` versus `InvariantValidationResponseIntegrity`, both added in this diff.

The new invariant asserts two things: every response references an existing request, and each request's stored `ResponseCount` equals the number of responses actually present:

```go
			if counts[string(hash)] != req.ResponseCount {
				return true, fmt.Errorf("validation response count mismatch: %x: stored %d actual %d", ...)
			}
```

The new genesis validator checks request-hash length, request-hash uniqueness, response→request existence, and response `(hash, seq)` uniqueness. It does **not** check `ResponseCount`, and it does not check that seq values are contiguous `0..count-1`.

Two consequences, both reachable from a genesis file that `ValidateGenesis` accepts:

1. **Chain halt.** The crisis module is wired into this app (`crisis.NewAppModule(app.CrisisKeeper, skipGenesisInvariants, ...)`). A genesis carrying one response with `ResponseCount: 0` starts a chain whose agent invariant is already broken; any account can then submit `MsgVerifyInvariant` and halt it. The gate that is supposed to prevent this — `ValidateGenesis` — passes that state.

2. **Silent overwrite of an "append-only" record.** `RespondValidation` derives the next slot purely from the counter:
   ```go
   	seq := req.ResponseCount
   ```
   Import responses at seq `0` and `1` with `ResponseCount: 1`, and the next response writes to `Join(hash, 1)`, destroying the existing seq-1 response. The PR body's stated property — "Responses are 0..100, append-only" — does not hold across a genesis boundary.

The module's own test demonstrates the shape of the problem: it constructs an invariant-breaking state by feeding `InitGenesis` a response with no matching request, precisely the case `Validate` *does* catch. The count case has no equivalent guard.

I have not raised this as attacker-controlled — genesis is authored by chain operators — but it is a data-integrity gate that this PR ships incomplete relative to the invariant it ships alongside, and the fix is a few lines in the loop that already exists.

**Acceptance criterion.** `GenesisState.Validate` builds a per-request-hash response tally and rejects when it disagrees with that request's `ResponseCount`, and rejects any response whose `Seq >= ResponseCount` (which also gives contiguity, given the existing uniqueness check). Add a test asserting a `ResponseCount` mismatch is rejected by `Validate`.

</details>

<details><summary>Finding 4 — New query CLI commands are unreachable dead code</summary>

**Class.** All three commands added to `x/agent/client/cli/query.go`, plus one unused keeper method.

`x/agent`'s module type exposes `GetTxCmd()` but no `GetQueryCmd()`. In cosmos-sdk v0.50, those are two independent optional interfaces — `BasicManager.AddQueryCommands` only picks up a module that implements `GetQueryCmd() *cobra.Command`:

```go
func (bm BasicManager) AddQueryCommands(rootQueryCmd *cobra.Command) {
	for _, b := range bm {
		if mod, ok := b.(interface {
			GetQueryCmd() *cobra.Command
		}); ok {
```

Verified three ways in this repo:

- `rg -rn "GetQueryCmd" x/agent/` returns only the definition itself — no caller.
- `rg -n "AddQueryCommands|AddTxCommands" --type go .` (excluding the module cache) returns **nothing**: the root command never uses the BasicManager path at all. `queryCommand()` hand-assembles its subcommands and lists only `forwardcli.GetQueryCmd()` among module queries.
- `x/agent/module.go` imports the `cli` package solely for `cli.GetTxCmd()`.

The agent module's queries are AutoCLI-generated instead. AutoCLI builds a command for *every* method on the service descriptor, using `RpcCommandOptions` only to customise specific ones, so the three new RPCs do get default commands — but with AutoCLI's default rendering, meaning `request_hash` arrives as a base64 flag rather than the hex positional argument the hand-written commands were written to accept. So the reachable UX is not the one the author designed, and it is inconsistent with the new *tx* commands, which are wired and do take hex. The PR body listing "CLI commands" among the delivered changes is inaccurate for queries as a result.

Same class, elsewhere in the diff: `Keeper.SetValidationRequest` in the new `x/agent/keeper/validation.go` is exported and never called — not by the handlers, not by genesis, not by tests.

**Acceptance criterion.** Either (a) add `RpcCommandOptions` entries for `ValidationRequest`, `ValidationResponses`, and `ValidationRequestsByAgent` to `AutoCLIOptions()` and delete `query_validation.go` plus the now-empty `GetQueryCmd`, or (b) add a `GetQueryCmd()` method to the agent module so the hand-written commands are actually reachable — noting that AutoCLI will then enhance rather than replace them. Either way, delete `SetValidationRequest`. Option (a) matches how the rest of this module's queries already work and is what I'd recommend.

</details>

<details><summary>Finding 5 — Non-blocking observations (no action required to merge)</summary>

These are noted for the author's judgement; none of them holds the merge.

- **Misleading error for an inactive subject.** `RequestValidation` returns `ErrValidatorInactive` when the *subject* agent is inactive, not the validator. Confusing in a failed-tx trace. A distinct error, or reusing `gerrc.ErrFailedPrecondition` with the agent id, would read better.

- **New keeper getters swallow real store errors.** `GetValidationRequest` and `GetValidationResponse` return `v, err == nil`, collapsing decode or store failures into "not found". The established pattern in this module (`GetAgent`, `GetActionLogEntry`) distinguishes `collections.ErrNotFound` and panics on anything else, so corruption surfaces instead of masquerading as absence. Not triggerable from well-formed state, but the drift is worth closing.

- **`events.proto` imports `validation.proto` and uses nothing from it.** Both new event messages are scalars only. Buf's lint config here doesn't flag unused imports and lint isn't run in CI, so nothing breaks — it's just a stray dependency edge in the descriptor.

- **No comments on the new proto messages or RPCs.** `buf.yaml` enables the `COMMENTS` category with only `COMMENT_FIELD` excepted, so `COMMENT_MESSAGE`/`COMMENT_RPC` would fire on every new message and both new `Msg` RPCs — but the CI proto workflow runs only `proto-format`, `proto-gen`, and breaking-change detection, never `buf lint`, and neighbouring existing messages already drift the same way. Cosmetic; the existing tx RPCs do carry comments, so matching them would keep the file consistent.

- **`request_hash` is not bound to the request it names.** The hash is caller-supplied, occupies a global first-writer-wins namespace, and the response references only the hash — so a third party can permanently occupy a hash an honest requester has already published off-chain, and a validator who works from the off-chain document rather than the on-chain record could have their verdict recorded against a substituted `agent_id`. I'm not raising this as blocking: ERC-8004 itself uses a caller-supplied request hash as the key, and the redirection case needs the validator to skip reading the record it was pointed at. If you want to close it anyway, deriving the key on-chain from `(requester, validator_id, agent_id, evidence_seq)` removes both the squatting race and the attribution ambiguity.

- **Unbounded progressive responses.** The named validator's owner can respond without limit, each write growing state. This is gas-metered and the PR describes progressive responses as intended, so it reads as a deliberate design choice rather than a gap.

- **No params migration needed.** I checked whether the three new `Params` fields would deserialize as zero on a live chain and brick the URI/tag length checks. `x/agent` appears in no upgrade handler under `app/upgrades/`, so the module is not yet live and `ConsensusVersion 1` with no migration is correct.

</details>

VERDICT: REQUEST_CHANGES

omriai1:
Resolved all four blocking findings in commit `652c3325`.

- Policy revocation now gates validator and subject at request and response time.
- Self-validation now rejects shared ownership across distinct agent IDs and is re-checked on response.
- Genesis validation now enforces response counts and contiguous sequences.
- Validation queries now use reachable AutoCLI descriptors; dead handwritten query commands and the unused setter were removed.

<details><summary>Fix-round verification</summary>

```text
go test ./x/agent/keeper -run Validation -count=1
ok  github.com/dymensionxyz/dymension/v3/x/agent/keeper

go test ./x/agent/types -run 'Genesis|Validation' -count=1
ok  github.com/dymensionxyz/dymension/v3/x/agent/types

go test ./x/agent -run TestAutoCLIValidationQueries -count=1
ok  github.com/dymensionxyz/dymension/v3/x/agent

golangci-lint run ./x/agent/...
0 issues.
```

</details>

omriai2 (approved):
## VERDICT: APPROVE

All four prior blocking findings are resolved by `652c3325` with non-vacuous tests behind each, and this round surfaced no new blocker — the trust boundaries now match the module's own established precedents, and the AutoCLI route the author chose turns out to preserve the hex UX the deleted hand-written commands were reaching for.

### Findings index

1. **Prior Finding 1 — policy-revocation denylist — RESOLVED.** `IsAgentLive` now gates validator and subject on both the request and response paths; the new test revokes real distinct policies for each role in each phase.
2. **Prior Finding 2 — owner-level self-validation — RESOLVED.** `validator.Owner == agent.Owner` rejects at request time and is re-checked at response time, each with its own test.
3. **Prior Finding 3 — genesis cannot uphold the invariant — RESOLVED.** `GenesisState.Validate` now enforces both `Seq < ResponseCount` and a per-hash response tally against `ResponseCount`, with a test covering both rejection paths.
4. **Prior Finding 4 — unreachable query CLI — RESOLVED.** Both dead files and the unused keeper setter are gone; the three query RPCs are declared in `AutoCLIOptions` with positional args. I verified AutoCLI's `bytes` handling accepts hex, so the sub-point about base64/hex inconsistency is closed too.
5. **Nits, none blocking** — redundant `Active` checks now duplicated ahead of `IsAgentLive`, `ErrValidatorInactive` reused for subject-agent failures, a descriptor-only AutoCLI test, and four carried-over non-blockers from the first round.

Proof of execution passes shape: five claims, each carrying a named command, plus an explicit owned "Not verified" for live-chain smoke. Every claim is CI-gated (`buf-break-check` green; `golangci-lint`, `build`, and `Analyze` still running on the new head), no external dependency call is added, and there is no QA Brief section, so I opened no transcripts. There is still no `x/agent` README, so no documentation was made inaccurate.

<details><summary>Finding 1 — Prior Finding 1 (policy-revocation denylist): RESOLVED</summary>

`RequestValidation` now routes both roles through the module's effective-liveness helper, and `RespondValidation` re-gates at response time and additionally loads and gates the subject:

```go
	if !k.IsAgentLive(ctx, msg.ValidatorId) { ... }
	...
	if !k.IsAgentLive(ctx, msg.AgentId) { ... }
```

```go
	if !k.IsAgentLive(ctx, req.ValidatorId) {
		return nil, types.ErrValidatorInactive
	}
	agent, found := k.GetAgent(ctx, req.AgentId)
	if !found { ... }
	if !agent.Active || !k.IsAgentLive(ctx, req.AgentId) { ... }
```

That covers what I asked for and goes one step further by re-gating the subject on the response path.

The test is not vacuous, which was my main concern: `TestValidationRejectsRevokedAgents` assigns two genuinely distinct pre-existing policy fixtures (`policyA()` from the validator, `policyB()` from the subject — both defined in `revoke_test.go`), derives real fingerprints via `types.PolicyFingerprint`, and walks four cases — validator revoked at request, validator revoked at response, subject revoked at request, subject revoked at response — with a `DeleteRevoked` round-trip in between proving the request genuinely succeeds when nothing is revoked. Had the fingerprint been a constant across agents, the subject cases would have passed for the wrong reason; they don't.

**One consequence worth the author knowing, not a finding.** Re-gating the *subject* at response time means a subject whose owner deactivates it after a request is filed can no longer receive a verdict. I checked whether that is a cheap way to censor an adverse verdict: `Active = true` is written only at registration and `Active = false` only at deactivation, with no reactivation path anywhere in the keeper. Deactivation is therefore terminal and destroys the agent's on-chain identity outright, which costs far more than any single verdict. The dodge is economically irrational, so the stricter gate is the right trade.

</details>

<details><summary>Finding 2 — Prior Finding 2 (owner-level self-validation): RESOLVED</summary>

The ID comparison is retained as a cheap short-circuit and the owner comparison is added behind it, exactly as asked:

```go
	if msg.ValidatorId == msg.AgentId {
		return nil, types.ErrSelfValidation
	}
	if validator.Owner == agent.Owner {
		return nil, types.ErrSelfValidation
	}
```

and re-checked on the response path, which is the part that matters because ownership can move between request and response:

```go
	if req.ValidatorId == req.AgentId || validator.Owner == agent.Owner {
		return nil, types.ErrSelfValidation
	}
```

Two tests cover the two distinct paths. `TestValidationRejectsSameOwnerAcrossAgentIDs` repoints the subject's owner to the validator's owner and asserts the *request* is rejected with `ErrSelfValidation`. `TestValidationResponseRechecksSubjectOwner` is the more interesting one — it files a legitimate request between independently-owned agents, *then* transfers subject ownership to the validator's owner, and asserts the response is rejected. That is precisely the ownership-moves-mid-flow case, and it would pass silently if only the request-time check existed.

The registry now enforces the same boundary the feedback subsystem enforces with `ErrSelfFeedback`, closing the one-account-two-agents path. As I said last round, this does not eliminate two-account collusion and was never meant to; consistency with the sibling subsystem was the criterion and it is met.

</details>

<details><summary>Finding 3 — Prior Finding 3 (genesis cannot uphold the invariant): RESOLVED</summary>

`Validate` now carries the full request rather than a presence marker, checks each response's sequence against its request's count, tallies per hash, and reconciles at the end:

```go
		if v.Seq >= request.ResponseCount {
			return fmt.Errorf("non-contiguous validation response sequence: hash %x seq %d count %d", ...)
		}
		...
		responseCounts[string(v.RequestHash)]++
	}
	for hash, request := range requests {
		if responseCounts[hash] != request.ResponseCount {
			return fmt.Errorf("validation response count mismatch: hash %x stored %d actual %d", ...)
		}
	}
```

`Seq < ResponseCount` combined with the pre-existing `(hash, seq)` uniqueness check gives genuine contiguity over `0..count-1`, which is what closes the silent-overwrite path: `RespondValidation` derives its slot as `seq := req.ResponseCount`, so contiguity is exactly the property that guarantees the next write lands on a free slot. The count reconciliation closes the chain-halt path, since the imported state can no longer disagree with what `InvariantValidationResponseIntegrity` asserts.

`TestValidationGenesisRejectsResponseCountAndSequenceMismatch` covers both branches with distinct assertions rather than one combined case.

I considered whether the closing `for hash, request := range requests` loop introduces non-determinism, since Go randomizes map iteration order. It does not matter here: the loop's observable outcome is error-or-nil, which is order-independent, and only the error *string* varies when multiple requests are inconsistent. More decisively, `GenesisState.Validate` reaches the chain only through `AppModuleBasic.ValidateGenesis`, which genutil invokes from `init`/`validate-genesis` tooling — `InitChain` calls `InitGenesis` directly and never routes through it. This is off the consensus path entirely.

</details>

<details><summary>Finding 4 — Prior Finding 4 (unreachable query CLI): RESOLVED, and the hex/base64 sub-point is closed</summary>

The author took the option I recommended. `x/agent/client/cli/query.go` and `query_validation.go` are deleted outright, `Keeper.SetValidationRequest` is removed, and the three RPCs are declared in `AutoCLIOptions` with positional arguments:

```go
				{
					RpcMethod:      "ValidationRequest",
					Use:            "validation-request [request-hash]",
					Short:          "Show a validation request by hash",
					PositionalArgs: []*autocliv1.PositionalArgDescriptor{{ProtoField: "request_hash"}},
				},
```

`rg -n "CmdQueryValidation|GetQueryCmd|SetValidationRequest|validationHashQuery" x/agent/` now returns nothing (exit 1), so no dangling references survive the deletion, and the remaining `client/cli` directory is tx-only.

I checked the one thing that could have turned this fix into a new regression: whether AutoCLI can build a *positional* argument from a proto `bytes` field, given that `root.go` does `panic(err)` on `EnhanceRootCommand` failure — a binary that won't start. It can. `client/v2@v2.0.0-beta.8`'s flag builder maps `protoreflect.BytesKind` to `binaryType`, and that type's `Set` tries file path, then hex, then base64:

```go
	if data, err := hex.DecodeString(s); err == nil {
		f.value = data
		return nil
	}
```

Hex is attempted before base64, so `dymd query agent validation-request <hex>` behaves exactly like the deleted hand-written command did, and matches the hex convention the wired tx commands already use. The UX-inconsistency sub-point I raised last round is therefore closed rather than merely traded away.

`module_test.go` asserts the three entries are present in the options struct. That is a weaker guard than building the command would be — it would not catch an unbuildable descriptor — but the descriptor path is what I verified above by reading the builder, and the acceptance criterion was wiring, which is met.

</details>

<details><summary>Finding 5 — Nits and carried-over non-blockers (no action required to merge)</summary>

New this round, both cosmetic:

- **Redundant liveness checks.** `RequestValidation` now reads `if !validator.Active { ... }` immediately followed by `if !k.IsAgentLive(ctx, msg.ValidatorId) { ... }`, and the same pair for the subject; `RespondValidation` does likewise. `IsAgentLive` re-fetches the agent from the store and re-checks `Active` itself, so each pair is one redundant store read and two ways of expressing one rule. Dropping the bare `Active` checks in favour of `IsAgentLive` alone would be DRYer and would leave a single place to change if the liveness definition evolves again. Behaviour is identical either way.

- **`ErrValidatorInactive` now names the subject in more places.** The misleading-error nit I raised last round widened from one site to four — the subject-agent rejections in both handlers return an error whose text says "validator inactive". A distinct error, or `gerrc.ErrFailedPrecondition` wrapped with the agent id, would read better in a failed-tx trace.

Carried over unchanged from the first round, still non-blocking:

- `GetValidationRequest` and `GetValidationResponse` return `v, err == nil`, collapsing decode or store failures into "not found", where the module's `GetAgent` and `GetActionLogEntry` distinguish `collections.ErrNotFound` and panic on anything else.
- `events.proto` imports `validation.proto` and uses nothing from it; both new event messages are scalars only.
- The new proto messages and both new `Msg` RPCs carry no comments, which the `COMMENTS` lint category in `buf.yaml` would flag — but CI runs only `proto-format`, `proto-gen`, and breaking-change detection, never `buf lint`, and neighbouring existing messages drift the same way.
- `request_hash` remains caller-supplied in a global first-writer-wins namespace, unbound to the request it names. As before I am not holding the merge for this: ERC-8004 itself keys on a caller-supplied request hash, and the verdict-redirection case needs a validator who skips reading the on-chain record it was pointed at.

</details>

VERDICT: APPROVE

PR #2201, opened 2026-08-22
fix(forward): clamp stale IBC timeouts

### LINKED ISSUE #2200: x/forward: clamp stale absolute IBC timeouts on completion-hook & Hyperlane-inbound forwards (autonomous cross-chain intents strand funds when Hub-side completion outlives the composer's deadline)
## Problem / opportunity

`x/forward` is Dymension's cross-chain routing hub: it takes funds that have just arrived on the Hub (via Hyperlane, or via an IBC/EIBC completion hook from a RollApp) and executes a **second hop** to their real destination. The IBC second hop is built in `forwardToIBC` (`x/forward/ibc.go`):

```go
func (k Forward) forwardToIBC(ctx sdk.Context, transfer *ibctransfertypes.MsgTransfer, fundsSrc sdk.AccAddress, maxBudget sdk.Coin) error {
	m := ibctransfertypes.NewMsgTransfer(
		transfer.SourcePort,
		transfer.SourceChannel,
		maxBudget,
		fundsSrc.String(),
		transfer.Receiver,
		ibcclienttypes.Height{},
		transfer.TimeoutTimestamp, // <-- copied verbatim from the composer's memo
		transfer.Memo,
	)
	_, err := k.transferK.Transfer(ctx, m)
	return err
}
```

`transfer.TimeoutTimestamp` is an **absolute** nanosecond timestamp chosen by whoever composed the route, at composition time. It is baked into the packet memo by `NewHookForwardToIBC(...)` / `MakeRolForwardToIBCMemoString(...)` (`x/forward/types/to_ibc.go`) and only surfaces here — potentially **much later** than it was chosen.

The gap: the interval between composition and execution of the second hop is **unbounded and unpredictable from the composer's point of view**.

- For the `rollToIBCHook` completion-hook path, `Run` fires either when the RollApp packet is **EIBC-fulfilled** (seconds/minutes) **or**, if no market maker fulfills it, only when the packet **finalizes** — which for an optimistic RollApp is the full dispute/finalization window (hours to days). The composer cannot know in advance which branch, or when.
- For the `OnHyperlaneMessage` -> `forwardToIBC` inbound path, execution is gated on Hyperlane relayer delivery, which is also not known at composition time.

So an absolute `TimeoutTimestamp` that was reasonable when the route was composed (e.g. `now + 10m`) is very often **already in the past** by the time `forwardToIBC` runs. When that happens, the outbound `MsgTransfer` is built with a stale timeout; ibc-go's send path rejects a packet whose timeout has already elapsed against the counterparty's latest consensus state, the hook's atomic body returns an error, and — per the module's deliberate "if it fails, the original target got the funds anyway ... relying on frontend here" fallback — the funds are **stranded at the intermediate recipient**. The user's declared intent (reach the final IBC destination) silently fails and requires manual, frontend-assisted recovery.

`HookForwardToIBC.ValidateBasic()` only calls `transfer.ValidateBasic()`, which checks the timeout is *set*, never that it is still *usable* at execution time — so nothing catches this.

## Why now

The agent economy composes exactly these delay-then-settle cross-chain intents, and the composer is increasingly an autonomous program with **no frontend** to catch a stranded intermediate balance and retry:

- **Intent-based cross-chain settlement (ERC-7683, Uniswap/Across)** is now the dominant pattern: a signer declares "inputs here, outputs there" and a fill happens later; robustness of the *fill* leg, not the composer's stale local clock, is what must guarantee delivery. https://www.eco.com/support/en/articles/15264366-what-is-a-cross-chain-intent-engine-2026-guide · https://eips.ethereum.org/EIPS/eip-7683
- **x402 agentic payments** crossed ~75M transactions / ~$24M in a single 30-day window in 2026 (Coinbase/Linux Foundation), with autonomous agents settling machine-to-machine and no human in the loop to babysit a half-completed multi-hop route. ``https://www.coindesk.com/tech/2026/07/15/visa-mastercard-and-ripple-join-the-standard-letting-ai-agents-pay-in-stablecoins``

For these callers, a route that composes a Roll->IBC or HL->IBC forward and then loses the funds at the intermediate hop because the Hub-side completion arrived after the composer's absolute deadline is a real, recurring failure mode.

## Proposed change

Clamp the outbound IBC timeout deterministically at execution time, at the single choke point `forwardToIBC` (`x/forward/ibc.go`), so it covers **both** callers (`rollToIBCHook.Run` and `OnHyperlaneMessage`).

Rule (conservative — only ever *extend* an unusable deadline, never *shorten* a healthy one, so composer intent is preserved whenever it is still honorable):

```go
// x/forward/ibc.go, inside forwardToIBC, before building the MsgTransfer:
nowNs := uint64(ctx.BlockTime().UnixNano()) //nolint:gosec // block time is never negative
timeoutTimestamp := transfer.TimeoutTimestamp
if timeoutTimestamp == 0 || timeoutTimestamp <= nowNs+uint64(types.MinForwardIBCTimeout) {
	// the composer's absolute deadline is stale/unusable at execution time;
	// give the final hop a fresh, deliverable deadline instead of stranding funds
	timeoutTimestamp = nowNs + uint64(types.DefaultForwardIBCTimeout)
}
```

and pass `timeoutTimestamp` (not `transfer.TimeoutTimestamp`) into `NewMsgTransfer`.

Add the two constants to `x/forward/types/const.go`:

```go
// DefaultForwardIBCTimeout is the fresh timeout applied to a forwarded IBC
// hop when the composer's absolute timeout is missing or already stale at
// execution time. Nanoseconds.
const DefaultForwardIBCTimeout = uint64(time.Hour)

// MinForwardIBCTimeout is the minimum remaining lead time a composer-supplied
// timeout must have (relative to block time) to be honored as-is; anything
// tighter is treated as stale and replaced with DefaultForwardIBCTimeout.
const MinForwardIBCTimeout = uint64(10 * time.Minute)
```

Behavior:
- Composer supplied a healthy future timeout (>= now + 10m): honored unchanged. No behavior change for the happy path.
- Composer supplied a past, zero, or too-tight timeout: replaced with `now + 1h`, so the final hop actually delivers.

`ctx.BlockTime()` is the block header time — identical on every validator — so the clamp is fully deterministic.

- **Constants, not params.** `x/forward` has no keeper store, params, or genesis (`New(...)` in `x/forward/forward.go` takes only the transfer keeper + warp query/msg servers). Introducing a param store purely for a timeout would be disproportionate. Use two hardcoded constants in `x/forward/types/const.go`. Values: `DefaultForwardIBCTimeout = 1 hour`, `MinForwardIBCTimeout = 10 minutes`, both in nanoseconds. These mirror common IBC relayer defaults and are generous enough for cross-chain relay under normal skew.
- **Only extend, never shorten.** If the composer's timeout is still at least `MinForwardIBCTimeout` in the future, it is preserved exactly. The clamp activates only when the deadline is `0` or `<= now + MinForwardIBCTimeout`. This preserves composer intent whenever it remains deliverable and changes behavior only in the case that currently strands funds.
- **Single choke point.** Apply the clamp in `forwardToIBC` so it covers both the completion-hook path (`rollToIBCHook`) and the Hyperlane-inbound path (`OnHyperlaneMessage`). Do not duplicate the logic in callers.
- **Base the clock on `ctx.BlockTime()`**, never wall-clock, for determinism.
- **Scope to the IBC hop only.** The Hyperlane hop (`forwardToHyperlane` / `warptypes.MsgRemoteTransfer`) does not use an IBC-style absolute timeout (it relies on relayer gas / `MaxFee`), so no equivalent change is made there.

## Acceptance criteria

1. `DefaultForwardIBCTimeout` and `MinForwardIBCTimeout` exist as nanosecond constants in `x/forward/types/const.go`.
2. `forwardToIBC` builds its outbound `MsgTransfer` with the clamped timeout per the rule above.
3. New unit test(s) in `x/forward` (a mock `TransferKeeper` — the interface is a single `Transfer(ctx, *MsgTransfer)` method, `x/forward/types/expected_keepers.go` — captures the outbound `MsgTransfer`) assert:
   - stale/past `TimeoutTimestamp` (e.g. `blockTime - 1m`) => captured `MsgTransfer.TimeoutTimestamp == blockTime + DefaultForwardIBCTimeout`;
   - zero `TimeoutTimestamp` => replaced with `blockTime + DefaultForwardIBCTimeout`;
   - healthy future `TimeoutTimestamp` (e.g. `blockTime + 2h`) => passed through unchanged;
   - a `TimeoutTimestamp` exactly at the `now + MinForwardIBCTimeout` boundary is treated as stale (replaced), consistent with the `<=` comparison.
4. The clamped value depends only on `ctx.BlockTime()` and the two constants — no wall-clock, randomness, or map iteration. `go test ./x/forward/...` passes.
5. No change to the happy-path behavior for healthy timeouts (regression covered by the pass-through test).

## Out of scope

- The Hyperlane outbound path (`forwardToHyperlane`) and its `MaxFee`/gas semantics.
- The module's existing "funds remain with the intermediate recipient on failure" fallback for *other* failure causes (denom mismatch, insufficient budget, etc.) — this issue only removes the stale-timeout cause.
- Leftover/change-stranding behavior on the HL path (fixed `Amount` vs `budget`).
- Adding params/genesis/store to `x/forward`.
- Any change to EIBC fulfillment or delayedack finalization timing.

---

omriai1:
Closes #2200

Outbound IBC forwards could inherit an absolute timeout that had already expired while waiting for RollApp completion or Hyperlane delivery, leaving funds at the intermediate recipient. This change deterministically replaces missing, stale, and too-tight deadlines with a fresh one-hour timeout at the shared forwarding choke point while preserving healthy composer deadlines unchanged.

- Clamp stale IBC timeouts once in `forwardToIBC`.
- Cover past, zero, healthy, and inclusive minimum-boundary behavior.
- Keep the result deterministic from consensus block time.

- Adds `DefaultForwardIBCTimeout` (one hour) and `MinForwardIBCTimeout` (ten minutes) as nanosecond constants.
- Computes `nowNs` exclusively from `ctx.BlockTime()`.
- Replaces `TimeoutTimestamp == 0` or `TimeoutTimestamp <= nowNs + MinForwardIBCTimeout` with `nowNs + DefaultForwardIBCTimeout`.
- Leaves healthy future timestamps untouched and makes no changes to Hyperlane outbound forwarding.

</details>

## Proof of execution

- `TestForwardToIBCClampsTimeoutTimestamp` — PASS for past, zero, healthy two-hour, and exact ten-minute boundary cases via `go test ./x/forward/... -run TestForwardToIBCClampsTimeoutTimestamp -count=1 -v`; CI Build and Test re-runs it.
- Full repository race/coverage suite — PASS via `go-acc -o /tmp/dymension-2200-coverage.txt ./... -- -v --race`; CI Build and Test re-runs this gate.
- Full repository build — PASS via `go build ./...`; CI Build and Test re-runs it.
- Full repository lint — PASS with `0 issues` via `golangci-lint run`; CI golangci-lint re-runs it.

**Not verified:** nothing; all changed backend behavior is covered by committed unit tests and the repository gates above.

### DIFF
--- a/x/forward/ibc.go
+++ b/x/forward/ibc.go
@@ -90,6 +90,13 @@ func (h rollToIBCHook) Run(ctx sdk.Context, fundsSource sdk.AccAddress, budget s
 }
 
 func (k Forward) forwardToIBC(ctx sdk.Context, transfer *ibctransfertypes.MsgTransfer, fundsSrc sdk.AccAddress, maxBudget sdk.Coin) error {
+	nowNs := uint64(ctx.BlockTime().UnixNano()) //nolint:gosec // block time is never negative
+	timeoutTimestamp := transfer.TimeoutTimestamp
+	if timeoutTimestamp == 0 || timeoutTimestamp <= nowNs+types.MinForwardIBCTimeout {
+		// Give the final hop a fresh deadline when the composer's absolute timeout is no longer usable.
+		timeoutTimestamp = nowNs + types.DefaultForwardIBCTimeout
+	}
+
 	m := ibctransfertypes.NewMsgTransfer(
 		transfer.SourcePort,
 		transfer.SourceChannel,
@@ -97,7 +104,7 @@ func (k Forward) forwardToIBC(ctx sdk.Context, transfer *ibctransfertypes.MsgTra
 		fundsSrc.String(),
 		transfer.Receiver,
 		ibcclienttypes.Height{}, // ignore, removed in ibc v2 also
-		transfer.TimeoutTimestamp,
+		timeoutTimestamp,
 		transfer.Memo, // include the original memo, so that we can have more functionality down the road (.e.g actions on rollapp)
 	)
 
--- a/x/forward/ibc_test.go
+++ b/x/forward/ibc_test.go
@@ -1,0 +1,76 @@
+package forward
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	sdkmath "cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	ibctransfertypes "github.com/cosmos/ibc-go/v8/modules/apps/transfer/types"
+	"github.com/stretchr/testify/require"
+)
+
+type capturingTransferKeeper struct {
+	msg *ibctransfertypes.MsgTransfer
+}
+
+func (k *capturingTransferKeeper) Transfer(_ context.Context, msg *ibctransfertypes.MsgTransfer) (*ibctransfertypes.MsgTransferResponse, error) {
+	k.msg = msg
+	return &ibctransfertypes.MsgTransferResponse{}, nil
+}
+
+func TestForwardToIBCClampsTimeoutTimestamp(t *testing.T) {
+	blockTime := time.Date(2026, time.August, 22, 12, 0, 0, 0, time.UTC)
+	blockTimeNs := uint64(blockTime.UnixNano())
+	healthyTimeout := uint64(blockTime.Add(2 * time.Hour).UnixNano())
+
+	testCases := []struct {
+		name             string
+		timeoutTimestamp uint64
+		expectedTimeout  uint64
+	}{
+		{
+			name:             "past timeout",
+			timeoutTimestamp: uint64(blockTime.Add(-time.Minute).UnixNano()),
+			expectedTimeout:  uint64(blockTime.Add(time.Hour).UnixNano()),
+		},
+		{
+			name:             "zero timeout",
+			timeoutTimestamp: 0,
+			expectedTimeout:  uint64(blockTime.Add(time.Hour).UnixNano()),
+		},
+		{
+			name:             "healthy timeout",
+			timeoutTimestamp: healthyTimeout,
+			expectedTimeout:  healthyTimeout,
+		},
+		{
+			name:             "minimum lead time boundary",
+			timeoutTimestamp: blockTimeNs + uint64(10*time.Minute),
+			expectedTimeout:  uint64(blockTime.Add(time.Hour).UnixNano()),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			transferKeeper := &capturingTransferKeeper{}
+			forward := New(transferKeeper, nil, nil)
+			transfer := &ibctransfertypes.MsgTransfer{
+				SourcePort:       "transfer",
+				SourceChannel:    "channel-0",
+				Receiver:         "destination",
+				TimeoutTimestamp: tc.timeoutTimestamp,
+				Memo:             "memo",
+			}
+			fundsSource := sdk.AccAddress([]byte("funds-source-address"))
+			budget := sdk.NewCoin("adym", sdkmath.NewInt(10))
+
+			err := forward.forwardToIBC(sdk.Context{}.WithBlockTime(blockTime), transfer, fundsSource, budget)
+
+			require.NoError(t, err)
+			require.NotNil(t, transferKeeper.msg)
+			require.Equal(t, tc.expectedTimeout, transferKeeper.msg.TimeoutTimestamp)
+		})
+	}
+}
--- a/x/forward/types/const.go
+++ b/x/forward/types/const.go
@@ -1,8 +1,19 @@
 package types
 
+import "time"
+
 const (
 	ModuleName = "forward"
 	// not to be confused with ibc apps PFM which uses 'forward' as the fungible packet json memo key
 	HookNameRollToHL  = "dym-fwd-roll-hl"
 	HookNameRollToIBC = "dym-fwd-roll-ibc"
+
+	// DefaultForwardIBCTimeout is the fresh timeout applied to a forwarded IBC
+	// hop when the composer's absolute timeout is missing or already stale at
+	// execution time. Nanoseconds.
+	DefaultForwardIBCTimeout = uint64(time.Hour)
+	// MinForwardIBCTimeout is the minimum remaining lead time a composer-supplied
+	// timeout must have (relative to block time) to be honored as-is; anything
+	// tighter is treated as stale and replaced with DefaultForwardIBCTimeout.
+	MinForwardIBCTimeout = uint64(10 * time.Minute)
 )


omriai2 (changes requested):
# Security Review — PR #2201 (`fix(forward): clamp stale IBC timeouts`)

**No HIGH or MEDIUM severity findings.**

## Analysis performed

Data flow traced from the attacker-controlled source to the sensitive operation:

- **Source:** `transfer.TimeoutTimestamp` is a `uint64` decoded from a composer-supplied memo (`UnpackForwardToIBC` on the Hyperlane inbound path, `proto.Unmarshal` of `HookForwardToIBC` on the completion-hook path). Fully attacker-controlled.
- **Sink:** the value is passed to `ibctransfertypes.NewMsgTransfer` and then to `transferK.Transfer`, emitting a fund-moving IBC packet.
- **New logic:** a branch that replaces the value with `ctx.BlockTime().UnixNano() + 1h` when it is `0` or `<= now + 10m`.

## Candidates considered and dismissed

**Integer overflow in the clamp arithmetic (`arithmetic_overflow`) — not exploitable.** Both additions are `nowNs + <constant>`, where `nowNs` is derived from block time (~`1.79e18` ns in 2026) and the constants are `3.6e12` and `6e11`. The sums stay four orders of magnitude below `math.MaxUint64` (`1.845e19`), leaving roughly 500 years of headroom. Neither addend is attacker-controlled, so an attacker cannot steer the sum toward wraparound. The attacker-controlled operand appears only on the left of a comparison, never in an addition.

**Signed-to-unsigned conversion of `UnixNano()` (`type_confusion`) — not exploitable.** A pre-1970 block time would wrap to a large `uint64`, but block time is the consensus header time, produced by the validator set and always well past genesis. It is not reachable by an external caller, so there is no attack path.

**Attacker forcing a longer packet lifetime (`business_logic`) — no privilege boundary crossed.** An attacker can guarantee the 1-hour branch by supplying `TimeoutTimestamp = 0` or any tight value. However, the funds moved are the attacker's own: `fundsSrc` is the intermediate recipient that the inbound transfer or EIBC fulfiller already credited, and on timeout the IBC transfer app refunds to that same address. The only effect is that the attacker's own funds sit in the transfer escrow longer before refunding to the attacker. No other account's balance, authorization state, or escrow accounting is reachable. The residual "more packet commitments live in state longer" angle is a resource-consumption concern and falls under the DOS exclusion.

**Override of a composer-declared deadline (`authorization_bypass`) — no boundary crossed.** The clamp can extend a deliberately tight deadline up to an hour. This is a semantic change, but the composer of the memo and the owner of the forwarded funds are the same principal by construction — the memo rides along with the inbound transfer whose recipient becomes `fundsSrc`. There is no third party whose stated policy is being overridden, so this is a product design trade-off rather than an authorization bypass.

**Consensus non-determinism (`logic_flaw`) — not present.** The clamped value is a pure function of `ctx.BlockTime()` and two compile-time constants. No wall-clock read, no randomness, no map iteration. Every validator computes the identical packet.

**Test file** — `x/forward/ibc_test.go` is a unit test only and is excluded from scope.

## Net security posture

The change strictly narrows an existing failure mode: previously a stale absolute timeout caused ibc-go's `SendPacket` elapsed-timeout check to reject the send, leaving funds parked at the intermediate recipient. The clamp only ever extends an unusable deadline and never shortens a healthy one, so it introduces no new path by which funds can reach an unintended recipient.

omriai1:
Security review disposition: no branch change is required because the review reports no HIGH/MEDIUM findings and dismisses every candidate it analyzes.

- Arithmetic overflow: no action; both additions use consensus block time plus fixed constants, with no attacker-controlled addend.
- Signed-to-unsigned conversion: no action; the reviewed source is consensus header time and the existing `gosec` suppression documents the non-negative invariant.
- Longer packet lifetime: no action; the review finds no privilege boundary or unintended-account impact.
- Composer deadline override: no action; the behavior is the issue's explicitly required product semantics, and the review finds no authorization boundary crossed.
- Consensus determinism: no action; the value remains a pure function of `ctx.BlockTime()` and compile-time constants.
- Unit test scope: no action; the review explicitly excludes the test file and identifies no test concern.

The branch and existing proof receipt therefore remain current. Please clear the changes-requested state since the review contains no requested code, test, documentation, or proof correction.

omriai2 (approved):
**VERDICT: APPROVE** — the prior round's posted review named no blocking finding (it was a security report that dismissed every candidate it raised), this round found none either, and the fix is a correct, deterministic, well-scoped nine-line change with committed regression coverage that passes.

## Index

No blocking findings.

1. *(Non-blocking nit)* The `--timeout` flag on the `create-memo` / `create-hl-message` builders defaults to `5m`, a value the new floor now always discards — worth a help-text or default refresh, but not merge-blocking.
2. *(Non-blocking observation)* The regression test asserts only the timeout field, not the pass-through of the other message fields.
3. *(Non-blocking observation)* The floor is measured against Hub block time, while the check that actually rejects a stale packet compares against the counterparty's consensus timestamp — a residual gap inherited from the issue's chosen design, not introduced here.

## Prior-round accounting

<details><summary>Disposition of the previous review</summary>

The previous review posted to this PR was a security report that opened with "No HIGH or MEDIUM severity findings" and then walked through five candidate vulnerability classes — arithmetic overflow, signed-to-unsigned conversion, extended packet lifetime, composer-deadline override, and consensus non-determinism — dismissing each with reasoning. It requested no code, test, documentation, or proof correction anywhere in its body. It was nonetheless submitted in the `CHANGES_REQUESTED` state, which was a mismatch between the review's content and its recorded state, not a finding the author failed to address.

The author's response comment on the PR walks the same six items and concludes no branch change is required. That reading is correct, and I am confirming it here rather than asking for anything further. There is consequently nothing outstanding to carry into this round: `git log main..origin/omridagan/workforce-issue-2200` shows the single commit `7c87cadb` unchanged, and no fix push was owed.

For completeness, I re-derived the security posture independently this round rather than deferring to the earlier report. The attacker-controlled value — `transfer.TimeoutTimestamp`, decoded from a composer-supplied memo on both the Hyperlane-inbound and completion-hook paths — appears only on the left-hand side of a comparison and is never an addend. Both additions are consensus block time plus a compile-time constant, landing around `1.79e18` against a `uint64` ceiling of `1.845e19`, so wraparound is unreachable and not steerable. The funds moved belong to the same principal that composed the memo, since the intermediate recipient credited by the inbound transfer becomes the funds source and is also the refund target on timeout, so the extended deadline cannot reach another account's balance. I agree with the no-findings conclusion.

</details>

## This round's verification

<details><summary>What I checked and what it proves</summary>

**Diff scope.** Three files, ninety-five lines: a six-line clamp inside the shared IBC forwarding function, two new nanosecond constants, and a new table-driven unit test. I read all of it plus the surrounding module — the two hook entry points, the Hyperlane inbound handler, the atomic-execution wrapper, the memo constructors, and the expected-keepers interface.

**The fix is at the right place and covers every caller.** The clamp sits in the single function that both the RollApp completion-hook path and the Hyperlane-inbound path call to build the outbound transfer. I also traced the third composer entry point — the IBC-to-IBC memo builder — and confirmed it dispatches through the same completion-hook name to the same function, so all three routes are covered by one edit with no duplicated logic in callers.

**The behavior matches the specified rule exactly.** A missing or too-tight deadline is replaced with block time plus one hour; anything with at least ten minutes of headroom passes through untouched. The comparison is inclusive at the boundary, matching the spec. The Hyperlane outbound hop is left alone, as scoped.

**Determinism.** The clamped value is a pure function of `ctx.BlockTime()` and two compile-time constants — no wall clock, no randomness, no map iteration. Every validator computes an identical packet, which is the property that matters for a state-machine change on this chain.

**Regression test — verified by execution, not by assertion.** I ran the single new test case in isolation rather than any suite:

```
$ go test ./x/forward/ -run TestForwardToIBCClampsTimeoutTimestamp -count=1 -v
=== RUN   TestForwardToIBCClampsTimeoutTimestamp
=== RUN   TestForwardToIBCClampsTimeoutTimestamp/past_timeout
=== RUN   TestForwardToIBCClampsTimeoutTimestamp/zero_timeout
=== RUN   TestForwardToIBCClampsTimeoutTimestamp/healthy_timeout
=== RUN   TestForwardToIBCClampsTimeoutTimestamp/minimum_lead_time_boundary
--- PASS: TestForwardToIBCClampsTimeoutTimestamp (0.00s)
    --- PASS: TestForwardToIBCClampsTimeoutTimestamp/past_timeout (0.00s)
    --- PASS: TestForwardToIBCClampsTimeoutTimestamp/zero_timeout (0.00s)
    --- PASS: TestForwardToIBCClampsTimeoutTimestamp/healthy_timeout (0.00s)
    --- PASS: TestForwardToIBCClampsTimeoutTimestamp/minimum_lead_time_boundary (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/forward	0.116s
```

All four cases the acceptance criteria call for are present and green, including the pass-through case that guards the happy path against regression. That the package compiles also settles a question I had about the constant declarations: converting a typed duration constant to `uint64` inside a `const` block is a valid constant expression here, and the values are the intended nanosecond counts.

**Proof of execution.** The section is present and well-formed. Four claims, each carrying a named command as its receipt, plus an explicit owned "Not verified: nothing" line. Every claim — the targeted test, the race/coverage suite, the build, and the lint run — is covered by a required check, so the named suite plus CI is the proof and I did not open transcripts. The diff adds no external dependency call, so no live-invocation receipt is owed.

**CI.** The lint gate has since gone green (`golangci-lint pass 4m17s`), which independently confirms the `gosec` suppression on the time conversion is accepted. The build/test gate and the code-scanning job are still pending; pending required checks are the merge gate's job, not a review finding, and I verified the one behavioral claim they carry myself above.

**Consensus-breaking change handling.** I checked whether this repository gates state-machine changes behind upgrade-handler registration. It does not — the most recent comparable fix on `main` changed liquidity-provider selection logic with no upgrade gating, consistent with the standard practice of shipping logic changes in a coordinated binary upgrade. No action needed.

</details>

<details><summary>Finding 1 (non-blocking nit) — the CLI `--timeout` default is now always discarded</summary>

The memo-builder commands take a `--timeout` flag, defaulting to five minutes, described as the IBC packet timeout duration with examples spanning thirty seconds to an hour. That flag composes an absolute deadline from the client's wall clock at build time.

With the new floor, any composed deadline sitting within ten minutes of Hub block time is replaced. So every value at or below ten minutes — including the flag's own default and two of the three examples in its help text — is guaranteed to be discarded at execution. A user who asks for a thirty-second deadline and later wonders why the transfer stayed in flight for an hour has nothing in the help text pointing at the Hub-side floor.

I am not holding the merge for this. The affected commands are developer-facing builders that produce a blob to paste elsewhere rather than broadcasting anything; the resulting behavior is strictly safer than what the help text implies; and the floor's semantics were an explicit, deliberate product decision in the issue rather than something this diff invented. It is a wart worth a one-line cleanup whenever that file is next touched — either nudging the default above the floor or adding a clause noting that the Hub enforces a minimum — not a reason to spend an author round-trip.

</details>

<details><summary>Finding 2 (non-blocking observation) — test asserts the timeout field only</summary>

The new test captures the outbound message and asserts its timeout, which is precisely the behavior under change and exactly what the acceptance criteria ask for. It does not additionally assert that the source port, channel, receiver, memo, token budget, or sender survive the rebuild unchanged.

Since this is the first and only unit test of this function, a future refactor could drop or transpose one of those fields and still see green. Adding a few field assertions alongside the existing one would turn this into a broader guard for very little effort. Optional, and not something I would gate on — the integration suite exercises the full forwarding path end to end.

</details>

<details><summary>Finding 3 (non-blocking observation) — the floor's reference clock differs from the check that rejects</summary>

Worth recording for whoever revisits this area, though it is not a defect in this diff.

The floor is measured against Hub block time. The check that actually rejects a stale outbound packet lives in the IBC channel layer and compares the deadline against the counterparty's latest consensus-state timestamp — the counterparty's own clock as of the last light-client update, not the Hub's. Those two clocks are not the same quantity.

In the common case this is harmless and in fact favorable: a lagging light client has a consensus timestamp behind Hub block time, which makes the elapsed check strictly more lenient than the floor assumes. The residual gap is the opposite direction — a counterparty whose consensus timestamp runs ahead of Hub block time by more than the granted headroom would still see the packet rejected, and the funds would still park at the intermediate recipient. One hour of granted headroom absorbs any realistic skew, so this is narrow.

Closing it properly would mean reading client state inside the forwarding function to reference the counterparty timestamp directly, which this module has no keeper access for and which the issue explicitly declined in favor of the block-time rule. The change as written strictly reduces the failure mode's frequency and never widens it, so there is nothing to fix here.

</details>

VERDICT: APPROVE

PR #2166, opened 2026-06-24
feat(agent): registry register + deactivate

omritoptix:
Closes #2159

Implements the `x/agent` registry: permissionless, fee-gated registration with an **immutable** policy, plus owner-only deactivation. No attestation logic (that's #2164, now merged — see integration note below).

## What's here

- **`MsgRegisterAgent { owner, agent_id, policy }`** — rejects a duplicate `agent_id`, charges `Params.agent_registration_fee` from the owner via send-to-module-then-burn (mirrors rollapp app registration), and stores `Agent{ id, owner, policy, active: true, action_seq: 0 }` with the embedded `x/common/tee` `Policy`. `ValidateBasic` enforces non-empty id, well-formed owner, and a well-formed policy (cert PEM parses, rego query/structure non-empty).
- **`MsgDeactivateAgent { owner, agent_id }`** — owner-only; sets `active: false`. A non-owner is rejected.
- **No policy-update message** — policy is immutable by design (a new version is a new registration).
- Both messages are wired into the proto `Msg` service (`cosmos.msg.v1.signer = "owner"`), the module codec, and the msg server, each emitting a typed event (`EventRegisterAgent` / `EventDeactivateAgent`).
- The agent module account gains `Burner` permission so the fee can be burned.
- `register-agent` / `deactivate-agent` CLI commands, plus a read-only `show-agent` query so registry state is inspectable on-chain.
- The whole module is wired into the app (keeper, keys, module registration) — this also makes the sibling `SubmitAttestedAction` reachable on-chain.

## Integration with #2164

#2164 (`SubmitAttestedAction`) merged first and brought its own copy of the agent module base, structured differently from the scaffold this PR was originally built on (`agent.proto` vs `d.proto`, keeper split across `agent.go`/`params.go`/`action_log.go`, value-receiver `msgServer`). I merged `main` and adopted that structure as canonical, then re-applied the registry on top:

- Added `owner` to `Agent` and `agent_registration_fee` to `Params`.
- Added a `bankKeeper` to the keeper for the fee path.
- Put `RegisterAgent`/`DeactivateAgent` in `keeper/registry.go`; merged all three RPCs into the proto `Msg` service and codec.
- Kept `SubmitAttestedAction` and its standalone keeper test untouched (its `setup` now passes a `nil` bankKeeper, which that path never uses).

## Proof of execution

All evidence regenerated after the merge.

### Both keeper suites + message tests pass together
**How:** `go test ./x/agent/... -v`
**Evidence:**
```
[...]

Issue #1890, opened 2025-06-04
chore: use github actions to enforce protobuf format, lint and gen


danwt:
Hyperlane have an example we can copy paste

https://github.com/bcp-innovations/hyperlane-cosmos/blob/main/.github/workflows/proto.yml

danwt:
@Angith go for it! thank you

Angith:
@danwt I’ve raised a PR to address the issue. I’ve tested the setup in my fork, and the protobuf checks are successfully triggered on pull requests and when changes are made to .proto files or generated .pb.go files.

Please note that there are currently some existing issues with a few protobuf files in the repository which may cause the checks to fail. These will need to be resolved separately.

pri-3x:
Hey.
I have been working with Golang with 3+ years, and I have good experience as well distributed system as well. It is a good first issue, shouldn't be tough. And I am worked on similar issues as well in the fast. Therefore I think I am the right person for this issue

danwt:
yep @pri-3x

afa-farkhod:
hey @danwt , I wonder why the current issue is still active? according the #1898 it should be closed or sub issue to be made?

PR #2205, opened 2026-08-29
fix(agent): bound validation responses per request

### LINKED ISSUE #2204: x/agent: bound responses per validation request (Validation Registry state-bloat & finalization guard for autonomous validators)
## Problem / opportunity

In the `x/agent` ERC-8004 **Validation Registry** (#2198), `MsgRespondValidation` appends a response and increments `req.ResponseCount` with **no upper bound** (`x/agent/keeper/msg_server_validation.go`):

```go
seq := req.ResponseCount
resp := types.ValidationResponse{RequestHash: msg.RequestHash, Seq: seq, ...}
k.validationResponses.Set(ctx, collections.Join(msg.RequestHash, seq), resp)
req.ResponseCount++
k.validationRequests.Set(ctx, msg.RequestHash, req)
```

The chosen validator's owner can respond to a single request an **unbounded** number of times. Each call writes a new permanent `ValidationResponse` under `(request_hash, seq)`. There is no param bounding this (`x/agent/types/agent.proto` Params has `max_action_bytes`, `feedback_tag_max_bytes`, `validation_tag_max_bytes`, `validation_uri_max_bytes`, `spend_recipient_allowlist_max` — but nothing for response count). Consequences:

- **Unbounded state growth / griefing:** a validator (or a compromised validator enclave) can bloat consensus state with arbitrarily many responses to one request. The only cost is gas, which does not bound total stored bytes.
- **No finalization signal:** a consumer reading validation results sees an ever-growing list with no notion of "this validation is complete." A validator can keep appending responses to reshape the record after the fact.

Unlike the request side — which is deduplicated (`validationRequests.Has(RequestHash)` rejects duplicates) — the response side has no such guard.

## Why now

Validation Registry validators in the ERC-8004 / x402 era are **autonomous agents**, not humans. Automated responders make unbounded append cheap and repeatable at machine speed, and the Validation Registry's TEE/response semantics are under active revision in the ERC-8004 community precisely around what a validation *result* commits to.

- https://eips.ethereum.org/EIPS/eip-8004
- https://github.com/erc-8004/erc-8004-contracts

## Proposed change

Add a governance-tunable cap on responses per validation request.

Files:

1. **`proto/dymensionxyz/dymension/agent/agent.proto`** — add `uint64 validation_max_responses_per_request = 10;` to `Params` (next free field number). Regenerate with `make proto-gen`.
2. **`x/agent/types/params.go`** — add a default (`DefaultValidationMaxResponsesPerRequest = 8`), wire into `DefaultParams()`, and validate in `Params.Validate()` (any `uint64` is structurally valid; `0` is the documented "unlimited" escape hatch, see decision below).
3. **`x/agent/types/errors.go`** — add `ErrTooManyValidationResponses`.
4. **`x/agent/keeper/msg_server_validation.go`** — in `RespondValidation`, after loading `req` and `p`, enforce:
   ```go
   if p.ValidationMaxResponsesPerRequest != 0 && req.ResponseCount >= p.ValidationMaxResponsesPerRequest {
       return nil, types.ErrTooManyValidationResponses
   }
   ```
   placed before writing the new response.

- **Default cap: `8`.** A request→validation is normally one authoritative result; a small allowance permits progressive or corrected results (re-validation) while bounding griefing. Re-validation beyond the cap is expressed as a **new** request (new `request_hash`), which is the existing dedup-keyed primitive.
- **`0` = unlimited.** Preserves today's behavior if governance explicitly wants it; the default is the bounded value `8`, not `0`.
- **Scope of the cap: per request** (keyed by `request_hash`), matching how `ResponseCount` is already tracked. No global cap.
- **No dedup of identical responses** and **no request expiry** in this issue — those are separate concerns (see Out of scope). This change is purely the count bound.
- **No migration of existing responses:** the cap applies to new `RespondValidation` calls from the upgrade height forward; already-stored responses are untouched.

## Consensus safety

Deterministic: a single `uint64` comparison against a param inside the atomic `RespondValidation` handler. No wall-clock, randomness, map iteration, Begin/EndBlocker, or panic. Identical on every validator.

## Acceptance criteria

- With default params, the 9th `RespondValidation` for one request fails with `ErrTooManyValidationResponses`; the first 8 succeed with `seq` 0..7.
- Setting the param to `0` allows unbounded responses (regression/escape-hatch test).
- `DefaultParams()` includes `validation_max_responses_per_request = 8`; genesis export/import round-trips the field.
- Distinct requests each get their own independent budget (cap is per `request_hash`).
- `make proto-gen` produces no uncommitted diff; `go test ./x/agent/...` passes; `golangci-lint run` and `gofumpt -w .` clean.

## Out of scope

- Validation **request expiry / TTL** (no automatic pruning of old requests).
- Deduplication of byte-identical responses.
- Any change to `RequestValidation`, the feedback/reputation registries, or the escrow.
- Slashing or economic penalties for spammy validators.

---

---

omriai1:
Closes #2204

TLDR: Validation responses could be appended forever to a single request, allowing permanent state growth and leaving consumers without a bounded final record. This change adds a per-request cap (default 8, with 0 preserving unlimited behavior), activates it for existing chains at upgrade height, and rejects further writes with a dedicated module error.

- Adds `validation_max_responses_per_request` as `Params` field 10 and regenerates protobuf code.
- Enforces the cap before writing a response or incrementing `ResponseCount`.
- Bumps the module consensus version to 3 and backfills stored params through `Migrate2to3`.
- Covers the default boundary, zero escape hatch, per-request independence, no-write rejection, and genesis/protobuf round trips.

`RespondValidation` compares the request's existing `ResponseCount` with the current parameter after loading both records. A non-zero cap rejects counts at or above the limit with `ErrTooManyValidationResponses`; zero skips the guard. Existing responses are untouched. `Migrate2to3` stamps the default value onto already-stored params, so the bound becomes effective for existing chains at upgrade height.

The value is set at genesis and can be changed only by a software upgrade with a params migration; `x/agent` does not currently expose a governance `MsgUpdateParams` route.

</details>

## Proof of execution

- **Build and Test — local green; CI pending:** `go test -race ./...` passed, including `x/agent/keeper` in 15.947s; `go build -v ./...` exited 0.
- **Agent regression suite — local green; CI pending:** `go test -count=1 ./x/agent/...` passed all four agent packages after the final protobuf regeneration.
- **Upgrade backfill — local green; CI pending:** `TestMigrate2to3_DefaultsValidationResponseCap` proves a pre-upgrade stored zero is replaced with 8; the focused version 1→2 and 2→3 migration tests both pass.
- **Protobuf Checks — local green; CI pending:** `make proto-format` and repeated `make proto-gen` runs produced a clean post-commit worktree.
- **Buf breaking compatibility — CI green:** `buf-break-check` passed in 14s against the immutable PR base SHA via the local Git input; the workflow no longer performs the failing second network clone.
- **Agent lint — local green; CI pending:** `golangci-lint run ./x/agent/...` reports `0 issues` after the migration changes.
- **golangci-lint — CI pending:** local golangci-lint v2.12.2 reports seven baseline findings reproduced unchanged at merge base `945d7361`; none is in this PR's changed files.

<details><summary>Baseline lint reproduction</summary>

The same `golangci-lint run` command was executed in an isolated detached worktree at merge base `945d7361` and returned the identical seven findings:

```text
cmd/dymd/cmd/debug.go:156:39: G115: integer overflow conversion int64 -> byte (gosec)
utils/rand/testutils.go:14:17: G115: integer overflow conversion int -> byte (gosec)
x/dymns/types/sell_order_test.go:696:38: G602: slice index out of range (gosec)
x/forward/cli/kaspaaddr/kaspaaddr.go:62:31: G115: integer overflow conversion rune -> byte (gosec)
x/rollapp/keeper/invariants.go:64:5: QF1012: use fmt.Fprintf (staticcheck)
x/rollapp/keeper/invariants.go:71:5: QF1012: use fmt.Fprintf (staticcheck)
x/rollapp/keeper/invariants.go:76:5: QF1012: use fmt.Fprintf (staticcheck)
7 issues: gosec: 4; staticcheck: 3
```

</details>

**Not verified:** no live-node transaction was broadcast; the changed Cosmos message handler and persistent collections are exercised directly by the committed keeper integration tests, and this change adds no external dependency or endpoint.


omriai2 (changes requested):
I have everything I need. Here is my review.

---

**REQUEST_CHANGES** — the guard itself is correct, well-tested, and faithfully implements the issue's spec, but it silently fails open on every upgrade path because the new parameter is never backfilled onto already-stored params, and the module has no governance message that could turn it on afterwards.

### Findings

1. **New `Params` field ships with no consensus-version bump and no migrator backfill, so the cap decodes as `0` (unlimited) on any chain that already has agent params in state.** — *Accept when:* `ConsensusVersion()` is bumped to 3, a registered `Migrate2to3` stamps `DefaultValidationMaxResponsesPerRequest` onto stored params, and a test asserts a pre-upgrade `Params` without field 10 ends up at 8 after migration.
2. **The PR body's headline claim that the cap is "governance-tunable" is false — `x/agent` has no `MsgUpdateParams` and no param subspace.** — *Accept when:* the PR body is corrected to describe the actual tuning path, or `MsgUpdateParams` is added so the claim becomes true.

Non-blocking nit (do not hold the merge for it) is in the last fold.

---

<details><summary>Finding 1 — New Params field ships with no consensus-version bump and no migrator backfill; the cap is inert on upgrade</summary>

**The problem.** This change adds `validation_max_responses_per_request` as `Params` field 10 and gives it a default of 8 in `DefaultParams()`. Under proto3 semantics, a `Params` record that was serialized before field 10 existed decodes with that field as `0` — and the new guard treats `0` as "unlimited". So on any chain whose `x/agent` params are already written to state, the stored value is `0`, the guard is skipped entirely, and the state-growth bound the issue asked for does not exist. The default of 8 only ever takes effect on a chain that runs `InitGenesis` from scratch after this build.

**Why this is not speculative hardening — the repo established the exact opposite convention one commit ago.** The merge base of this PR, `945d7361` (#2197), added `spend_recipient_allowlist_max` as field 9 to this same `Params` message. That commit did three things: added the field, bumped the module's `ConsensusVersion` from 1 to 2, and introduced `Migrate1to2`, whose entire purpose is stamping the newly-added validation params onto already-stored records. `Migrate1to2` was created by that commit and by nothing else:

```
$ git log --oneline -3 -- x/agent/keeper/migrations.go
945d7361 feat(agent): restrict attested transfer recipients (#2197)
```

Its body backfills precisely the fields that #2199 and #2197 had added — the validation fee, the tag/URI byte caps, and the allowlist max — with the allowlist max stamped unconditionally:

```go
// Migrate1to2 initializes fields added to the agent params at version 2.
func (m Migrator) Migrate1to2(ctx sdk.Context) error {
	params, err := m.k.GetParams(ctx)
	...
	if params.ValidationUriMaxBytes == 0 {
		params.ValidationUriMaxBytes = defaults.ValidationUriMaxBytes
	}
	params.SpendRecipientAllowlistMax = types.DefaultSpendRecipientAllowlistMax
	return m.k.SetParams(ctx, params)
}
```

This PR adds field 10 to the same message and does neither half of that pattern. `ConsensusVersion()` is still `2`, and only the `1 → 2` migration is registered:

```
x/agent/module.go:114:	if err := cfg.RegisterMigration(types.ModuleName, 1, migrator.Migrate1to2); err != nil {
x/agent/module.go:135:func (AppModule) ConsensusVersion() uint64 { return 2 }
```

**The resulting state matrix.** A chain sitting at agent consensus version 1 (a build between #2166 and #2197) upgrades onto this code: `Migrate1to2` runs, backfills fields 6 through 9, and leaves field 10 at zero — cap inert. A chain sitting at agent consensus version 2 (current `main`, i.e. #2197 or #2199) upgrades onto this code: the consensus version is unchanged, so *no migration runs at all* — cap inert. Only a chain where the module is being introduced fresh, so that `RunMigrations` falls through to `InitGenesis` with `DefaultGenesis`, gets the cap. The existence of `Migrate1to2` is itself the evidence that the first two states are real: nobody writes an in-place param backfill for a module that is only ever genesis-initialized.

**Why this is blocking rather than a nit.** The thing being delivered is a guard against permanent, attacker-driven state growth, and it fails open — the failure mode is indistinguishable from not merging the PR at all, on exactly the deployments that already carry the bug. There is no runtime escape hatch to compensate (see Finding 2: no `MsgUpdateParams` exists, so an operator who notices cannot fix it without another software upgrade). And the fix is a one-line addition to a function the repo already has, following a precedent set in the immediately preceding commit. The issue's own framing expects this: it says the cap "applies to new `RespondValidation` calls **from the upgrade height forward**", which is only true if the param is non-zero at the upgrade height.

**Scope of the class.** The class is: adding a field to `x/agent` `Params` without pairing it with a consensus-version bump and a migrator backfill, on a module with no `MsgUpdateParams` fallback. Field 10 is the only `Params` field this diff adds, so the class has exactly one instance here — I checked, and no other new persisted field is introduced anywhere in the diff. But the same requirement binds every future `Params` addition in this module, which is why I am naming it as a class rather than as a one-off.

**Acceptance criterion.** Bump `ConsensusVersion()` to `3`; register a `Migrate2to3` on the configurator; have it set `ValidationMaxResponsesPerRequest` to `DefaultValidationMaxResponsesPerRequest`. Either the conditional form (`if params.ValidationMaxResponsesPerRequest == 0 { ... }`) or the unconditional stamp that `Migrate1to2` uses for `SpendRecipientAllowlistMax` is acceptable — since no on-chain path can set this param today, there is no operator-chosen `0` to clobber. Add a keeper test that writes a `Params` value with the field left at zero, runs the migrator, and asserts the stored value is 8. Update the PR body to state that the cap becomes effective at the upgrade height via this migration.

**What I verified.** The four validation keeper tests, including both new ones, pass on this branch:

```
$ go test ./x/agent/keeper/ -run 'TestValidationResponsesAreBoundedPerRequest|TestValidationResponsesAreUnlimitedWhenCapIsZero|TestValidationGenesisRoundTripAndRejectsOrphan|TestValidationRequestResponseProgressiveAndQueries' -count=1 -v
--- PASS: TestValidationRequestResponseProgressiveAndQueries (0.01s)
--- PASS: TestValidationResponsesAreBoundedPerRequest (0.00s)
--- PASS: TestValidationResponsesAreUnlimitedWhenCapIsZero (0.00s)
--- PASS: TestValidationGenesisRoundTripAndRejectsOrphan (0.00s)
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.179s
```

Those tests all construct params via `DefaultParams()` or `SetParams`, which is why none of them exercises — or could catch — the decode-as-zero upgrade path. That gap is what the requested migration test closes.

**One piece of context that bounds the severity rather than excusing it.** `agenttypes.ModuleName` is absent from `app/upgrades/v5/constants.go`'s `StoreUpgrades.Added` list, so the module is not scheduled to ship in the v5 upgrade as currently written, and the latest tag is `v4.0.2`. That means mainnet is not affected today. It is a pre-existing gap in a file this PR does not touch, so I am not raising it as a finding — but it is the reason I am treating this as blocking-normal rather than critical.

</details>

<details><summary>Finding 2 — PR body claims the cap is "governance-tunable"; no governance path to set it exists</summary>

**The problem.** The PR body's summary sentence reads: "This change adds a **governance-tunable** per-request cap (default 8, with 0 preserving unlimited behavior)". There is no mechanism by which governance can tune this parameter.

**Evidence.** The `x/agent` `Msg` service exposes fourteen RPCs and none of them is `UpdateParams`:

```
$ rg -n "rpc " proto/dymensionxyz/dymension/agent/tx.proto
RegisterAgent, DeactivateAgent, UpdateAgentPolicy, SubmitAttestedAction,
FundAgentEscrow, WithdrawAgentEscrow, UpdateAgentSpendPolicy,
SubmitAttestedTransfer, RevokePolicy, UnrevokePolicy, SubmitFeedback,
RevokeFeedback, RequestValidation, RespondValidation
```

The two gov-authority messages in that file (`RevokePolicy` / `UnrevokePolicy`) manage the policy denylist, not params. Every sibling module in this repo does have the message — `rollapp`, `sequencer`, `sponsorship`, `delayedack`, `streamer`, `eibc`, `dymns`, `incentives`, `iro`, `lockup`, and `otcbuyback` each declare `rpc UpdateParams(MsgUpdateParams)`. `x/agent` is the outlier. Nor is there a legacy fallback: `rg -n "ParamsKeeper.Subspace" app/*.go` returns nothing for agent, and grepping the module for writers shows `SetParams` is reachable from exactly two places — `InitGenesis` and the `Migrator`:

```
x/agent/keeper/genesis.go:11:	if err := k.SetParams(ctx, g.Params); err != nil {
x/agent/keeper/migrations.go:36:	return m.k.SetParams(ctx, params)
```

**Why this is blocking rather than cosmetic.** This is a claim about the operational controls available for a security parameter, recorded permanently in the merge commit and in the project's PR history. An operator reading it during an incident would go looking for a param-change proposal that does not exist, and would discover only then that changing this value requires cutting, testing, and coordinating a software upgrade. The claim also interacts with Finding 1: the absence of an `UpdateParams` route is precisely why the missing migration cannot be worked around after the fact, so getting the body right is what tells the next reader that the migration is load-bearing rather than a nicety.

**Acceptance criterion.** Either correct the PR body to describe the real tuning path — the value is set at genesis and changed only by a software upgrade with a migration, not by governance proposal — or add `MsgUpdateParams` with a gov-authority signer to the agent `Msg` service, matching the eleven sibling modules, and keep the wording. Correcting the body is the smaller and, given the issue's stated scope, the more appropriate of the two; adding `MsgUpdateParams` is a real capability addition and would deserve its own issue. Note that adding `MsgUpdateParams` alone does not resolve Finding 1 — a param nobody has proposed a change to is still `0` at the upgrade height.

</details>

<details><summary>Nit — cap check runs ahead of the responder authorization check (not blocking, no action required)</summary>

The new guard sits immediately after the `ValidationRequest` is loaded, which places it before the validator liveness checks and before `if msg.Responder != validator.Owner { return nil, types.ErrUnauthorized }`. The effect is that an arbitrary unauthorized address probing a request that has hit its cap receives `ErrTooManyValidationResponses` instead of `ErrUnauthorized`.

I considered and dismissed this as an information-disclosure issue: `ResponseCount` is already returned in the clear by the `ValidationRequest` gRPC query, so the error tells an attacker nothing they could not read for free without submitting a transaction. No state is written before the check, so there is no partial-write or rollback concern either. The placement also matches the issue's instruction to enforce "after loading `req` and `p`". Mentioning it only because a future reader may wonder; I would not hold the merge for it.

For completeness, the other things I examined in this diff and found clean: the regenerated `agent.pb.go` is exactly correct gogoproto output for a `uint64` at field 10 — struct tag, getter, `Size`, `Marshal` writing key byte `0x50`, and `Unmarshal` `case 10` with the varint wire-type guard — with every other removed line being gzipped `fileDescriptor` bytes, and `buf-break-check` passing on CI confirms no wire-compatibility break. The `errors.go` churn is pure gofumpt realignment plus code 25, which is unused and would panic at package init on collision. `gofumpt -l x/agent/` reports nothing. `Params.Validate()` deliberately does not constrain the field, matching the issue's explicit decision that any `uint64` is structurally valid with `0` as the documented escape hatch. No README or doc in the repo enumerates agent params, and `x/agent/doc.go` does not describe validation limits, so nothing became stale. The committed regression tests cover all four of the issue's behavioral acceptance criteria — the 0..7 sequence with the ninth rejected, the zero escape hatch, per-`request_hash` independence, and the genesis round trip.

On proof of execution: the section is present and every claim carries a receipt line, including an explicit owned "Not verified" entry for the absence of a live-node broadcast. This diff adds no external dependency call or endpoint, so the live-invocation requirement does not apply. Build, test, and lint claims are all CI-gated — `build`, `golangci-lint`, and `Analyze` were still pending when I checked, with `buf-break-check` already green — so a named suite plus CI is the proof and I did not open any transcript. `protobuf-checks` reports "skipping", so the proto-gen no-diff claim is not CI-covered on this run; I settled it instead by reading the regenerated descriptor changes directly, as described above, and am satisfied.

</details>

---

VERDICT: REQUEST_CHANGES

omriai2 (approved):
**APPROVE** — both prior blocking findings are fully resolved by `cecb0f53`: the cap is now backfilled onto already-stored params through a registered `Migrate2to3` behind a consensus-version bump, and the false "governance-tunable" claim has been replaced with an accurate description of the real tuning path.

### Disposition of prior findings

1. **Finding 1 (prior, blocking) — new `Params` field shipped with no consensus-version bump or migrator backfill, leaving the cap inert on upgrade.** — **RESOLVED.** `ConsensusVersion()` is 3, `Migrate2to3` is registered at version 2 and stamps the default, and `TestMigrate2to3_DefaultsValidationResponseCap` proves a stored zero becomes 8. I ran it green.
2. **Finding 2 (prior, blocking) — PR body claimed a "governance-tunable" cap that no `MsgUpdateParams` could tune.** — **RESOLVED.** The claim is gone from the TLDR and the implementation-details fold now states the genuine constraint explicitly.
3. **Observation, non-blocking — `buf-break-check` is red on this run due to a CI clone flake (`stat .git/objects/maintenance.lock: file does not exist`), not a proto change.** — Needs a re-run before merge; nothing for the author to fix.
4. **Observation, non-blocking — `Migrate2to3` stamps the default unconditionally, which would overwrite a genesis-imported `0`.** — This is the shape I pre-authorized last round and matches `Migrate1to2`; no action wanted.

---

<details><summary>Finding 1 — RESOLVED: consensus-version bump and migrator backfill</summary>

The fix implements exactly the acceptance criterion I stated, using the `Migrate1to2` precedent I named as acceptable.

`x/agent/module.go` bumps the version and registers the new migration alongside the existing one:

```go
	if err := cfg.RegisterMigration(types.ModuleName, 1, migrator.Migrate1to2); err != nil {
		panic(...)
	}
	if err := cfg.RegisterMigration(types.ModuleName, 2, migrator.Migrate2to3); err != nil {
		panic(fmt.Sprintf("failed to migrate x/agent from version 2 to 3: %v", err))
	}
...
-func (AppModule) ConsensusVersion() uint64 { return 2 }
+func (AppModule) ConsensusVersion() uint64 { return 3 }
```

`Migrate2to3` mirrors the unconditional stamp that `Migrate1to2` already uses for `SpendRecipientAllowlistMax`:

```go
// Migrate2to3 initializes fields added to the agent params at version 3.
func (m Migrator) Migrate2to3(ctx sdk.Context) error {
	params, err := m.k.GetParams(ctx)
	if err != nil {
		return err
	}
	params.ValidationMaxResponsesPerRequest = types.DefaultValidationMaxResponsesPerRequest
	return m.k.SetParams(ctx, params)
}
```

This closes both inert-cap states I laid out last round. A chain at agent consensus version 2 — current `main` — now runs `2 → 3` where previously no migration ran at all. A chain at version 1 runs `1 → 2` then `2 → 3` consecutively, so it lands at 8 as well; `Migrate1to2` did not need to change. A chain taking the module fresh still gets 8 via `InitGenesis` with `DefaultParams()`.

The regression test is committed beside the existing `TestMigrate1to2_DefaultsVersion2Params` in `escrow_test.go`, which is where this module's migration tests already live, and it asserts precisely the pre-upgrade-zero-becomes-eight transition I asked for:

```go
func (s *EscrowTestSuite) TestMigrate2to3_DefaultsValidationResponseCap() {
	params := types.DefaultParams()
	params.ValidationMaxResponsesPerRequest = 0
	s.Require().NoError(s.k.SetParams(s.Ctx, params))

	migrator := keeper.NewMigrator(*s.k)
	s.Require().NoError(migrator.Migrate2to3(s.Ctx))

	params, err := s.k.GetParams(s.Ctx)
	s.Require().NoError(err)
	s.Require().Equal(uint64(8), params.ValidationMaxResponsesPerRequest)
}
```

I ran both migration tests plus the three validation tests that carry the behavior, and they pass on `cecb0f53`:

```
$ go test ./x/agent/keeper/ -run 'TestEscrowTestSuite/TestMigrate1to2_DefaultsVersion2Params|TestEscrowTestSuite/TestMigrate2to3_DefaultsValidationResponseCap|TestValidationResponsesAreBoundedPerRequest|TestValidationResponsesAreUnlimitedWhenCapIsZero|TestValidationGenesisRoundTripAndRejectsOrphan' -count=1 -v
--- PASS: TestEscrowTestSuite (0.17s)
    --- PASS: TestEscrowTestSuite/TestMigrate1to2_DefaultsVersion2Params (0.09s)
    --- PASS: TestEscrowTestSuite/TestMigrate2to3_DefaultsValidationResponseCap (0.08s)
--- PASS: TestValidationResponsesAreBoundedPerRequest (0.00s)
--- PASS: TestValidationResponsesAreUnlimitedWhenCapIsZero (0.00s)
--- PASS: TestValidationGenesisRoundTripAndRejectsOrphan (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	0.358s
```

I also confirmed the bump breaks nothing downstream: `ConsensusVersion` appears in exactly one place in the module and nothing in `app/upgrades/` references `agenttypes.ModuleName`, so no upgrade handler or test pins the old value. `gofumpt -l x/agent/` reports nothing.

</details>

<details><summary>Finding 2 — RESOLVED: PR body corrected to describe the real tuning path</summary>

The TLDR previously read "adds a **governance-tunable** per-request cap". It now reads "adds a per-request cap (default 8, with 0 preserving unlimited behavior), **activates it for existing chains at upgrade height**, and rejects further writes with a dedicated module error", and the implementation-details fold states the constraint directly:

> The value is set at genesis and can be changed only by a software upgrade with a params migration; `x/agent` does not currently expose a governance `MsgUpdateParams` route.

That matches what I verified in the code last round — the agent `Msg` service declares fourteen RPCs and none is `UpdateParams`, there is no param subspace, and `SetParams` is reachable only from `InitGenesis` and the `Migrator`. The author chose the body-correction option rather than adding `MsgUpdateParams`, which is the right call for this issue's scope; adding that message is a real capability addition and belongs in its own change.

</details>

<details><summary>Observation 3 — `buf-break-check` red from a CI clone flake, not the diff</summary>

`buf-break-check` passed on the previous head (`5e6fa764`) and fails on `cecb0f53`. The answering commit touched three files — `x/agent/keeper/escrow_test.go`, `x/agent/keeper/migrations.go`, `x/agent/module.go` — and no proto file, so the input to the breaking-change comparison is byte-identical to the run that passed. The log confirms the cause is the action's own clone of the base ref failing, not a detected break:

```
Failure: could not clone https://github.com/dymensionxyz/dymension.git: stat .git/objects/maintenance.lock: file does not exist
```

Everything above that line is the same pre-existing `buf.yaml` transitive-dependency warning set that appears on passing runs too. This needs a re-run before the merge button goes green, but there is nothing here for the author to change, so I am not holding the review for it. `build`, `Analyze`, and `golangci-lint` were still pending when I checked; `label-prs` passed and `protobuf-checks` reports "skipping" as it did on the prior run.

</details>

<details><summary>Observation 4 — unconditional stamp in `Migrate2to3`, and everything else I re-checked this round</summary>

`Migrate2to3` sets the field unconditionally rather than only when zero, so a chain that genesis-imported an explicit `0` would be switched to `8` at the upgrade. I flagged this shape as acceptable in my prior review and I stand by that: there is no `MsgUpdateParams`, so no operator can have deliberately chosen `0` through an on-chain action, and `Migrate1to2` makes the identical tradeoff one line above for `SpendRecipientAllowlistMax`. Deviating from that precedent would be the more surprising choice. No action wanted.

Re-checked and unchanged from my prior pass: the guard in `RespondValidation` is byte-identical to what I reviewed, still placed after the request load and before any state write, so the ordering nit I raised last round (an unauthorized caller sees `ErrTooManyValidationResponses` rather than `ErrUnauthorized`, disclosing nothing that the `ValidationRequest` query does not already return in the clear) stands as a non-issue. The proto comment "Zero means unlimited" remains accurate — the migration ensures stored params are not zero; it does not change what zero means to the handler. `Migrate2to3`'s doc comment mirrors `Migrate1to2`'s phrasing and describes what the function does. No README or doc in the repo enumerates agent params, so nothing went stale.

On proof of execution: the section gained two bullets, both carrying receipts — an "Upgrade backfill" item naming `TestMigrate2to3_DefaultsValidationResponseCap` and the outcome of the focused 1→2 and 2→3 runs, and an "Agent lint" item reporting `0 issues` from `golangci-lint run ./x/agent/...`. The pre-existing bullets and the explicit owned "Not verified" entry for the absent live-node broadcast are intact. Every claim is CI-gated, so I opened no transcripts; I independently ran the two migration tests anyway since the backfill was the deciding evidence for Finding 1, and they pass. This diff still adds no external dependency call or endpoint, so the live-invocation requirement does not apply. There is no QA Brief section.

</details>

VERDICT: APPROVE

omriai2 (approved):
**APPROVE** — the only change since my approval is a three-line CI fix to the `buf-break-check` gate, which I verified is now comparing against the correct base SHA and reporting a genuine, non-vacuous pass; both prior blocking findings remain resolved with the Go code byte-identical to what I already ran green.

### Disposition

1. **Finding 1 (round 1, blocking) — new `Params` field shipped with no consensus-version bump or migrator backfill.** — **RESOLVED in `cecb0f53`, still resolved.** Untouched by this round's commit.
2. **Finding 2 (round 1, blocking) — PR body claimed a "governance-tunable" cap with no `MsgUpdateParams` behind it.** — **RESOLVED in `cecb0f53`, still resolved.** Body unchanged in the relevant sections.
3. **Observation 3 (round 2, non-blocking) — `buf-break-check` red from a CI clone flake.** — **RESOLVED in `4ae03b27`,** and the fix is a correctness improvement to the gate rather than a bypass. Verified from the job log.
4. **New observation, non-blocking — the CI workflow change is mild scope creep in a validation-cap PR.** — Cleanly isolated in its own conventional commit and it unblocks a required check; no action wanted.

---

<details><summary>Finding 3 — RESOLVED and verified non-vacuous: buf-break-check now compares against the local base SHA</summary>

The failure I described last round was the breaking-change action's own network clone of the base ref dying on `stat .git/objects/maintenance.lock: file does not exist`. `4ae03b27` removes that clone entirely and points buf at the checkout that is already on disk:

```yaml
       - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
...
-          against: "https://github.com/${{ github.repository }}.git#branch=${{ github.event.pull_request.base.ref }},ref=HEAD~1,subdir=proto"
+          against: ".git#ref=${{ github.event.pull_request.base.sha }},subdir=proto"
```

**Why I checked this beyond the green tick.** When the thing a commit changes *is* a required gate, a green gate is not evidence that the gate still works — a misconfigured `against` could pass vacuously. So I opened the job log, which is the one case where the CI-covered rule does not settle the question. It confirms both halves of what I needed:

```
against: .git#ref=945d73614b53a2acd7e1ed757c124cfe58e3af3b,subdir=proto
...
No breaking errors were found.
```

The SHA `945d7361` is exactly this PR's merge base — the commit that added `spend_recipient_allowlist_max` as field 9 — so the expression resolved to the right commit rather than to an empty or stale ref. And buf emitted its comparison verdict, meaning it built both image sets and diffed them; the previous run's failure mode proves the alternative is a loud job failure, not a silent pass.

**This strengthens the gate rather than weakening it.** The old `against` pinned `branch=<base ref>,ref=HEAD~1` — one commit *before* the tip of `main`. That baseline is both stale and unrelated to where the PR actually branched from, so a proto change landing in main's most recent commit would have been invisible to the comparison. The new baseline is the PR's real base SHA. `fetch-depth: 0` is what makes the local `.git` input resolvable and is correctly paired with the change.

**Things I checked for and did not find.** `github.event.pull_request.base.sha` is only populated on `pull_request` events, and the job carries `if: github.event_name == 'pull_request'`, so there is no path where the expression interpolates empty and turns `against` into a malformed no-op. The SHA is repo-side and 40 hex characters, not PR-author-controlled text, so interpolating it into the input string is not an injection surface. Workflow `permissions` stays at `contents: read`; no credential handling, `persist-credentials`, or action pin changed — the action is still `bufbuild/buf-breaking-action@v1.1.4`. The only residual edge is that a force-push rewriting `main` could make the recorded base SHA unreachable, which would fail the job loudly rather than pass it silently; that is acceptable behavior for a gate.

`protobuf-checks` still reports "skipping" — that job is label-gated behind `enable-proto-checks` and documented as disabled pending #1890, which is pre-existing and unrelated to this commit. Current check state:

```
buf-break-check  pass      14s
label-prs        pass       8s
protobuf-checks  skipping
build / golangci-lint / Analyze   pending
```

</details>

<details><summary>Findings 1 and 2 — still resolved; Go code untouched this round</summary>

`4ae03b27` changes one file and nothing else:

```
$ git diff cecb0f53 4ae03b27 --stat
 .github/workflows/proto.yaml | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)
```

So the migration work I verified last round is byte-identical: `ConsensusVersion()` at 3, `Migrate2to3` registered at version 2 stamping `DefaultValidationMaxResponsesPerRequest`, and `TestMigrate2to3_DefaultsValidationResponseCap` asserting a stored zero becomes 8. I ran those tests green on `cecb0f53` and re-running them here would test the same bytes, so I did not.

The PR body's correction for Finding 2 is also unchanged — the TLDR no longer claims the cap is governance-tunable, and the implementation-details fold still states plainly that the value is set at genesis and changeable only by a software upgrade with a params migration, with no `MsgUpdateParams` route in `x/agent`. That remains accurate against the code.

No new reviewer input to account for: the PR carries my two prior reviews (`CHANGES_REQUESTED`, then `APPROVED`) and has zero issue comments and zero inline review threads.

</details>

<details><summary>Observation 4 — scope creep in carrying a CI fix, and the proof section</summary>

Strictly, `.github/workflows/proto.yaml` is outside the subject of a PR titled "bound validation responses per request", and my prior round explicitly said the flake was "nothing for the author to fix". The author fixed the root cause anyway. I am not treating this as a finding: the check is required and was red, so the PR could not merge without it; the change is three lines confined to one job; it improves the gate's semantics rather than bypassing it; and it is carried in its own separate conventional commit (`fix(ci): compare protobufs against local base`) rather than smuggled into a functional commit, which is the right hygiene for an incidental fix. If the maintainers prefer such fixes to land standalone on `main`, that is a repo-convention call rather than a defect in this change.

On proof of execution: the section gained one bullet — "**Buf breaking compatibility — CI green:** `buf-break-check` passed in 14s against the immutable PR base SHA via the local Git input; the workflow no longer performs the failing second network clone." It carries a receipt naming the check, the duration, and the mechanism, and every claim in it matches what I independently observed in the job log, including the "immutable PR base SHA" characterization. The remaining bullets and the explicit owned "Not verified" entry for the absent live-node broadcast are unchanged from the version I already accepted. This diff still adds no external dependency call or endpoint, so the live-invocation requirement does not apply, and there is no QA Brief section. No README or doc in the repo describes the buf comparison baseline, so nothing became stale.

</details>

VERDICT: APPROVE

PR #2207, opened 2026-08-31
feat(ante): check circuit breaker for wrapped msgs

### LINKED ISSUE #1827: Enhance circuit breaker decorator to check for wrapped msgs
from cosmos-sdk docs:

> The `CircuitBreakerDecorator` works for most use cases, but [does not check the inner messages of a transaction](https://docs.cosmos.network/main/learn/beginner/tx-lifecycle#antehandler). This some transactions (such as `x/authz` transactions or some `x/gov` transactions) may pass the ante handler. **This does not affect the circuit breaker** as the message router check will still fail the transaction.
This tradeoff is to avoid introducing more dependencies in the `x/circuit` module. Chains can re-define the `CircuitBreakerDecorator` to check for inner messages if they wish to do so.

It's not high priority, as circuit breaker checked in baseapp as well.

## AI triage
Research complete. Here is the re-triage.

---

## Research summary

The issue (filed by mtsitrin, a repo contributor) asks to close a known gap in the Cosmos SDK circuit breaker: the stock `CircuitBreakerDecorator` from `cosmossdk.io/x/circuit/ante` only checks a transaction's **top-level** message type URLs against the tripped-message list. A tripped message wrapped inside `authz.MsgExec`, a gov `MsgSubmitProposal`, or a group `MsgSubmitProposal` passes the ante handler untouched. This is not a security hole — the baseapp message router re-checks the circuit breaker at execution time and fails the transaction there — which is why the filer marked it low priority. The ask is defense in depth: reject such transactions at the ante stage (mempool admission) instead of letting them into blocks only to fail at execution.

The single comment on the thread is the human update this re-triage folds in: per mtsitrin's follow-up, the fix "can probably be easily done" by implementing the check as a `Predicate` for the repo's own `RejectMessagesDecorator`. **No implementation work has been done** — the comment is a design pointer, not a progress report; no PR references this issue (verified via the issue timeline). All implementation remains and is owned by the dispatched worker.

I verified the suggested route against the clone (`app/ante/`), and it is confirmed sound:

- Both ante chains (`cosmos_handler.go`, `eth_handler.go`) install the stock `circuitante.NewCircuitBreakerDecorator(options.CircuitKeeper)`, which loops only over `tx.GetMsgs()` — top level only.
- The repo already has `RejectMessagesDecorator` (`app/ante/reject_msgs.go`) in the cosmos ante chain, which **recursively** walks inner messages of `authz.MsgExec`, gov `MsgSubmitProposal`, and group `MsgSubmitProposal` (bounded by `maxInnerDepth = 6`), applying a list of `Predicate` functions at every depth. It also checks `authz.MsgGrant` authorization type URLs against the same predicates.
- The one wrinkle in the comment's suggestion: the current `Predicate` type is `func(typeURL string, depth int) bool` — it receives neither `ctx` nor returns an error, while the circuit keeper's check is `IsAllowed(ctx, typeURL) (bool, error)`. The predicate signature (a type alias with two production call sites and two test call sites) must be extended to accept the context. `checkMsg` already threads `ctx`, so this is contained.
- Existing test coverage: `app/ante/ante_test.go` has `TestCircuitBreaker` exercising top-level trip/reset through the full ante chain, and `reject_msgs_test.go` covers the predicate mechanism — both give the worker a template for wrapped-message tests.
- The EVM ante chain does not need the new check: `MsgEthereumTx` cannot wrap cosmos messages, and the cosmos chain already blocks `MsgEthereumTx` at all depths.

## Root cause hypothesis

Not a bug — a known upstream design tradeoff, confirmed by the SDK docs quoted in the issue body: the stock `CircuitBreakerDecorator` deliberately skips inner messages to avoid extra module dependencies, and chains are expected to re-implement the check themselves if they want ante-stage coverage. Dymension installed the stock decorator unmodified, so wrapped tripped messages are admitted to the mempool and only fail at baseapp execution.

## Areas involved

- `app/ante/reject_msgs.go` — extend the `Predicate` type to receive `sdk.Context` (it already receives `typeURL` and `depth`); add a circuit-breaker predicate constructed from the `circuitante.CircuitBreaker` interface that returns reject when `IsAllowed` returns false **or errors** (fail closed; the swallowed error is acceptable since baseapp re-checks and surfaces it at execution). Check at all depths — re-checking depth 0 duplicates the stock decorator but keeps the predicate branch-free.
- `app/ante/cosmos_handler.go` — wire the new predicate into the existing `NewRejectMessagesDecorator().WithPredicate(...)` chain using `options.CircuitKeeper` (already present in `HandlerOptions`).
- `app/ante/reject_msgs_test.go`, `app/ante/ante_test.go` — update existing predicate call sites for the new signature; add wrapped-message coverage.
- Accepted side effect, intended behavior: because `RejectMessagesDecorator` also screens `authz.MsgGrant` authorization type URLs through the predicates, granting authz for a currently-tripped message type will also be rejected while the breaker is tripped. This is consistent defense in depth, not a regression.
- Keep the stock `circuitante.NewCircuitBreakerDecorator` in both chains unchanged — the new predicate supplements it, minimizing drift from upstream.

## Acceptance criteria

1. A transaction whose **top-level** message type is tripped is rejected in the ante handler (existing behavior, `TestCircuitBreaker` still passes).
2. A transaction carrying a tripped message type **wrapped** in `authz.MsgExec` is rejected in the ante handler, proven by a new test that trips a type via the circuit msg server, submits the wrapped tx through the full ante chain, and asserts rejection.
3. Same for a tripped type nested inside a gov `MsgSubmitProposal` (the group path shares the identical code branch; covering it is cheap but authz + gov are the required cases).
4. After `ResetCircuitBreaker`, the same wrapped transactions pass the ante handler (no false rejection when nothing is tripped).
5. A circuit-keeper error during the check rejects the transaction rather than admitting it.
6. `go test ./app/...` passes; `golangci-lint run` and `gofumpt` clean on modified files.

## Open questions

None.

omriai1:
Closes #1827

The Cosmos SDK's `CircuitBreakerDecorator` only inspects a transaction's top-level messages — a deliberate upstream tradeoff to avoid extra module dependencies in `x/circuit`, with the SDK docs telling chains to re-implement the check themselves if they want it. So a tripped message type wrapped in an authz `MsgExec` or a gov/group `MsgSubmitProposal` passed our ante handler untouched and was admitted to the mempool, only failing later in baseapp's message router at execution time. This was never a security hole — baseapp always caught it — but tripped transactions were reaching blocks just to fail there.

This change closes the gap at the ante stage by reusing the decorator we already have for exactly this shape of problem, rather than forking the SDK's decorator.

<details><summary>Mechanics</summary>

The repo already has `RejectMessagesDecorator`, which recursively walks the inner messages of authz `MsgExec` and gov/group `MsgSubmitProposal` (bounded by `maxInnerDepth`), applying a list of predicates at every depth. The circuit breaker check is now just another predicate on it.

- Added a predicate built from the `circuitante.CircuitBreaker` interface that consults `IsAllowed` for the message type. It applies at every depth; re-checking depth 0 duplicates the stock decorator but keeps the predicate branch-free.
- `Predicate` returns `(blocked bool, err error)`. A policy rejection and a failure to evaluate policy are different outcomes and must not be conflated: the ante handler is terminal, so an error discarded here is lost for good rather than resurfaced later. A circuit keeper error still fails closed, but the original error propagates wrapped, so `errors.Is` and operator diagnosis keep working — matching how the upstream SDK decorator returns `IsAllowed` errors directly.
- Both predicate evaluation points — the message type and the type an `authz.MsgGrant` authorizes — go through one helper, so the error and rejection handling cannot drift apart.
- The `Predicate` signature also takes an `sdk.Context`, which the keeper lookup needs. `checkMsg` already threaded the context, so this was contained to the type alias and its call sites.
- The SDK's `circuitante.NewCircuitBreakerDecorator` stays installed unchanged in both ante chains — the predicate supplements it, keeping drift from upstream minimal.
- The EVM ante chain needs nothing: `MsgEthereumTx` cannot wrap cosmos messages, and the cosmos chain already blocks `MsgEthereumTx` at all depths.

Intended side effect: because `RejectMessagesDecorator` also screens `authz.MsgGrant` authorization type URLs through the same predicates, granting authz for a currently-tripped message type is rejected while the breaker is tripped. That is consistent defense in depth, not a regression.

</details>

## Proof of execution

- **Wrapped tripped msg is rejected at ante (authz `MsgExec` and gov `MsgSubmitProposal`)** — `TestCircuitBreakerWrappedMsgs`, green in the `build` CI suite.
- **Top-level rejection unchanged** — pre-existing `TestCircuitBreaker`, green in the `build` CI suite.
- **No false rejection when nothing is tripped, before trip and after reset** — asserted inside `TestCircuitBreakerWrappedMsgs`, green in the `build` CI suite.
- **Circuit keeper error fails closed AND stays diagnosable, on both the msg and the grant path** — `TestCircuitBreakerPredicateFailsClosed`, green in the `build` CI suite.
- **Full repository suite** — `go test ./...`, 64 packages ok, 0 failures, green in the `build` CI suite.
- **Lint and format** — `golangci-lint run ./app/...` reports 0 issues; `gofumpt -l .` clean. Covered by the `golangci-lint` CI suite. See the fold below for pre-existing whole-repo findings unrelated to this change.
- **Both new tests genuinely catch their bug** — negative controls, not covered by CI; evidence folded below.

<details><summary>Negative control 1: wrapped-msg tests fail without the recursive check</summary>

With the production wiring reverted (predicate removed from the cosmos ante chain) and the new tests left in place, both wrapped cases fail — the wrapped message sails past `RejectMessagesDecorator` and only stops at the fee decorator, which is precisely the reported behavior:

```
--- FAIL: TestAnteTestSuite/TestCircuitBreakerWrappedMsgs (0.16s)
    --- FAIL: TestAnteTestSuite/TestCircuitBreakerWrappedMsgs/authz_MsgExec (0.10s)
        Error: Error "fee payer address:  does not exist: unknown address" does not contain "disabled: /cosmos.bank.v1beta1.MsgSend"
    --- FAIL: TestAnteTestSuite/TestCircuitBreakerWrappedMsgs/gov_v1_MsgSubmitProposal (0.06s)
        Error: Error "fee payer address:  does not exist: unknown address" does not contain "disabled: /cosmos.bank.v1beta1.MsgSend"
FAIL	github.com/dymensionxyz/dymension/v3/app/ante	0.335s
```

Each wrapper case builds its own app so circuit breaker state cannot leak between subtests.

</details>

<details><summary>Negative control 2: error-preservation tests fail if the keeper error is swallowed</summary>

Restoring the collapsed `return err != nil || !allowed, nil` form makes both paths fail, confirming the test locks in error preservation rather than merely rejection:

```
--- FAIL: TestAnteTestSuite/TestCircuitBreakerPredicateFailsClosed (0.16s)
    --- FAIL: TestAnteTestSuite/TestCircuitBreakerPredicateFailsClosed/msg (0.10s)
        Error: Target error should be in err chain:
    --- FAIL: TestAnteTestSuite/TestCircuitBreakerPredicateFailsClosed/authz_MsgGrant_authorization (0.06s)
        Error: Target error should be in err chain:
FAIL	github.com/dymensionxyz/dymension/v3/app/ante	0.320s
```

With the fix in place the whole `app/ante` package passes:

```
--- PASS: TestAnteTestSuite (0.82s)
    --- PASS: TestAnteTestSuite/TestCircuitBreaker (0.14s)
    --- PASS: TestAnteTestSuite/TestCircuitBreakerPredicateFailsClosed (0.12s)
        --- PASS: TestAnteTestSuite/TestCircuitBreakerPredicateFailsClosed/msg (0.06s)
        --- PASS: TestAnteTestSuite/TestCircuitBreakerPredicateFailsClosed/authz_MsgGrant_authorization (0.06s)
    --- PASS: TestAnteTestSuite/TestCircuitBreakerWrappedMsgs (0.11s)
        --- PASS: TestAnteTestSuite/TestCircuitBreakerWrappedMsgs/authz_MsgExec (0.05s)
        --- PASS: TestAnteTestSuite/TestCircuitBreakerWrappedMsgs/gov_v1_MsgSubmitProposal (0.05s)
ok  	github.com/dymensionxyz/dymension/v3/app/ante
```

</details>

<details><summary>Pre-existing whole-repo lint findings, reproduced on the merge base</summary>

`golangci-lint run` over the whole repo reports 7 issues (4 gosec, 3 staticcheck) in `x/dymns`, `x/forward` and `x/rollapp` — none of which this PR touches. Reproduced identically with these changes stashed, at merge base `945d7361`:

```
x/rollapp/keeper/invariants.go:76:5: QF1012: Use fmt.Fprintf(...) instead of WriteString(fmt.Sprintf(...)) (staticcheck)
7 issues:
* gosec: 4
* staticcheck: 3
```

Scoped to the changed package, `golangci-lint run ./app/...` reports `0 issues`.

</details>

<details><summary>Shared-fixture scope</summary>

`Predicate` is a type alias whose every consumer lives inside `app/ante`: one production construction site (`cosmos_handler.go`), three in tests, and the two invocation points now unified in `checkPredicates`. Verified by searching the repo for `Predicate`, `BlockTypeUrls`, `WithPredicate` and `BlockTrippedByCircuitBreaker`; no consumers outside `app/ante`. The full `go test ./...` run above covers every consuming package.

</details>

### DIFF
--- a/app/ante/ante_test.go
+++ b/app/ante/ante_test.go
@@ -13,12 +13,15 @@ import (
 	circuitkeeper "cosmossdk.io/x/circuit/keeper"
 	circuittypes "cosmossdk.io/x/circuit/types"
 	"github.com/cosmos/cosmos-sdk/client"
+	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/tx/signing"
 	"github.com/cosmos/cosmos-sdk/x/auth/migrations/legacytx"
+	"github.com/cosmos/cosmos-sdk/x/authz"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
+	govtypesv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
 	"github.com/evmos/ethermint/crypto/ethsecp256k1"
 	"github.com/evmos/ethermint/ethereum/eip712"
 
@@ -121,6 +124,72 @@ func (suite *AnteTestSuite) TestCircuitBreaker() {
 	// Try again - should pass
 	_, err = suite.anteHandler(suite.ctx, suite.txBuilder.GetTx(), false)
 	suite.Require().NotContains(err.Error(), "tx type not allowed") // not circuit breaker error
+}
+
+// The SDK's CircuitBreakerDecorator only inspects top level msgs, so wrapped ones are covered
+// by our RejectMessagesDecorator predicate instead.
+func (suite *AnteTestSuite) TestCircuitBreakerWrappedMsgs() {
+	msgSend := &banktypes.MsgSend{
+		FromAddress: "cosmos1...",
+		ToAddress:   "cosmos1...",
+		Amount:      sdk.NewCoins(sdk.NewInt64Coin("stake", 1000)),
+	}
+	sendURL := sdk.MsgTypeURL(msgSend)
+
+	wrap := map[string]func(*codectypes.Any) sdk.Msg{
+		"authz MsgExec": func(inner *codectypes.Any) sdk.Msg {
+			return &authz.MsgExec{Grantee: "cosmos1...", Msgs: []*codectypes.Any{inner}}
+		},
+		"gov v1 MsgSubmitProposal": func(inner *codectypes.Any) sdk.Msg {
+			return &govtypesv1.MsgSubmitProposal{
+				Messages:       []*codectypes.Any{inner},
+				InitialDeposit: sdk.NewCoins(sdk.NewInt64Coin("stake", 1000)),
+				Proposer:       "cosmos1...",
+			}
+		},
+	}
+
+	for name, wrapFn := range wrap {
+		suite.Run(name, func() {
+			// fresh app per case, so circuit breaker state cannot leak between them
+			suite.SetupTestCheckTx(false)
+
+			circuitMsgServ := circuitkeeper.NewMsgServerImpl(suite.app.CircuitBreakerKeeper)
+			authority, err := suite.app.AccountKeeper.AddressCodec().BytesToString(suite.app.CircuitBreakerKeeper.GetAuthority())
+			suite.Require().NoError(err)
+
+			msg := wrapFn(packMsg(suite.T(), msgSend))
+			runAnte := func() error {
+				txBuilder := suite.clientCtx.TxConfig.NewTxBuilder()
+				txBuilder.SetGasLimit(200000)
+				suite.Require().NoError(txBuilder.SetMsgs(msg))
+				_, err := suite.anteHandler(suite.ctx, txBuilder.GetTx(), false)
+				return err
+			}
+
+			if err := runAnte(); err != nil {
+				suite.Require().NotContains(err.Error(), "disabled: "+sendURL)
+			}
+
+			_, err = circuitMsgServ.TripCircuitBreaker(suite.ctx, &circuittypes.MsgTripCircuitBreaker{
+				Authority:   authority,
+				MsgTypeUrls: []string{sendURL},
+			})
+			suite.Require().NoError(err)
+
+			suite.Require().ErrorContains(runAnte(), "disabled: "+sendURL)
+
+			_, err = circuitMsgServ.ResetCircuitBreaker(suite.ctx, &circuittypes.MsgResetCircuitBreaker{
+				Authority:   authority,
+				MsgTypeUrls: []string{sendURL},
+			})
+			suite.Require().NoError(err)
+
+			if err := runAnte(); err != nil {
+				suite.Require().NotContains(err.Error(), "disabled: "+sendURL)
+			}
+		})
+	}
 }
 
 func (suite *AnteTestSuite) CreateTestEIP712CosmosTxBuilder(
--- a/app/ante/cosmos_handler.go
+++ b/app/ante/cosmos_handler.go
@@ -30,7 +30,9 @@ func newCosmosAnteHandler(options HandlerOptions) sdk.AnteHandler {
 			WithPredicate(BlockTypeUrls(
 				0,
 				sdk.MsgTypeURL(&evmtypes.MsgEthereumTx{}),
-				sdk.MsgTypeURL(&ibcclienttypes.MsgSubmitMisbehaviour{}))), // blocked to avoid skipping our validation logic in lightclient ante handler
+				sdk.MsgTypeURL(&ibcclienttypes.MsgSubmitMisbehaviour{}))). // blocked to avoid skipping our validation logic in lightclient ante handler
+			// extends the circuit breaker above, which only sees top level msgs
+			WithPredicate(BlockTrippedByCircuitBreaker(options.CircuitKeeper)),
 
 		// Use Mempool Fee TransferEnabledDecorator from our txfees module instead of default one from auth
 		mempoolFeeDecorator,
--- a/app/ante/reject_msgs.go
+++ b/app/ante/reject_msgs.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	errorsmod "cosmossdk.io/errors"
+	circuitante "cosmossdk.io/x/circuit/ante"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/cosmos/cosmos-sdk/x/authz"
@@ -23,7 +24,7 @@ type RejectMessagesDecorator struct {
 }
 
 // Predicate should return true if message is not allowed
-type Predicate = func(typeURL string, depth int) bool
+type Predicate = func(ctx sdk.Context, typeURL string, depth int) bool
 
 // Blocks any message with depth of depthMax OR MORE
 // Depth 0 is top level message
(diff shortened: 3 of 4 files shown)


omriai2 (changes requested):
REQUEST_CHANGES

The recursive circuit-breaker coverage is sound, but keeper failures are incorrectly converted into “message disabled” errors, hiding the underlying state failure.

1. Keeper errors are swallowed — preserve and return the original error; update the predicate abstraction and regression test accordingly.

<details><summary>Finding 1 — Circuit-keeper failures are misclassified and permanently hidden</summary>

`BlockTrippedByCircuitBreaker` collapses both legitimate circuit trips and `IsAllowed` failures into the same Boolean rejection:

```go
allowed, err := ck.IsAllowed(ctx, url)
return err != nil || !allowed
```

`checkMsg` then reports either `disabled: <type>` or `disabled grant: <type>`. Because this occurs in the ante handler, processing stops immediately. Contrary to the new comment and PR description, baseapp cannot subsequently execute the message and surface the real keeper error.

This affects the full class of stateful predicates: the Boolean-only `Predicate` contract cannot distinguish a policy rejection from a failure to evaluate policy. For circuit-state read failures, that removes the actionable error, misleads clients and operators into believing the circuit was deliberately tripped, and defeats normal error classification and diagnosis. The upstream SDK decorator preserves this distinction by returning `IsAllowed` errors directly.

Evidence:

- `app/ante/reject_msgs.go:48-52` discards `err`.
- `app/ante/reject_msgs.go:103-107` and `app/ante/reject_msgs.go:125-129` replace it with a synthetic disabled-message error.
- `app/ante/reject_msgs_test.go:203-224` explicitly locks in this masking behavior.
- The pinned SDK implementation returns the keeper error unchanged:

```go
isAllowed, err := cbd.circuitKeeper.IsAllowed(ctx, sdk.MsgTypeURL(msg))
if err != nil {
    return ctx, err
}
```

Source: [Cosmos SDK v0.50.14 circuit decorator](https://github.com/cosmos/cosmos-sdk/blob/v0.50.14/x/circuit/ante/circuit.go)

Acceptance criterion: represent predicate evaluation errors separately from policy rejections—such as changing `Predicate` to return `(blocked bool, err error)` and propagating `err` from both message and grant checks, or implementing the recursive circuit check in an error-returning decorator. Retain fail-closed behavior, but return the original/wrapped keeper error. Update the regression test to assert that the keeper error remains discoverable with `errors.Is`/`ErrorContains`, and correct the comment claiming baseapp will surface it later.

Review evidence: PR proof shape is complete; Build and Test, lint, and CodeQL checks are green. No full suite was rerun locally, per review policy.

</details>

VERDICT: REQUEST_CHANGES

omriai2 (approved):
APPROVE

The prior blocker is fully resolved: predicate evaluation errors now fail closed while preserving their original identity, with coverage for both message and authz-grant paths.

1. Prior finding resolved — keeper errors are propagated distinctly from deliberate circuit-breaker rejections, satisfying the acceptance criterion.

<details><summary>Finding 1 — Prior keeper-error blocker resolved</summary>

The fix commit `9b81af842372f9b236ca877a17a5f1c6211ee1b5` changes `Predicate` to return `(blocked bool, err error)` and centralizes handling in `checkPredicates`.

The implementation now:

- Propagates `IsAllowed` failures wrapped with message-type context.
- Preserves the original error for `errors.Is`.
- Avoids incorrectly labeling evaluation failures as `disabled`.
- Applies identical handling to ordinary messages and `authz.MsgGrant` authorization types.
- Removes the inaccurate claim that baseapp would later surface a swallowed error.

The updated `TestCircuitBreakerPredicateFailsClosed` covers both evaluation sites and asserts rejection, `ErrorIs(err, errCircuitUnavailable)`, and absence of the misleading `disabled` label.

No new blocking issues were introduced. The proof section remains substantive, and current build, lint, CodeQL, and coverage checks are green. No full local suite was rerun.

</details>

VERDICT: APPROVE

PR #2208, opened 2026-08-31
fix(rollapp): validate fraud proposal fields

### LINKED ISSUE #1558: Incomplete validation for MsgRollappFraudProposal
In x/rollapp/types/message_fraud_proposal.go:14-32, the ValidateBasic method of the MsgRollappFraudProposal is incomplete. It does not validate fields such as RollappId, RollappRevision, FraudHeight, and PunishSequencerAddress to ensure they are correctly set and valid.
Without these validations, an invalid fraud proposal could be accepted for voting, run the full proposal lifecycle, only to fail during execution. This creates inefficiencies and risks undermining the governance process.

## AI triage
All research is done. Here is the re-triaged brief.

---

## Research summary

The original report (from omritoptix, Nov 2024) says `ValidateBasic` on `MsgRollappFraudProposal` in `x/rollapp/types/message_fraud_proposal.go` is incomplete: it validates only the `Authority` and `Rewardee` addresses, leaving `RollappId`, `FraudRevision` (the report calls it "RollappRevision"), `FraudHeight`, and `PunishSequencerAddress` unchecked. Because this message is executed via governance, a malformed proposal passes submission and voting and only fails at execution, wasting a full proposal lifecycle.

**Confirmed against the clone (current `main`):** the gap still exists exactly as reported. `ValidateBasic` checks only the two bech32 addresses; `RollappId`, `FraudHeight`, and `PunishSequencerAddress` are never validated at message level. No open or merged PR addresses this issue. Two concrete consequences verified in the code:

- `SubmitRollappFraud` calls `HardFork(ctx, msg.RollappId, msg.FraudHeight-1)`. A `FraudHeight` of 0 underflows to max-uint64 before any check catches it. Because the pending-state revert then computes `lastValidHeight+1`, this wraps back to 0 — meaning a zero-height proposal is not merely rejected late, it takes a nonsensical path through the hard-fork logic.
- The existing rewardee check has a copy-paste bug: its error message interpolates `m.Authority` instead of `m.Rewardee`.

**The human update, folded in:** the only comment is from `mtsitrin` (repo contributor), linking this issue to #1838, which states that message validation should be driven by the `MsgServer` handler rather than relying on baseapp's `ValidateBasic` call (relevant because the repo is on cosmos-sdk v0.50, where baseapp no longer universally invokes `ValidateBasic`). Verified in the clone: the fraud-proposal handler `SubmitRollappFraud` **already calls `msg.ValidateBasic()` explicitly** as its second step. So enriching `ValidateBasic` is fully consistent with #1838's direction — the handler drives the call; no baseapp reliance is introduced. Nothing has been implemented or fixed since the report; mtsitrin's update adds direction, not completed work. Remaining work is the whole fix, owned by whoever this issue is dispatched to; the broader `ValidateBasic`→handler migration across all messages stays with #1838 and is out of scope here.

**Verified handler context (bounds the scope):** stateful checks — rollapp existence, revision-for-height match, sequencer existence — already live in the handler and belong there; they need *state* and cannot move to `ValidateBasic`. `MustRewardee()` silently discards the bech32 parse error, which is safe only because `ValidateBasic` pre-validates `Rewardee` — the handler's explicit `ValidateBasic` call is what keeps that sound. Also verified: rollapps are created with an initial `Revision{Number: 0}`, so `FraudRevision == 0` is legitimate and must **not** be rejected; correctness of the revision value is a stateful check the handler already performs.

## Root cause hypothesis

Not a hypothesis — confirmed by reading the code. `ValidateBasic` was written to cover only the address fields and was never extended to the remaining message fields. The stateless checks that can catch a malformed proposal before it enters the governance lifecycle are simply missing.

## Areas involved

- `x/rollapp/types/message_fraud_proposal.go` — the fix itself: extend `ValidateBasic`.
- `x/rollapp/types/chain_id.go` — existing `NewChainID` validator; the established pattern for validating rollapp IDs statelessly (used by `Rollapp.ValidateBasic`).
- `x/rollapp/keeper/fraud_proposal.go` — read-only context: the handler that invokes `ValidateBasic`; note it currently swallows the wrapped validation error detail (`errorsmod.Wrap(gerrc.ErrInvalidArgument, "invalid msg")` discards `err`) — including the underlying error in that wrap is an in-scope one-line improvement since the issue is about surfacing invalid proposals usefully.
- New/existing tests: a table-driven `ValidateBasic` unit test in `x/rollapp/types` (none exists for this message today); `x/rollapp/keeper/fraud_proposal_test.go` exists for the handler path if adjustments are needed.

## Acceptance criteria

1. `ValidateBasic` on `MsgRollappFraudProposal` rejects, with `gerrc.ErrInvalidArgument`-classed errors matching the file's existing style:
   - `RollappId` that fails `types.NewChainID` validation (covers empty and malformed IDs).
   - `FraudHeight == 0` (prevents the `FraudHeight-1` underflow into `HardFork`).
   - Non-empty `PunishSequencerAddress` that is not a valid bech32 account address (empty stays legal — it is documented as optional).
2. `FraudRevision` is **not** given a lower-bound check — revision 0 is valid; revision correctness remains the handler's stateful check.
3. The rewardee error message interpolates `m.Rewardee`, not `m.Authority`.
4. Existing valid-message flows still pass: the keeper test suite for `SubmitRollappFraud` remains green.
5. Table-driven unit tests cover each new rejection case plus a fully-valid message.
6. `go build ./...` and the rollapp module tests pass; linters clean.

## Open questions

None.

omriai1:
Closes #1558

MsgRollappFraudProposal previously allowed malformed stateless fields through governance and could underflow a zero fraud height during execution. This change validates rollapp IDs, nonzero fraud heights, and optional punish-sequencer addresses, while preserving revision 0 and surfacing the original validation detail from the handler.

- Rejects malformed proposal fields as invalid arguments before governance execution.
- Keeps optional fields and revision-zero proposals valid.
- Corrects the rewardee diagnostic and preserves handler validation context.

- Reuses `NewChainID` for rollapp ID validation.
- Rejects `FraudHeight == 0` before the handler subtracts one.
- Validates `PunishSequencerAddress` only when supplied.
- Adds table-driven message tests and a keeper regression test.

</details>

## Proof of execution

- **Build and Test (CI): pending on GitHub; equivalent gate green locally** — `go-acc -o /tmp/issue-1558-coverage.txt ./... -- -v --race` exited 0.
- **Build and Test (CI): pending on GitHub; build green locally** — `go build -v ./...` exited 0.
- **golangci-lint (CI): green on GitHub and locally** — golangci-lint v2.1.0 reported `0 issues`.
- **CodeQL (CI): green on GitHub** — both Analyze and CodeQL checks passed.
- **Fraud proposal regressions: green** — `TestMsgRollappFraudProposalValidateBasic`, `TestSubmitRollappFraud`, and `TestSubmitRollappFraudPreservesValidationError` pass and are rerun by Build and Test CI.

**Not verified:** manually dispatched cross-repository E2E and upgrade workflows; they are not PR-gated and this change has no external endpoint, dependency, protobuf, or UI behavior.

### DIFF
--- a/x/rollapp/keeper/fraud_proposal.go
+++ b/x/rollapp/keeper/fraud_proposal.go
@@ -24,7 +24,7 @@ func (k Keeper) SubmitRollappFraud(goCtx context.Context, msg *types.MsgRollappF
 	}
 
 	if err := msg.ValidateBasic(); err != nil {
-		err = errorsmod.Wrap(gerrc.ErrInvalidArgument, "invalid msg")
+		err = errorsmod.Wrap(err, "invalid msg")
 		ctx.Logger().Error("Fraud proposal", "error", err)
 		return nil, err
 	}
--- a/x/rollapp/keeper/fraud_proposal_test.go
+++ b/x/rollapp/keeper/fraud_proposal_test.go
@@ -112,3 +112,13 @@ func (s *RollappTestSuite) TestSubmitRollappFraud() {
 		})
 	}
 }
+
+func (s *RollappTestSuite) TestSubmitRollappFraudPreservesValidationError() {
+	msg := &types.MsgRollappFraudProposal{
+		Authority: s.App.AccountKeeper.GetModuleAddress(govtypes.ModuleName).String(),
+	}
+
+	_, err := s.k().SubmitRollappFraud(s.Ctx, msg)
+
+	s.Require().ErrorContains(err, "rollapp ID is invalid")
+}
--- a/x/rollapp/types/message_fraud_proposal.go
+++ b/x/rollapp/types/message_fraud_proposal.go
@@ -19,11 +19,28 @@ func (m *MsgRollappFraudProposal) ValidateBasic() error {
 			"authority is not a valid bech32 address: %s", m.Authority,
 		)
 	}
+	if _, err := NewChainID(m.RollappId); err != nil {
+		return errorsmod.Wrap(
+			errors.Join(gerrc.ErrInvalidArgument, err),
+			"rollapp ID is invalid",
+		)
+	}
+	if m.FraudHeight == 0 {
+		return errorsmod.Wrap(gerrc.ErrInvalidArgument, "fraud height must be greater than zero")
+	}
+	if m.PunishSequencerAddress != "" {
+		if _, err := sdk.AccAddressFromBech32(m.PunishSequencerAddress); err != nil {
+			return errorsmod.Wrapf(
+				errors.Join(gerrc.ErrInvalidArgument, err),
+				"punish sequencer address is not a valid bech32 address: %s", m.PunishSequencerAddress,
+			)
+		}
+	}
 	if m.Rewardee != "" {
 		if _, err := sdk.AccAddressFromBech32(m.Rewardee); err != nil {
 			return errorsmod.Wrapf(
 				errors.Join(gerrc.ErrInvalidArgument, err),
-				"rewardee is not a valid bech32 address: %s", m.Authority,
+				"rewardee is not a valid bech32 address: %s", m.Rewardee,
 			)
 		}
 	}
--- a/x/rollapp/types/message_fraud_proposal_test.go
+++ b/x/rollapp/types/message_fraud_proposal_test.go
@@ -1,0 +1,95 @@
+package types_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/dymensionxyz/dymension/v3/testutil/sample"
+	"github.com/dymensionxyz/dymension/v3/x/rollapp/types"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
+)
+
+func TestMsgRollappFraudProposalValidateBasic(t *testing.T) {
+	valid := types.MsgRollappFraudProposal{
+		Authority:              sample.AccAddress(),
+		RollappId:              "rollapp_1234-1",
+		FraudHeight:            1,
+		FraudRevision:          0,
+		PunishSequencerAddress: sample.AccAddress(),
+		Rewardee:               sample.AccAddress(),
+	}
+
+	tests := []struct {
+		name        string
+		modify      func(*types.MsgRollappFraudProposal)
+		wantErr     error
+		wantMessage string
+	}{
+		{
+			name: "valid message with revision zero",
+		},
+		{
+			name: "empty rollapp ID",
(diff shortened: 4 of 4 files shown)


omriai2 (approved):
# Security Review — PR #2208 (`fix(rollapp): validate fraud proposal fields`)

No HIGH or MEDIUM confidence security vulnerabilities were identified in this diff.

## Analysis summary

The change is strictly additive validation on a governance-gated message plus an error-message correction. Each modified surface was traced:

**`x/rollapp/types/message_fraud_proposal.go` — new stateless checks**
- `NewChainID(m.RollappId)` is the same validator `Rollapp.ValidateBasic` already applies at rollapp creation, so the new rejection rule cannot desynchronize from what is storable on chain. It is regex + `big.Int`/`ParseUint` parsing — deterministic, returns errors rather than panicking, and `MustNewChainID` is not used here. No consensus-determinism risk introduced.
- The `FraudHeight == 0` rejection *closes* an unsigned integer underflow rather than opening one: the handler computes `msg.FraudHeight-1` before calling `HardFork`, which wraps to `MaxUint64`, and `RevertPendingStates` then computes `lastValidHeight+1`, wrapping back to `0`. This is a hardening fix.
- `PunishSequencerAddress` bech32 validation matches the format `MsgPunishSequencer.ValidateBasic` and `Keeper.RealSequencer` already assume for sequencer addresses. Preserving `""` as legal matches the proto's documented "optional" semantics and the handler's `if msg.PunishSequencerAddress != ""` guard, so no previously-reachable punish path is newly blocked or newly opened.
- The `Rewardee` fix changes only the interpolated value in an error string (`m.Authority` → `m.Rewardee`). Both values originate from the proposal itself and are already public on chain; this is not a data-exposure change.

**`x/rollapp/keeper/fraud_proposal.go` — error wrap change**
- The authority check (`msg.Authority != k.authority`, returning `gerrc.ErrUnauthenticated`) is untouched and still executes *before* validation, so no authorization boundary moved.
- Replacing `errorsmod.Wrap(gerrc.ErrInvalidArgument, "invalid msg")` with `errorsmod.Wrap(err, "invalid msg")` preserves the `gerrc.ErrInvalidArgument` class for `errors.Is` traversal, because every `ValidateBasic` return path joins `gerrc.ErrInvalidArgument` into the error or wraps it directly. The newly surfaced detail contains only proposal-supplied field values — no secrets, credentials, or PII.

**Test files** are excluded per the hard exclusions.

No injection, deserialization, path-traversal, crypto, authentication, or authorization surface is touched by this diff.

PR #2209, opened 2026-08-31
fix(sponsorship): adopt gerrc error taxonomy

### LINKED ISSUE #1077: Fix error handling in x/sponsorship
Adjust error handling in x/sponsorship to use https://pkg.go.dev/github.com/dymensionxyz/gerr-cosmos@v1.0.0/gerrc

## AI triage
All research is done — here is the re-triage.

The issue is a one-line code-standards task from August 2024: migrate `x/sponsorship` error handling onto the org's `gerr-cosmos/gerrc` error taxonomy. It carries no groomed brief. The single comment — the human update — is maintainer mtsitrin scoping it as "part of #1053", the still-open umbrella issue "Code standards: remove bogus error types". The update reports no work done, and I verified none has happened: two years later the module still has zero `gerrc` usage while every sibling module adopted the pattern. The work is a contained, mechanical, well-anchored refactor with an existing test suite — dispatchable.

## Research summary

The issue asks that error handling in the `x/sponsorship` module be adjusted to use `gerr-cosmos/gerrc` (the org's standard library of gRPC-style category errors: `ErrInvalidArgument`, `ErrNotFound`, `ErrFailedPrecondition`, etc.).

Per `mtsitrin`'s update on the thread, this issue is part of the umbrella effort #1053 "Code standards: remove bogus error types" (still open). That update reparents the issue but reports nothing as implemented, and nothing has been: verified read-only in the current clone (HEAD `945d736`, Aug 2026), the `x/sponsorship` module contains **zero** references to `gerrc`. Confirmed current state:

- `x/sponsorship/types/errors.go` registers seven module-local sentinel errors via `errorsmod.Register(ModuleName, 1..7, ...)` — exactly the "bogus error types" the umbrella issue targets.
- The keeper (`votes.go`, `helpers.go`, `msg_server.go`, `query_server.go`, `invariants.go`, genesis/module code) returns bare `fmt.Errorf(...)` throughout, several with the `"failed to ..."` prefix style, with no category classification.
- `types/msgs.go` and `keeper/msg_server.go` use legacy `sdkerrors.ErrInvalidAddress` / `sdkerrors.ErrorInvalidSigner` from the Cosmos SDK.
- `github.com/dymensionxyz/gerr-cosmos v1.1.0` is already a direct dependency in `go.mod` — no dependency work needed.
- The target pattern is well established in sibling modules and confirmed in the clone: `x/sequencer/types/errors.go` defines sentinels as wraps of `gerrc` categories (e.g. `gerrc.ErrInvalidArgument.Wrap("pubkey")`, `errorsmod.Wrap(gerrc.ErrFailedPrecondition, ...)`); `x/streamer`'s msg server wraps `gerrc.ErrUnauthenticated` for authority checks; `x/rollapp`, `x/dymns`, and `x/iro` follow the same style.
- No open or merged PR references this issue; the only cross-reference is the original sponsorship implementation PR #1044 where the debt was created. The assignee (`keruch`, also the filer) has not acted since filing.

The entire remaining scope — the refactor itself — is unowned and is what this brief dispatches.

## Root cause hypothesis

Not a bug — deliberate tech debt. The sponsorship module was implemented (PR #1044) before/apart from the org-wide error-taxonomy standardization tracked in #1053, so it shipped with module-local registered error codes and uncategorized `fmt.Errorf` returns. The consequence is inconsistent error classification across modules: sponsorship errors don't map to the standard gRPC-style categories that the rest of the codebase (and any client interpreting ABCI error codes) relies on.

## Areas involved

- `x/sponsorship/types/errors.go` — redefine the seven sentinels as wraps of the appropriate `gerrc` category (the `x/sequencer/types/errors.go` style is the reference), removing the `errorsmod.Register` calls.
- `x/sponsorship/types/` (`msgs.go`, `types.go`, `genesis.go`, `params.go`) — validation errors; replace legacy `sdkerrors.*` references with `gerrc` equivalents (`ErrInvalidAddress` → `gerrc.ErrInvalidArgument`-based).
- `x/sponsorship/keeper/` (`msg_server.go`, `votes.go`, `helpers.go`, `query_server.go`, `invariants.go`, genesis/hooks files) — classify keeper-level errors with `gerrc` categories where the error is user-facing (e.g. voting power below minimum → `gerrc.ErrFailedPrecondition`, non-rollapp/non-perpetual gauge → `gerrc.ErrInvalidArgument`, authority check → `gerrc.ErrUnauthenticated` per the `x/streamer` precedent); plain wrapped propagation (`fmt.Errorf("...: %w", err)`) of already-classified inner errors may remain.
- `x/sponsorship/**/*_test.go` — the suite asserts error identity via `require.ErrorIs` against the current sentinels; assertions must be updated to match the new definitions and continue to pass.

Decisions settled for the implementer: keep the exported sentinel variable names where they carry meaning (matching how `x/sequencer` retained named sentinels over `gerrc` bases) so call sites and tests stay small diffs; the change to ABCI error codes returned to clients is an accepted, intended consequence of the #1053 standardization — do not build any compatibility shim. Follow the repo's existing style of `errorsmod.Wrap(gerrc.ErrX, ...)` / `gerrc.ErrX.Wrap(...)`; do not introduce new registered module error codes.

## Acceptance criteria

- `x/sponsorship` contains no `errorsmod.Register` calls and no references to legacy `sdkerrors` sentinels; every user-facing error is based on a `gerrc` category, following the sibling-module pattern.
- Sentinel errors in `types/errors.go` wrap the semantically correct `gerrc` category (invalid input → `ErrInvalidArgument`, missing entity → `ErrNotFound`, state-dependent rejection → `ErrFailedPrecondition`, authority failure → `ErrUnauthenticated`).
- All existing `x/sponsorship` tests pass (`go test ./x/sponsorship/...`), with `ErrorIs` assertions updated to the new definitions; the full build and linters (`errorlint` is enabled in CI) pass.
- No behavior change other than error identity/messages: no state-machine logic, storage, or event changes.

## Open questions

None.

omriai1:
Closes #1077

`x/sponsorship` used legacy Cosmos SDK errors and uncategorized user-facing failures. This change assigns semantic gRPC categories to the module's distinct registered sentinels and uses direct `gerrc` cause chains at query and keeper boundaries, leaving state-machine behavior unchanged.

- Keeps all seven exported sponsorship sentinels mutually distinguishable.
- Assigns `InvalidArgument` or `FailedPrecondition` gRPC status to each sentinel.
- Classifies address, authority, gauge, voting-power, and missing-query errors.
- Adds ABCI codespace/code and negative `errors.Is` regression coverage.

<details><summary>Error taxonomy</summary>

- Invalid messages, params, genesis data, votes, distributions, and gauge shapes use `InvalidArgument` semantics.
- Missing gauges, votes, endorsements, and claim inputs resolve to `gerrc.ErrNotFound` (`gerrc` code 4).
- Minimum voting power, minimum allocation, and absent endorsers use `FailedPrecondition` semantics.
- Governance authority mismatches use `gerrc.ErrUnauthenticated`.
- Gauge decode/corruption errors retain their original cause and are not mislabeled `NotFound`.

</details>

<details><summary>Behavior and compatibility</summary>

Exported sponsorship sentinel names and their distinct `errors.Is` identities remain intact, including the `ErrNoEndorsers` control-flow contract used by `x/incentives`. Their registered errors now carry explicit gRPC status codes. Direct query and keeper classifications retain a `gerrc` root in the `Cause()` chain so `ABCIInfo` and `GRPCStatus` resolve correctly; storage writes, events, and state transitions are untouched.

</details>

## Proof of execution

- Build and Test CI suite — initial full-gate `go build ./...` and `go test -race ./...`: passed locally, including `x/sponsorship/keeper` and `x/sponsorship/types`.
- golangci-lint CI suite — CI-pinned v2.1 initial full gate and fix-round `golangci-lint run ./x/sponsorship/...`: passed locally with `0 issues`.
- Sponsorship fix-round regression suite — `go test ./x/sponsorship/... -count=1`: passed all packages, including sentinel identity and ABCI codespace/code assertions.

**Not verified:** nothing; all changed behaviors are covered by the suites above.

### DIFF
--- a/x/sponsorship/keeper/msg_server.go
+++ b/x/sponsorship/keeper/msg_server.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 	"github.com/dymensionxyz/sdk-utils/utils/uevent"
 
 	"github.com/dymensionxyz/dymension/v3/x/sponsorship/types"
@@ -82,7 +82,7 @@ func (m MsgServer) UpdateParams(ctx context.Context, msg *types.MsgUpdateParams)
 	}
 
 	if msg.Authority != m.k.authority {
-		return nil, sdkerrors.ErrorInvalidSigner.Wrapf("Only the gov module can update params")
+		return nil, gerrc.ErrUnauthenticated.Wrap("only the gov module can update params")
 	}
 
 	oldParams, err := m.k.GetParams(ctx)
--- a/x/sponsorship/keeper/msg_server_test.go
+++ b/x/sponsorship/keeper/msg_server_test.go
@@ -1,10 +1,10 @@
 package keeper_test
 
 import (
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	"github.com/cosmos/gogoproto/proto"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 
 	"github.com/dymensionxyz/dymension/v3/app/apptesting"
 	"github.com/dymensionxyz/dymension/v3/x/sponsorship/types"
@@ -38,7 +38,7 @@ func (s *KeeperTestSuite) TestUpdateParams() {
 					MinVotingPower:      types.DefaultMinVotingPower,
 				},
 			},
-			error: sdkerrors.ErrorInvalidSigner,
+			error: gerrc.ErrUnauthenticated,
 		},
 		{
 			name: "invalid params",
--- a/x/sponsorship/keeper/query_server.go
+++ b/x/sponsorship/keeper/query_server.go
@@ -36,13 +36,13 @@ func (q QueryServer) Vote(goCtx context.Context, request *types.QueryVoteRequest
 
 	voter, err := sdk.AccAddressFromBech32(request.GetVoter())
 	if err != nil {
-		return nil, errorsmod.Wrap(errors.Join(gerrc.ErrInvalidArgument, err), "invalid voter address")
+		return nil, errorsmod.Wrapf(gerrc.ErrInvalidArgument, "invalid voter address: %s", err)
 	}
 
 	vote, err := q.k.GetVote(ctx, voter)
 	if err != nil {
 		if errors.Is(err, collections.ErrNotFound) {
-			return nil, errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "vote")
+			return nil, errorsmod.Wrapf(gerrc.ErrNotFound, "vote: %s", err)
 		}
 		return nil, err
 	}
@@ -63,12 +63,12 @@ func (q QueryServer) EstimateClaim(goCtx context.Context, query *types.QueryEsti
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	addr, err := sdk.AccAddressFromBech32(query.Address)
 	if err != nil {
-		return nil, errorsmod.Wrap(errors.Join(gerrc.ErrInvalidArgument, err), "invalid claimer address")
+		return nil, errorsmod.Wrapf(gerrc.ErrInvalidArgument, "invalid claimer address: %s", err)
 	}
 	reward, err := q.k.EstimateClaim(ctx, addr, query.RollappId)
 	if err != nil {
 		if errors.Is(err, collections.ErrNotFound) {
-			return nil, errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "claim estimate")
+			return nil, errorsmod.Wrapf(gerrc.ErrNotFound, "claim estimate: %s", err)
 		}
 		return nil, err
 	}
@@ -80,7 +80,7 @@ func (q QueryServer) Endorsement(goCtx context.Context, query *types.QueryEndors
 	endorsement, err := q.k.GetEndorsement(ctx, query.RollappId)
 	if err != nil {
 		if errors.Is(err, collections.ErrNotFound) {
-			return nil, errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "endorsement")
+			return nil, errorsmod.Wrapf(gerrc.ErrNotFound, "endorsement: %s", err)
 		}
 		return nil, err
 	}
--- a/x/sponsorship/keeper/query_server.go
+++ b/x/sponsorship/keeper/query_server.go
@@ -2,9 +2,12 @@ package keeper
 
 import (
 	"context"
-	"fmt"
+	"errors"
 
+	"cosmossdk.io/collections"
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 
 	"github.com/dymensionxyz/dymension/v3/x/sponsorship/types"
 )
@@ -33,11 +36,14 @@ func (q QueryServer) Vote(goCtx context.Context, request *types.QueryVoteRequest
 
 	voter, err := sdk.AccAddressFromBech32(request.GetVoter())
 	if err != nil {
-		return nil, fmt.Errorf("invalid voter address: %w", err)
+		return nil, errorsmod.Wrap(errors.Join(gerrc.ErrInvalidArgument, err), "invalid voter address")
 	}
 
 	vote, err := q.k.GetVote(ctx, voter)
 	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return nil, errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "vote")
+		}
 		return nil, err
 	}
 
@@ -57,10 +63,13 @@ func (q QueryServer) EstimateClaim(goCtx context.Context, query *types.QueryEsti
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	addr, err := sdk.AccAddressFromBech32(query.Address)
 	if err != nil {
-		return nil, err
+		return nil, errorsmod.Wrap(errors.Join(gerrc.ErrInvalidArgument, err), "invalid claimer address")
 	}
 	reward, err := q.k.EstimateClaim(ctx, addr, query.RollappId)
 	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return nil, errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "claim estimate")
+		}
 		return nil, err
 	}
 	return &types.QueryEstimateClaimResponse{Reward: reward}, nil
@@ -70,6 +79,9 @@ func (q QueryServer) Endorsement(goCtx context.Context, query *types.QueryEndors
 	ctx := sdk.UnwrapSDKContext(goCtx)
 	endorsement, err := q.k.GetEndorsement(ctx, query.RollappId)
 	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
+			return nil, errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "endorsement")
+		}
 		return nil, err
 	}
 	return &types.QueryEndorsementResponse{Endorsement: endorsement}, nil
--- a/x/sponsorship/keeper/query_server_test.go
+++ b/x/sponsorship/keeper/query_server_test.go
@@ -1,0 +1,52 @@
+package keeper_test
+
+import (
+	"cosmossdk.io/collections"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
+
+	"github.com/dymensionxyz/dymension/v3/app/apptesting"
+	"github.com/dymensionxyz/dymension/v3/x/sponsorship/keeper"
+	"github.com/dymensionxyz/dymension/v3/x/sponsorship/types"
+)
+
+func (s *KeeperTestSuite) TestQueryMissingEntitiesUseNotFound() {
+	queryServer := keeper.NewQueryServer(s.App.SponsorshipKeeper)
+	ctx := sdk.WrapSDKContext(s.Ctx)
+	voter := apptesting.CreateRandomAccounts(1)[0]
+
+	tests := []struct {
+		name  string
+		query func() error
+	}{
+		{
+			name: "vote",
+			query: func() error {
+				_, err := queryServer.Vote(ctx, &types.QueryVoteRequest{Voter: voter.String()})
+				return err
+			},
+		},
+		{
+			name: "claim estimate",
+			query: func() error {
+				_, err := queryServer.EstimateClaim(ctx, &types.QueryEstimateClaim{Address: voter.String(), RollappId: "missing"})
+				return err
+			},
+		},
+		{
+			name: "endorsement",
+			query: func() error {
+				_, err := queryServer.Endorsement(ctx, &types.QueryEndorsement{RollappId: "missing"})
+				return err
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		s.Run(tt.name, func() {
+			err := tt.query()
+			s.Require().ErrorIs(err, gerrc.ErrNotFound)
+			s.Require().ErrorIs(err, collections.ErrNotFound)
+		})
+	}
+}
--- a/x/sponsorship/keeper/query_server_test.go
+++ b/x/sponsorship/keeper/query_server_test.go
@@ -1,7 +1,7 @@
 package keeper_test
 
 import (
-	"cosmossdk.io/collections"
+	errorsmod "cosmossdk.io/errors"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 
@@ -46,7 +46,43 @@ func (s *KeeperTestSuite) TestQueryMissingEntitiesUseNotFound() {
 		s.Run(tt.name, func() {
 			err := tt.query()
 			s.Require().ErrorIs(err, gerrc.ErrNotFound)
-			s.Require().ErrorIs(err, collections.ErrNotFound)
+			codespace, code, _ := errorsmod.ABCIInfo(err, false)
+			s.Require().Equal(gerrc.DefaultCodespace, codespace)
+			s.Require().Equal(uint32(4), code)
+		})
+	}
+}
+
+func (s *KeeperTestSuite) TestQueryInvalidAddressesUseInvalidArgument() {
+	queryServer := keeper.NewQueryServer(s.App.SponsorshipKeeper)
+	ctx := sdk.WrapSDKContext(s.Ctx)
+
+	tests := []struct {
+		name  string
+		query func() error
+	}{
+		{
+			name: "vote",
+			query: func() error {
+				_, err := queryServer.Vote(ctx, &types.QueryVoteRequest{Voter: "invalid"})
+				return err
+			},
+		},
+		{
+			name: "claim estimate",
+			query: func() error {
+				_, err := queryServer.EstimateClaim(ctx, &types.QueryEstimateClaim{Address: "invalid"})
+				return err
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		s.Run(tt.name, func() {
+			err := tt.query()
+			codespace, code, _ := errorsmod.ABCIInfo(err, false)
+			s.Require().Equal(gerrc.DefaultCodespace, codespace)
+			s.Require().Equal(uint32(2), code)
 		})
 	}
 }
--- a/x/sponsorship/keeper/votes.go
+++ b/x/sponsorship/keeper/votes.go
@@ -1,11 +1,14 @@
 package keeper
 
 import (
+	"errors"
 	"fmt"
 
+	errorsmod "cosmossdk.io/errors"
 	"cosmossdk.io/math"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 	"github.com/dymensionxyz/sdk-utils/utils/uevent"
 
 	incentivestypes "github.com/dymensionxyz/dymension/v3/x/incentives/types"
@@ -33,7 +36,7 @@ func (k Keeper) Vote(ctx sdk.Context, voter sdk.AccAddress, weights []types.Gaug
 
 	// Validate that the user has min voting power
 	if vpBreakdown.TotalPower.LT(params.MinVotingPower) {
-		return types.Vote{}, types.Distribution{}, fmt.Errorf("voting power '%s' is less than min voting power expected '%s'", vpBreakdown.TotalPower, params.MinVotingPower)
+		return types.Vote{}, types.Distribution{}, gerrc.ErrFailedPrecondition.Wrapf("voting power '%s' is less than min voting power expected '%s'", vpBreakdown.TotalPower, params.MinVotingPower)
 	}
 
 	// Apply the vote weights to the power -> get a distribution update in absolute values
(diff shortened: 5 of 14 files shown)


omriai2 (changes requested):
Review complete. Two blocking findings, both confirmed with runtime evidence.

---

**VERDICT: REQUEST_CHANGES** — the migration is mechanically clean and the tests it adds do pass, but it silently destroys the distinguishability of the module's exported error sentinels (one of which is a live control-flow gate on reward distribution in `x/incentives`), and the query-path classification it introduces does not actually produce a `gerrc` ABCI/gRPC identity, so the PR's headline claim is unmet.

### Index

1. **Exported sentinels collapse into a single identity — `errors.Is` no longer distinguishes them, and `x/incentives` gates epoch reward distribution on one of them.** Accept when `ErrNoEndorsers` (and, ideally, the six `invalid-*` sentinels) are once again mutually exclusive under `errors.Is`, with a committed negative-direction regression test asserting an unrelated same-category error does *not* match the sentinel.
2. **The `errors.Join(gerrc.X, err)` classification sites do not yield a `gerrc` ABCI/gRPC identity — the taxonomy adoption is a no-op on exactly the paths the PR adds it to.** Accept when `errorsmod.ABCIInfo` on the classified query/keeper errors resolves to codespace `gerrc` with the intended code, proven by a committed assertion on the resolved codespace/code (not only `errors.Is`).

Non-blocking notes are folded at the end.

---

<details><summary>Finding 1 — Exported sentinels collapse into a single identity; a live reward-distribution gate depends on the property that was removed</summary>

**What the change does.** All seven registered sentinels moved from `errorsmod.Register(ModuleName, N, ...)` — which mints a distinct root `*errorsmod.Error` — to `errorsmod.Wrap(gerrc.ErrX, ...)`, which produces a `*wrappedError` whose cause chain terminates at the shared `gerrc` root.

**Why that breaks identity.** `(*errorsmod.Error).Is(target)` does not check whether the receiver appears in the *caller's* chain; it unwraps the **target** looking for itself. So when Go's `errors.Is` walks any chain and reaches `gerrc.ErrFailedPrecondition`, it calls `gerrc.ErrFailedPrecondition.Is(ErrNoEndorsers)` — and because `ErrNoEndorsers` now wraps `gerrc.ErrFailedPrecondition`, that returns true. The same inversion makes all six `ErrInvalidArgument`-based sentinels mutually equal.

**Measured, on this branch:**

```
BEFORE (registered sentinel): errors.Is(unrelatedFailedPrecondition, ErrNoEndorsers) = false
AFTER  (wrapped   sentinel): errors.Is(unrelatedFailedPrecondition, ErrNoEndorsers) = true
cross-sentinel errors.Is matches among the 6 invalid-* sentinels: 30/30
errors.Is(bech32AddressError, types.ErrInvalidVote)   = true
errors.Is(bech32AddressError, types.ErrInvalidGenesis)= true
```

(`unrelatedFailedPrecondition` is `gerrc.ErrFailedPrecondition.Wrap("gauge weight is less than min allocation weight")` — the exact shape this PR introduces at [x/sponsorship/keeper/votes.go:182](x/sponsorship/keeper/votes.go#L182). `bech32AddressError` is the shape introduced at [x/sponsorship/types/msgs.go:19](x/sponsorship/types/msgs.go#L19).) `30/30` means every ordered pair of the six `invalid-*` sentinels now matches — the module went from seven distinguishable error identities to two equivalence classes.

**Why this is blocking rather than cosmetic.** `x/incentives` uses one of these sentinels as a control-flow gate on the epoch-end reward path:

[x/incentives/keeper/gauge_endorsement.go:61](x/incentives/keeper/gauge_endorsement.go#L61)
```go
err := k.spk.UpdateEndorsementTotalCoins(ctx, endorsementGauge.Endorsement.RollappId, epochRewards)
if errors.Is(err, sponsorshiptypes.ErrNoEndorsers) {
    // Don't fill this epoch, save rewards for the future
    return nil
}
```

A match here swallows the error and silently skips filling the epoch. Today the specific call chain behind `UpdateEndorsementTotalCoins` happens not to emit any other `FailedPrecondition`-derived error — its other two failure modes are collections get/save errors — so I am not claiming a currently-firing miscount. What I am claiming is that the guard's safety used to be guaranteed by the sentinel being a distinct registered root, and after this change it is guaranteed only by the incidental fact that nothing downstream currently returns `FailedPrecondition`. Any future `gerrc.ErrFailedPrecondition` introduced anywhere under that call — including in `GetEndorsement` or `SaveEndorsement` — converts a hard error into a silent skipped reward epoch, on a money path, inside an epoch-end hook where nothing surfaces the swallow.

**Class.** This is not one sentinel. It is the whole of [x/sponsorship/types/errors.go](x/sponsorship/types/errors.go) — all seven. The committed test [x/sponsorship/types/errors_test.go](x/sponsorship/types/errors_test.go) asserts only the direction that passes regardless (`ErrorIs(sentinel, category)`); it has no power to detect the collapse, so green CI here is not evidence.

**Acceptance criterion.** `ErrNoEndorsers` must again be non-matchable from an unrelated `gerrc.ErrFailedPrecondition` error, and the six `invalid-*` sentinels must again be mutually exclusive. Two shapes I would accept:

- Keep the sentinels as distinct registered roots and attach the gRPC semantics at registration — `errorsmod.RegisterWithGRPCCode(ModuleName, N, codes.FailedPrecondition, "no endorsers")` — which gives you a real ABCI codespace/code *and* the gRPC classification the issue asks for, without the identity collapse. This also resolves Finding 2 for the sentinel paths for free.
- Or, if the module genuinely wants `gerrc` identity on the sentinels, drop the shared sentinels for the cases anything switches on and have `x/incentives` gate on something unambiguous instead of `errors.Is` against a category-derived error.

Either way, commit a negative-direction test: `require.False(t, errors.Is(gerrc.ErrFailedPrecondition.Wrap("anything"), types.ErrNoEndorsers))`, plus at least one cross-sentinel non-match for the `invalid-*` group.

</details>

<details><summary>Finding 2 — The errors.Join classification sites do not produce a gerrc ABCI/gRPC identity, so the taxonomy adoption is inert exactly where the PR adds it</summary>

**What the PR body claims.** "The intended ABCI/gRPC error identities change to `gerrc`."

**What actually happens.** `errorsmod`'s `ABCIInfo` resolves codespace and code by walking the `Cause() error` chain only. `errors.Join` returns a `*joinError`, which implements `Unwrap() []error` and has no `Cause()`. So the walk dead-ends at the join node and falls through to the internal default. `(*wrappedError).GRPCStatus()` walks the same `causer` chain and hits the same dead end, returning `codes.Unknown` with only the local wrap message.

**Measured, on this branch:**

```
C1 join(NotFound) ABCI  codespace="undefined" code=1 log="vote: not found\ncollections: not found"
C2 gerrc.NotFound ABCI  codespace="gerrc" code=4 log="vote: not found"
C3 ErrInvalidVote ABCI  codespace="gerrc" code=2
C4 old registered ABCI  codespace="probe_codespace" code=5

D1 join(NotFound) gRPC  code=Unknown  msg="vote"
D2 gerrc.NotFound gRPC  code=NotFound msg="codespace gerrc code 4: not found: vote"
```

`C1` is the exact expression this PR introduces — `errorsmod.Wrap(errors.Join(gerrc.ErrNotFound, err), "vote")`. It resolves to codespace `undefined`, code `1`, i.e. *unclassified internal error*. `C2` is the same intent expressed without the join and resolves correctly to `gerrc`/`4`.

**Client-visible consequence.** Cosmos SDK v0.50's ABCI query handler routes through `gRPCErrorToSDKError`, which switches on the gRPC status code ([baseapp/abci.go:1152](https://github.com/cosmos/cosmos-sdk/blob/v0.50.14/baseapp/abci.go#L1152) in the pinned SDK). With `codes.NotFound` it maps to `ErrKeyNotFound`; with `codes.Unknown` it falls to `default` → `ErrUnknownRequest`. Per `D1`, these queries produce `Unknown`, so a missing vote/endorsement/claim surfaces to clients as "unknown request", not "not found". The pre-PR behavior returned the bare collections error, which mapped to `ErrInvalidRequest`. So the observable classification changed from one wrong answer to a different wrong answer, while the intended answer was never reached.

**Class.** Six sites, all added or rewritten by this PR: the five in [x/sponsorship/keeper/query_server.go](x/sponsorship/keeper/query_server.go) (voter address, missing vote, claimer address, missing claim estimate, missing endorsement) and the gauge lookup in [x/sponsorship/keeper/votes.go:187](x/sponsorship/keeper/votes.go#L187). I am aware `errors.Join(gerrc.X, err)` appears elsewhere in this repo; in those places the classification loss is pre-existing and not this PR's problem. Here it is, because classification is the entire deliverable.

The committed test [x/sponsorship/keeper/query_server_test.go](x/sponsorship/keeper/query_server_test.go) asserts only `ErrorIs(err, gerrc.ErrNotFound)` and `ErrorIs(err, collections.ErrNotFound)`. Both pass — Go's `errors.Is` *does* traverse `Unwrap() []error`, so `errors.Is` works even though `ABCIInfo` and `GRPCStatus` do not. The test therefore cannot distinguish a working taxonomy from an inert one, which is why CI is green on a feature that does not function.

**Acceptance criterion.** Keep the `gerrc` error as the root of the `Cause()` chain rather than burying it behind `errors.Join` — e.g. `errorsmod.Wrapf(gerrc.ErrNotFound, "vote: %s", err)`, or `gerrc.ErrNotFound.Wrapf(...)`. If preserving `errors.Is` reachability to the underlying `collections.ErrNotFound` matters to you, say so and pick a shape that keeps both; but the `gerrc` root must be reachable via `Cause()`. Then extend the regression test to assert the resolved identity, not just membership:

```go
codespace, code, _ := errorsmod.ABCIInfo(err, false)
s.Require().Equal(gerrc.DefaultCodespace, codespace)
s.Require().Equal(uint32(4), code) // gerrc.ErrNotFound
```

If the team decides `errors.Is` identity is the only contract in scope and ABCI/gRPC codes are explicitly out of scope, that is a legitimate call — but then the PR body's "The intended ABCI/gRPC error identities change to `gerrc`" claim is false and must be corrected, and the scope decision recorded. I would not prefer that outcome, since issue #1077 is about the user-facing taxonomy and the query paths are the user-facing surface.

</details>

<details><summary>Non-blocking notes</summary>

**A. All `GetGaugeByID` failures are now labeled `NotFound`, including store corruption.** [x/sponsorship/keeper/votes.go:187](x/sponsorship/keeper/votes.go#L187) unconditionally joins `gerrc.ErrNotFound` onto whatever `GetGaugeByID` returns. That function has two failure modes — a genuine missing-key case, and `proto.Unmarshal` failing on a stored gauge ([x/incentives/keeper/gauge.go:117](x/incentives/keeper/gauge.go#L117)). Labeling a decode failure as "not found" tells an operator the gauge is absent when the real condition is corrupted state. Consider gating on the missing-key case the way [x/sponsorship/keeper/query_server.go](x/sponsorship/keeper/query_server.go) now gates on `collections.ErrNotFound`, or leaving unrecognized failures uncategorized.

**B. Error text convention.** The rewritten line at [x/sponsorship/keeper/votes.go:187](x/sponsorship/keeper/votes.go#L187) retains `"failed to get gauge by id: %d"`. Project convention is not to write "failed to…" in Go error strings. The text is pre-existing, but the line was rewritten anyway, so it is a free fix.

**C. Proof of execution — shape is fine, not spot-checked.** The section carries three receipted claims (`go build ./...` + `go test -race ./...`, CI-pinned `golangci-lint run ./...`, `go test ./x/sponsorship/... -count=1`) and an explicit "Not verified: nothing". All three are CI-gated suites, so per policy the named suite plus CI is the proof and I did not open transcripts. Note that the `build` check was still pending at review time; `golangci-lint`, `Analyze`, and `CodeQL` are green. No external dependency call is added or re-pointed by this diff, so no live-invocation evidence is required. I did not run the repository suites myself; the runtime evidence above comes from two throwaway probe files that were removed, leaving the tree clean.

**D. Not a finding, recorded for completeness.** Dropping `errorsmod.Register` changes the ABCI codespace/code on failed sponsorship transactions, which alters `LastResultsHash` for those blocks. That is normal for an error-taxonomy migration deployed at an upgrade boundary and consistent with how this repo has handled `gerrc` migrations elsewhere; I am not asking for anything here.

</details>

VERDICT: REQUEST_CHANGES

omriai2 (approved):
**VERDICT: APPROVE** — both prior blocking findings are resolved by the shapes I named, verified by running the committed regression tests; what remains is one brittleness nit and two observations I would not hold the merge for.

### Index

1. **[RESOLVED — prior blocker]** Sentinel identity collapse. Fixed via `errorsmod.RegisterWithGRPCCode`, restoring seven distinct registered roots while attaching the gRPC semantics; verified 30/30 cross-sentinel non-match and the `ErrNoEndorsers` negative case now pass.
2. **[RESOLVED — prior blocker]** Inert `gerrc` classification at the query/keeper boundaries. Fixed by dropping `errors.Join` for direct `errorsmod.Wrapf(gerrc.X, ...)` and by converting the one `fmt.Errorf` re-wrap that would have severed the `Cause()` chain; committed assertions now pin resolved codespace/code.
3. **[RESOLVED — prior note A]** `GetGaugeButID` decode failures mislabeled `NotFound`. Fixed by `classifyGaugeLookupError`, with an internal unit test pinning both branches.
4. **[RESOLVED — prior note B]** "failed to…" error text. Rewritten to `gauge %d: …`.
5. **[NIT, non-blocking]** `classifyGaugeLookupError` distinguishes the two cases by exact string equality against another module's error text.
6. **[OBSERVATION, non-blocking]** Query errors no longer carry `collections.ErrNotFound` in the chain; the PR body no longer claims they do, so this is consistent, just worth knowing.

---

<details><summary>Finding 1 — Sentinel identity collapse: RESOLVED</summary>

The author took the first of the two shapes I named. [x/sponsorship/types/errors.go](x/sponsorship/types/errors.go) now registers all seven sentinels as distinct roots with explicit gRPC status:

```go
ErrInvalidGaugeWeight  = errorsmod.RegisterWithGRPCCode(ModuleName, 1, codes.InvalidArgument, "invalid gauge weight")
...
ErrNoEndorsers         = errorsmod.RegisterWithGRPCCode(ModuleName, 7, codes.FailedPrecondition, "no endorsers")
```

This is structurally correct, not incidentally correct: `RegisterWithGRPCCode` mints a root `*errorsmod.Error` with no `Cause()`, so the `(*Error).Is` inversion that made every sentinel match every other one has no chain to walk. The `errors.Is` guard in `x/incentives` at [x/incentives/keeper/gauge_endorsement.go:61](x/incentives/keeper/gauge_endorsement.go#L61) — the reward-distribution gate that motivated the block — is safe again, and safe by construction rather than by the accident of nothing downstream emitting `FailedPrecondition`.

The negative-direction coverage I asked for is committed in [x/sponsorship/types/errors_test.go](x/sponsorship/types/errors_test.go): a 6×5 cross-sentinel `require.False(errors.Is(...))` sweep, plus `require.False(t, errors.Is(gerrc.ErrFailedPrecondition.Wrap("anything"), types.ErrNoEndorsers))`. Ran it:

```
=== RUN   TestErrorsHaveDistinctIdentitiesAndGRPCCategories
--- PASS: TestErrorsHaveDistinctIdentitiesAndGRPCCategories (0.00s)
```

That directly reverses the `30/30` collapse and the `AFTER = true` false positive I measured last round. The three `msgs_test.go` suites also pass, confirming that `ErrInvalidDistribution` and `ErrInvalidParams` are once again matchable as themselves and not as any bech32 address error:

```
--- PASS: TestMsgVote (0.00s)
    --- PASS: TestMsgVote/Invalid_distribution,_Weight_>_100_*_10^18 (0.00s)
--- PASS: TestMsgUpdateParams (0.00s)
    --- PASS: TestMsgUpdateParams/Invalid_params,_MinAllocationWeight_<_0 (0.00s)
```

A bonus the author may not have noticed, and which is worth recording: because `Register` and `RegisterWithGRPCCode` both produce codespace `ModuleName` with the same numeric code, the ABCI codespace/code on failed sponsorship transactions is now byte-identical to `main` for all seven sentinels. Only the gRPC status changed. That eliminates the tx-result-code churn I noted last round as an accepted-but-present cost; there is now no such cost.

</details>

<details><summary>Finding 2 — Inert gerrc classification at query and keeper boundaries: RESOLVED</summary>

All six `errors.Join(gerrc.X, err)` sites are gone. [x/sponsorship/keeper/query_server.go](x/sponsorship/keeper/query_server.go) now uses `errorsmod.Wrapf(gerrc.ErrNotFound, "vote: %s", err)` and siblings, which keeps the registered `gerrc` root as the terminus of the `Cause()` chain — exactly the property `errorsmod.ABCIInfo` and `(*wrappedError).GRPCStatus` walk, and exactly what the join node was severing.

The author also caught a second-order break I had not called out and which would have silently defeated the fix on the transaction path: `Keeper.Vote` previously re-wrapped `validateWeights` output with `fmt.Errorf("error validating weights: %w", err)`. `fmt`'s wrapper implements `Unwrap` but not `Cause`, so `ABCIInfo` would have dead-ended there even after the query-side fix. That is now `errorsmod.Wrap(err, "validate weights")` at [x/sponsorship/keeper/votes.go:27](x/sponsorship/keeper/votes.go#L27), which preserves the chain. I traced the rest of the path and it holds: `MsgServer.Vote`, `RevokeVote`, and `ClaimRewards` all return keeper errors unwrapped, so the classification survives intact to the tx result.

The acceptance criterion — a committed assertion on the *resolved* codespace and code rather than only `errors.Is` — is met in three places: `TestQueryMissingEntitiesUseNotFound` and the new `TestQueryInvalidAddressesUseInvalidArgument` assert `gerrc.DefaultCodespace` with codes 4 and 2, and `TestMsgVote` now compares `ABCIInfo(err)` against `ABCIInfo(tc.errorIs)` for every classified case, which covers the `FailedPrecondition` min-voting-power and min-allocation paths end-to-end through the msg server. I read those four paths and each resolves to a `gerrc` root; the keeper package compiles clean, which I confirmed by running a test in it. The suite itself is CI-gated, so I let the `build` check carry execution rather than rerunning it — it was still in progress at review time along with `golangci-lint` and `Analyze`, with `label-prs` green.

</details>

<details><summary>Findings 3 and 4 — Prior non-blocking notes: both RESOLVED</summary>

**Note A (gauge decode failures mislabeled `NotFound`).** `validateWeights` no longer blankets every `GetGaugeByID` failure with `NotFound`. The new `classifyGaugeLookupError` at [x/sponsorship/keeper/votes.go:204](x/sponsorship/keeper/votes.go#L204) routes the genuine missing-key case to `gerrc.ErrNotFound` and leaves a `proto.Unmarshal` failure uncategorized with its original cause intact — so corrupted state now reads as an internal error rather than telling an operator the gauge is absent. The committed internal test pins both branches, including that the corrupt case preserves `errors.Is(err, corrupt)` and resolves to `UndefinedCodespace`/1:

```
=== RUN   TestClassifyGaugeLookupError
--- PASS: TestClassifyGaugeLookupError (0.00s)
```

**Note B (error text convention).** `"failed to get gauge by id: %d"` is now `"gauge %d: %s"`, matching the project's no-"failed to" convention. The `TestMsgVote` expectation was updated in step.

</details>

<details><summary>Finding 5 — NIT: gauge lookup classification keys off another module's exact error string</summary>

```go
func classifyGaugeLookupError(gaugeID uint64, err error) error {
	if err.Error() == fmt.Sprintf("gauge with ID %d does not exist", gaugeID) {
		return errorsmod.Wrapf(gerrc.ErrNotFound, "gauge %d: %s", gaugeID, err)
	}
	return fmt.Errorf("get gauge %d: %w", gaugeID, err)
}
```

This reproduces, character for character, the message `x/incentives` builds at [x/incentives/keeper/gauge.go:114](x/incentives/keeper/gauge.go#L114). It is deterministic, so there is no consensus concern, and the failure mode is benign — if that message is ever reworded, sponsorship quietly stops classifying missing gauges as `NotFound` and falls through to the generic branch. The error is still returned and the transaction still fails correctly; only the category is lost, and it would be lost silently, since the internal test pins the string on this side rather than asserting agreement with the producer.

I am not holding the merge for it. The durable shape, if you want one later, is for `GetGaugeByID` to return a `gerrc.ErrNotFound`-rooted error for the missing-key case so consumers can use `errors.Is`, or for sponsorship to ask `x/incentives` whether the gauge exists before fetching it. Either removes the textual coupling. A cheaper stopgap is a comment on `classifyGaugeLookupError` naming the producer, so the next person to reword that message has a chance of finding this.

</details>

<details><summary>Finding 6 — OBSERVATION: query errors no longer carry the collections cause</summary>

The query server now interpolates the underlying error with `%s` rather than wrapping it with `%w`, so `errors.Is(err, collections.ErrNotFound)` on a query result is false where the first revision made it true. The original text still appears in the message, and the earlier PR body's "preserves underlying causes" claim has been dropped in favor of the accurate "retain a `gerrc` root in the `Cause()` chain", so the description matches the code. The test assertion that checked for `collections.ErrNotFound` was correctly removed rather than left to rot.

Nothing consumes that identity — query servers are leaf boundaries — so this is a deliberate, honest trade, recorded here only so it is not mistaken for an oversight later. If you ever do need both, `errorsmod.Wrapf` over an inner `fmt.Errorf("%w", err)` keeps the `Cause()` chain rooted at `gerrc` while leaving the collections error reachable through `Unwrap`.

</details>

<details><summary>Proof of execution and docs</summary>

The section carries three receipted claims naming their commands (`go build ./...` + `go test -race ./...`, CI-pinned v2.1 `golangci-lint run ./x/sponsorship/...`, `go test ./x/sponsorship/... -count=1`) and an explicit `Not verified: nothing`. Shape is sound. All three are CI-gated, so per policy the named suite plus the required checks are the proof and I did not open transcripts; the `build`, `golangci-lint`, and `Analyze` checks were still in progress on `1ad63f6d` at review time and remain the merge gate. No external dependency call is added or re-pointed, so no live-invocation evidence is required.

The PR body was rewritten for this round and now matches the implementation — the prior "The intended ABCI/gRPC error identities change to `gerrc`" claim, which was false for the join sites, is replaced with an accurate description of registered sentinels carrying gRPC status plus direct `gerrc` roots at the boundaries, and it correctly calls out the `ErrNoEndorsers` contract with `x/incentives` and the gauge-decode carve-out. `x/sponsorship` ships no README, and the one markdown file in the module makes no reference to errors or codespaces, so nothing was made stale. My two throwaway probe files from the previous round were removed; the working tree is clean.

</details>

VERDICT: APPROVE

PR #2210, opened 2026-08-31
fix(rollapp): handle missing revision heights

### LINKED ISSUE #1596: Tech debt: make rollapp.GetRevisionForHeight more robust to missing heights
Should return also and error or OK bool. The current method is bug prone

## AI triage
Research is complete. The issue is a small, well-scoped robustness refactor; the only comment is a maintainer asking for the bug to be spelled out, which code analysis answers. Here is the re-triage.

## Research summary

Issue #1596 (filed by `danwt`, Dec 2024) asks that `Rollapp.GetRevisionForHeight` stop silently returning a zero-value `Revision{}` when no revision covers the requested height, and instead signal not-found via an ok-bool or error.

Confirmed in the clone (head `945d736`, 2026-08-15): the function is unchanged at `x/rollapp/types/rollapp.go:144` and carries a `TODO` comment that links to this very issue, so the team already accepted it as tracked debt. No open or merged PR references the function or the issue, and the issue has no linked closing PRs — nothing has been done.

The single comment is `mtsitrin`'s update (2024-12-03) asking `danwt` to elaborate on what the possible bug is, quoting the function. It reports no work done and was never answered. What is left is the entire fix, and the elaboration `mtsitrin` asked for — which the dispatched worker owns delivering in the brief/PR description, since code analysis fully answers it:

- The not-found sentinel `Revision{}` equals `{Number: 0, StartHeight: 0}`, which is byte-identical to a legitimate genesis revision. Callers cannot distinguish "no revision covers this height" from "revision 0 covers it."
- `NewRollapp` always seeds `Revisions` with `{0, 0}`, so on the normal creation path the sentinel is unreachable — but `ValidateBasic` performs zero validation on `Revisions` (confirmed by reading it in full), so genesis import or direct construction can yield an empty or gapped revision list, and `LatestRevision`'s own comment documents empty lists occurring in tests.
- Concrete hazard site: `SubmitRollappFraud` (`x/rollapp/keeper/fraud_proposal.go:40`) gates a hard fork on `GetRevisionForHeight(h).Number != msg.FraudRevision`. Against a rollapp with an empty/malformed revision list, a proposal with `FraudRevision == 0` spuriously passes this duplicate-proposal guard. The path is gov-authority-gated, which bounds the practical risk but doesn't remove the ambiguity.

## Root cause hypothesis

Not a live production bug but confirmed latent fragility: the method conflates its not-found case with a valid value because `Revision`'s zero value is a meaningful revision. Any future caller (or any state reachable through genesis import, which is unvalidated for revisions) that hits the fallthrough silently proceeds with revision 0 instead of failing. The fix `danwt` requested is to make not-found explicit in the signature.

## Areas involved

- `x/rollapp/types/rollapp.go` — change `GetRevisionForHeight` to return `(Revision, bool)` (ok-bool, matching the codebase's `GetRollapp`-style convention; an error return was considered and rejected as heavier than needed for a pure lookup). Update `IsRevisionStartHeight` to return `false` when not found.
- `x/rollapp/keeper/fraud_proposal.go` — on not-found, reject the proposal with a `gerrc.ErrFailedPrecondition`-wrapped error (same family as the existing mismatch branch), preserving the existing log-and-return pattern of that handler.
- `x/rollapp/keeper/fraud_proposal_test.go` — update the six existing assertions to the new signature and add a case covering the not-found path (height below all revision start-heights / empty revision list).
- Remove the `TODO` comment referencing #1596 once resolved.
- Out of scope: `LatestRevision` has the same zero-value pattern but is explicitly documented and is not what this issue requests — leave it (noting it in the PR is fine).

## Acceptance criteria

- `GetRevisionForHeight` returns an explicit not-found indicator; the zero-value `Revision{}` is never returned as if it were a valid lookup result.
- For every rollapp with the standard genesis revision `{0, 0}`, all existing behavior is unchanged (lookups for any height succeed exactly as before).
- `SubmitRollappFraud` rejects a fraud proposal whose height matches no revision, rather than comparing against a fabricated revision 0; covered by a test.
- `IsRevisionStartHeight` returns `false` for heights matching no revision; existing `IsFirstHeightOfLatestFork` behavior for valid forks unchanged.
- All existing tests updated and passing; build, linters, and `go vet` clean.

## Open questions

None.

omriai1:
Closes #1596

`GetRevisionForHeight` previously returned a zero-value revision when no revision covered a height, making missing state indistinguishable from a valid genesis revision. This change returns an explicit found boolean, rejects missing fraud-height revisions before mutation, and keeps valid lookup behavior unchanged.

- Makes missing revision heights explicit.
- Rejects malformed-state fraud proposals with `ErrFailedPrecondition`.
- Covers valid, below-first, and empty revision lists.

- Changes `Rollapp.GetRevisionForHeight` to return `(Revision, bool)`.
- Makes `IsRevisionStartHeight` return false when no revision covers the height.
- Performs one revision lookup in `SubmitRollappFraud` and exits before `HardFork` when it is missing.
- Removes the resolved issue TODO.

</details>

## Proof of execution

- Targeted rollapp regression suites: PASS — `go test ./x/rollapp/types ./x/rollapp/keeper -run "TestRollappGetRevisionForHeight|TestRollappKeeperTestSuite/TestSubmitRollappFraud" -count=1 -v`.
- Build and Test / race coverage equivalent: PASS — `go-acc -o /tmp/issue1596-coverage.txt ./... -- -race`; changed packages `x/rollapp/keeper` and `x/rollapp/types` passed.
- Full repository tests: PASS — `go test ./... -count=1`.
- Build and vet: PASS — `go build ./...` and `go vet ./...`.
- GitHub Build and Test (`build`): PASS.
- GitHub CodeQL (`Analyze`, `CodeQL`): PASS.
- GitHub `golangci-lint`: PASS on rerun; the first attempt timed out fetching the upstream lint schema before analysis began.
- GitHub codecov (`codecov/patch`, `codecov/project`): PASS.
- Independent Go review: PASS — no actionable findings.

**Not verified:** nothing; all changed behaviors are covered by automated tests.

### DIFF
--- a/x/rollapp/keeper/fraud_proposal.go
+++ b/x/rollapp/keeper/fraud_proposal.go
@@ -37,8 +37,14 @@ func (k Keeper) SubmitRollappFraud(goCtx context.Context, msg *types.MsgRollappF
 	}
 
 	// check correct revision number (to avoid sending duplicated proposals)
-	if rollapp.GetRevisionForHeight(msg.FraudHeight).Number != msg.FraudRevision {
-		err := errorsmod.Wrapf(gerrc.ErrFailedPrecondition, "fraud revision number mismatch: %d != %d", rollapp.GetRevisionForHeight(msg.FraudHeight).Number, msg.FraudRevision)
+	revision, found := rollapp.GetRevisionForHeight(msg.FraudHeight)
+	if !found {
+		err := errorsmod.Wrapf(gerrc.ErrFailedPrecondition, "revision not found for fraud height: %d", msg.FraudHeight)
+		ctx.Logger().Error("Fraud proposal", "error", err)
+		return nil, err
+	}
+	if revision.Number != msg.FraudRevision {
+		err := errorsmod.Wrapf(gerrc.ErrFailedPrecondition, "fraud revision number mismatch: %d != %d", revision.Number, msg.FraudRevision)
 		ctx.Logger().Error("Fraud proposal", "error", err)
 		return nil, err
 	}
--- a/x/rollapp/keeper/fraud_proposal_test.go
+++ b/x/rollapp/keeper/fraud_proposal_test.go
@@ -1,8 +1,10 @@
 package keeper_test
 
 import (
+	"cosmossdk.io/errors"
 	govtypes "github.com/cosmos/cosmos-sdk/x/gov/types"
 	"github.com/dymensionxyz/dymension/v3/x/rollapp/types"
+	"github.com/dymensionxyz/gerr-cosmos/gerrc"
 )
 
 // TODO: test slashing and Rewardee
@@ -12,10 +14,11 @@ func (s *RollappTestSuite) TestSubmitRollappFraud() {
 	initialHeight := uint64(1)
 
 	testCases := []struct {
-		name          string
-		msgRevision   uint64
-		msgHeight     uint64
-		expectedError bool
+		name            string
+		msgRevision     uint64
+		msgHeight       uint64
+		expectedError   bool
+		removeRevisions bool
 	}{
 		{
 			name:          "first revision proposal",
@@ -40,6 +43,13 @@ func (s *RollappTestSuite) TestSubmitRollappFraud() {
 			msgRevision:   5,
 			msgHeight:     300,
 			expectedError: true,
+		},
+		{
+			name:            "height without revision",
+			msgRevision:     0,
+			msgHeight:       40,
+			expectedError:   true,
+			removeRevisions: true,
 		},
 	}
 
@@ -86,12 +96,16 @@ func (s *RollappTestSuite) TestSubmitRollappFraud() {
 
 			// assert revision correctness
 			rollapp = s.k().MustGetRollapp(s.Ctx, rollappId)
-			s.Require().EqualValues(rollapp.GetRevisionForHeight(1).Number, 0)
-			s.Require().EqualValues(rollapp.GetRevisionForHeight(49).Number, 0)
-			s.Require().EqualValues(rollapp.GetRevisionForHeight(50).Number, 1)
-			s.Require().EqualValues(rollapp.GetRevisionForHeight(55).Number, 1)
-			s.Require().EqualValues(rollapp.GetRevisionForHeight(120).Number, 2)
-			s.Require().EqualValues(rollapp.GetRevisionForHeight(300).Number, 2)
+			for height, expectedRevision := range map[uint64]uint64{1: 0, 49: 0, 50: 1, 55: 1, 120: 2, 300: 2} {
+				revision, found := rollapp.GetRevisionForHeight(height)
+				s.Require().True(found)
+				s.Require().EqualValues(expectedRevision, revision.Number)
+			}
+
+			if tc.removeRevisions {
+				rollapp.Revisions = nil
+				s.k().SetRollapp(s.Ctx, rollapp)
+			}
 
 			msg := &types.MsgRollappFraudProposal{
 				Authority:              s.App.AccountKeeper.GetModuleAddress(govtypes.ModuleName).String(),
@@ -108,6 +122,9 @@ func (s *RollappTestSuite) TestSubmitRollappFraud() {
 				s.Require().NoError(err)
 			} else {
 				s.Require().Error(err)
+				if tc.removeRevisions {
+					s.Require().True(errors.IsOf(err, gerrc.ErrFailedPrecondition))
+				}
 			}
 		})
 	}
--- a/x/rollapp/types/rollapp.go
+++ b/x/rollapp/types/rollapp.go
@@ -140,19 +140,18 @@ func (r Rollapp) LatestRevision() Revision {
 	return r.Revisions[len(r.Revisions)-1]
 }
 
-// TODO: rollapp type method should be more robust https://github.com/dymensionxyz/dymension/issues/1596
-func (r Rollapp) GetRevisionForHeight(h uint64) Revision {
+func (r Rollapp) GetRevisionForHeight(h uint64) (Revision, bool) {
 	for i := len(r.Revisions) - 1; i >= 0; i-- {
 		if r.Revisions[i].StartHeight <= h {
-			return r.Revisions[i]
+			return r.Revisions[i], true
 		}
 	}
-	return Revision{}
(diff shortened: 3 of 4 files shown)


omriai2 (approved):
**APPROVE** — this is a faithful, minimal fix for the tracked ambiguity in `GetRevisionForHeight`: the not-found case is now explicit in the signature, both production call sites handle it, the only behavior narrowing is on malformed state that no rollapp-construction path can produce, and committed regression tests pin both the type-level semantics and the gov handler's rejection.

**Index (no blocking findings):**

1. *Non-blocking* — Keeper regression case asserts `ErrFailedPrecondition`, which both the not-found and the pre-existing mismatch branch return, so it doesn't pin which branch fired. Covered in substance by the new type-level unit test; no change requested.
2. *Non-blocking* — `LatestRevision` keeps the same zero-value-on-empty pattern and its "should happen only in tests" comment, now sitting next to code that treats an empty revision list as reachable. Explicitly scoped out by the issue triage; worth a follow-up issue, not a change here.

<details><summary>Verification I performed (what I checked and ruled out)</summary>

**Call-site completeness.** `GetRevisionForHeight` has exactly two production callers: `IsRevisionStartHeight` in the same file and the fraud handler. `IsRevisionStartHeight` has exactly one production caller, `IsFirstHeightOfLatestFork`. Both were updated; no other package (lightclient, ibctesting, testutil) touches either.

**Behavior preservation at the surviving caller.** `IsFirstHeightOfLatestFork` is gated on `DidFork()` (more than one revision) and `revision == latest`, so `latest.Number >= 1` whenever it can return true. Under the old code the not-found sentinel had `Number == 0`, which already failed that comparison. The only input where old and new differ is `IsRevisionStartHeight(0, 0)` against a revision list whose first entry starts above zero — previously a spurious `true`, now `false`. That input is unreachable from the fork predicate, so the fork path is unchanged and strictly safer.

**Reachability of the new rejection.** Every construction path seeds `Revisions` with `{Number: 0, StartHeight: 0}` — `NewRollapp` and the v4 upgrade's rollapp converter — and `BumpRevision` only appends. Since heights are unsigned, the first revision covers every height for any correctly-constructed rollapp, so no live rollapp loses the ability to be fraud-proposed against. The not-found path is reachable only through an empty or gapped revision list, which is possible because `ValidateBasic` never validates the `Revisions` field and genesis import writes rollapps verbatim. Rejecting there is the right call: hard-forking a revisionless rollapp would append `{Number: 1, StartHeight: N}` and leave every height below `N` permanently uncovered.

**That the guard fires before side effects.** The `found` check sits ahead of the sequencer-punish call and ahead of `HardFork`, so a rejected proposal mutates nothing.

**Determinism.** The only map iteration introduced is in the assertion loop of the keeper test — no consensus path.

**Regression coverage is real, not decorative.** The keeper case reuses the exact parameters of the existing passing case ("first revision proposal", revision 0 at height 40) and differs only by nilling the revision list. Under the old code that lookup returned `Revision{}` with `Number == 0`, matched `FraudRevision: 0`, and proceeded through `HardFork` to success — so the case fails pre-fix and passes post-fix. I confirmed `SetRollapp` marshals the rollapp as given with no revision side effects, so the nil list genuinely persists into the handler. The type-level test independently pins the `(Revision, bool)` contract on both sides of the first `StartHeight` and asserts the specific pre-fix false positive, `IsRevisionStartHeight(0, 0) == false`.

**Docs honesty.** No README or living doc describes revision lookup. `CHANGELOG.md` exists but its convention is dead in this repo — last touched at #1695, and none of the four most recently merged PRs on this branch's ancestry updated it — so omitting an entry matches current practice.

**Independent review passes.** I ran a five-agent code review (CLAUDE.md compliance, shallow bug scan, git-history context, prior-PR review comments, comment/doc accuracy) and a security review over the diff. No agent returned a blocking finding above the reporting threshold. The history pass confirmed the zero-value fallthrough was never a deliberate sentinel — it was the loop-miss default, flagged as debt by the same engineer who wrote it. The prior-PR pass surfaced that this exact ambiguity was objected to in review on #1476 ("not really a guard, it will cause a silent failure"), that `ErrFailedPrecondition` is the established error family for this check per #1520, and that the structured log form `("Fraud proposal", "error", err)` matches the format corrected in #1773 — the new log line uses it. The security pass returned zero qualifying findings and confirmed the change only ever adds a rejection path; it never widens acceptance.

**CI.** All required checks green on the head commit: `build` (33m45s), `Analyze`/`CodeQL`, `golangci-lint`, `codecov/patch`, `codecov/project`, `label-prs`.

</details>

<details><summary>Finding 1 — Keeper test's error assertion doesn't discriminate between the two precondition branches (non-blocking)</summary>

The `height without revision` case asserts `errors.IsOf(err, gerrc.ErrFailedPrecondition)`. Both the new not-found branch and the pre-existing revision-mismatch branch wrap that same sentinel, so the assertion alone cannot tell you which one rejected the proposal. In this specific case the mismatch branch is unreachable — with a nil revision list the handler returns before reaching it — so the test is not wrong, just weaker than it reads.

This is fully compensated elsewhere: the new type-level test asserts the lookup contract directly rather than through the error type, which is the more precise place to pin it. Substring-matching the error message would couple the test to wording for little gain. No change requested; I note it only so a future reader doesn't mistake the assertion for branch discrimination.

</details>

<details><summary>Finding 2 — `LatestRevision` retains the pattern this PR just declared unsafe (non-blocking, follow-up)</summary>

`LatestRevision` still returns a synthetic `Revision{Number: 0}` for an empty revision list, under a comment stating that case "should happen only in tests." The premise of this PR is that an empty or malformed revision list *is* reachable in production, because nothing validates the `Revisions` field on genesis import. Two production call sites consume `LatestRevision` unguarded — the state-update revision check and the light-client update revision check — and would compare against that fabricated revision 0 rather than failing.

The issue's triage explicitly scoped this out ("`LatestRevision` has the same zero-value pattern but is explicitly documented and is not what this issue requests — leave it"), and the comment sits on lines this PR does not modify, so I am not holding the merge for it. It is a genuine loose end and a good follow-up issue: either give `LatestRevision` the same explicit not-found signal, or validate `Revisions` in `ValidateBasic` so the malformed state stops being reachable at all — at which point the comment becomes true again.

</details>

VERDICT: APPROVE

PR #2212, opened 2026-08-31
refactor(errors): preserve wrapped error chains

### LINKED ISSUE #1280: Refactor usage of `err.Error()`
code like
```
errorsmod.Wrap(ErrInvalidCreatorAddress, err.Error())
```

should be refactored to
```
errors.Join(ErrInvalidCreatorAddress, err)
```

## AI triage
I have everything I need. The repo has no CONTRIBUTING.md (that side-task never happened), only 1 chain-losing `%v` embed remains, ~75 `Wrap/Wrapf(..., err.Error())` construction sites remain across ~30 files, and ~20 test files still assert on error strings. Here is the re-triage.

---

## Research summary

The issue (filed by `mtsitrin`, Oct 2024, `good first issue`) asks that error-construction sites of the form `errorsmod.Wrap(ErrX, err.Error())` be refactored to `errors.Join(ErrX, err)`. Converting the underlying error to a string before wrapping destroys the error chain: `errors.Is`/`errors.As` can no longer see the inner error, and callers are pushed toward string matching.

History confirmed against the thread, linked PRs, and the clone at commit `945d736` (2026-08-15):

- **Done:** PR dymensionxyz/dymension#1286 by `AshutoshKD` was merged (Oct 2024), converting the pattern in five `x/rollapp/types` message-validation files. This establishes maintainer approval of the exact `errors.Join(ErrX, err)` replacement.
- **Reopened:** `mtsitrin` reopened the issue because more instances remain, explicitly inviting fixes in other modules (naming `x/sponsorship`, `x/sequencers`).
- **Abandoned:** A follow-up draft PR dymensionxyz/dymension#1305 by `AshutoshKD` (rollapp + sequencer types) was closed as stale without merging — its changes never landed; the sequencer files it touched still carry the pattern in the clone.
- **Latest human update (per `danwt`, June 2025):** the same problem also appears as embedding errors into other errors via `fmt.Errorf` with `%v`/`%w`, and as the closely related anti-pattern of asserting on errors with `strings.Contains(...)` in tests (and logic). This update reports nothing as fixed; it widens the observed scope.
- **Side task from the thread:** `mtsitrin` asked `danwt` to add error-handling guidelines to `CONTRIBUTING.md`. Confirmed in the clone: no `CONTRIBUTING.md` exists, so that step was never done and remains owned by `danwt` per the thread; it is not part of this fix.

Verified current state of the clone (read-only): roughly 75 `errorsmod.Wrap`/`Wrapf` call sites in non-test code still embed `err.Error()` (both as the message argument and as a `%s`/`%v` format argument), spread over ~30 files across `x/dymns`, `x/denommetadata`, `x/iro`, `x/lockup`, `x/eibc`, `x/agent`, `x/bridgingfee`, `x/sequencer`, `x/rollapp`, `x/delayedack`, and `app/upgrades/v5`. Exactly one chain-losing `fmt.Errorf("...%s", err.Error())` embed remains outside tests (`x/bridgingfee`); the many `fmt.Errorf("...: %w", err)` sites preserve the chain and are idiomatic Go. Around 20 test files assert on error text via `Contains(err.Error(), ...)` or equivalent. What is left is owned by whoever picks up this issue — no contributor currently holds it (`AshutoshKD`'s draft died in 2024).

## Root cause hypothesis

Not a bug — a codebase-wide style debt. Early code stringified inner errors when wrapping registered/sentinel errors, which severs the error chain. Partial cleanup landed in one module; the rest of the codebase was never swept, and tests grew a dependency on error-message strings that the stringification pattern encourages.

## Areas involved

- Error-construction sites in module code: `x/dymns`, `x/denommetadata`, `x/iro`, `x/lockup`, `x/eibc`, `x/agent`, `x/bridgingfee`, `x/sequencer`, `x/rollapp`, `x/delayedack`, and the `app/upgrades/v5` param-validation types.
- Test files that assert on error strings (notably `x/sponsorship`, `x/eibc`, `x/sequencer`, `x/rollapp` types tests and `app/ante` tests).

## Acceptance criteria

Scope is settled as follows (the smaller, precedent-backed sweep; wholesale rewriting of idiomatic `fmt.Errorf("...: %w", err)` chains into `errors.Join` is explicitly out of scope — those preserve the chain and converting them is large opinionated churn beyond what the merged precedent sanctions):

1. No non-test call site constructs an error by embedding `err.Error()` — neither `errorsmod.Wrap(ErrX, err.Error())` nor `errorsmod.Wrapf(ErrX, "...%s/%v", err.Error())` nor `fmt.Errorf` with a stringified error. Replace with `errors.Join(ErrX, err)` following the merged PR #1286 precedent; where a Wrapf carries additional context beyond the inner error, keep that context (e.g. join the sentinel with a `fmt.Errorf("context: %w", err)`), so no diagnostic information is lost.
2. The one remaining chain-losing `%s`/`%v` embed outside tests is converted to `%w` or `errors.Join`.
3. `err.Error()` uses that merely *display* an error (event attributes, log fields, API response strings) are left untouched — the sweep targets error construction only.
4. Test assertions that string-match errors (`Contains(err.Error(), ...)`, `strings.Contains`) are converted to `require.ErrorIs` (or `require.ErrorContains` against a sentinel's message) wherever a sentinel error exists to match on; assertions on error text that has no sentinel may remain as `ErrorContains`.
5. Error return sites keep returning a non-nil error exactly where they did before — this refactor changes error identity/formatting, never control flow.
6. `go build ./...`, `go test ./...`, and `golangci-lint` (config present at `.golangci.yml`) pass; tests broken by changed error strings are updated per criterion 4, not silenced.

Note for the worker: message texts produced at these sites change shape (Join renders differently than Wrap), which is acceptable per the maintainers' merged precedent; registered-error codespace/ABCI codes must still resolve — `cosmossdk.io/errors` walks `Unwrap() []error`, and the merged PR relied on that.

## Open questions

None.

---

The fix is a contained, mechanical, CI-gated refactor with an explicit maintainer-merged precedent for the exact transformation, on a `good first issue`. It touches error paths broadly but changes no control flow, no funds/auth surface, and no migration — standard route, full ceremony because error identity in a chain node is behavior, not copy.

omriai1:
Closes #1280

This refactor removes error construction that stringified underlying causes, restoring `errors.Is`/`errors.As` traversal while preserving contextual diagnostics and registered ABCI metadata. Joined errors retain the pre-refactor single-line `context: cause: sentinel` message shape and avoid routing Cosmos errors through `fmt.Formatter`, so user-facing output does not gain source paths.

- Error chains are preserved across all affected construction sites.
- Registered Cosmos ABCI codespace/code behavior is preserved.
- Existing single-line error ordering is preserved without source-location leakage.
- Error assertions use `ErrorIs`/`ErrorContains`.

<details>
<summary>Implementation details</summary>

- Replaced direct `Wrap(..., err.Error())` sites with multi-error joins.
- Replaced contextual `Wrapf(..., err.Error())` sites with `internal/errors.Joinf`, which formats context separately and reads the cause through `Error()`.
- Implemented stable `cause: sentinel` and `context: cause: sentinel` rendering while exposing both errors through `Unwrap() []error`.
- Preserved the registered sentinel's `ABCICode` and `Codespace` for `cosmossdk.io/errors.ABCIInfo`.
- Returned the join implementation by pointer for safe comparability and matched wrapping behavior by returning nil for nil causes.
- Added regression coverage for message ordering, single-line output, Cosmos source-path suppression (including `AggregationHook.Validate`), chain traversal, nil behavior, comparability, and ABCI metadata.
- Left logs, events, status/API strings, invariant formatting, simulation descriptions, and control flow untouched.

</details>

<details>
<summary>Acceptance-criteria audit</summary>

1. No non-test error constructor embeds `err.Error()`; contextual diagnostics use `Joinf` without formatting the cause value.
2. The bridging-fee aggregation-hook validation uses `Joinf`, preserving its cause without formatting the Cosmos error value.
3. Display-only stringification call sites remain unchanged, and affected constructed errors retain their prior single-line ordering.
4. Applicable test assertions use `require.ErrorIs` or `ErrorContains`; text checks without a matching sentinel remain `ErrorContains`.
5. Return branches and nil/non-nil control flow are unchanged.
6. After merging main at `8497b49e`, full race tests, build, and the repository-pinned lint gate passed on the merged tree.

</details>

## Proof of execution

- **Build and Test / Build:** `go build -p 2 ./...` passed after merging main at `8497b49e`.
- **Build and Test / Test & Coverage behavior:** `go test -race -p 2 -timeout 30m ./... -count=1` passed across the full repository, including the error-chain regressions and sibling changes.
- **Sponsorship merge resolution:** `go test ./x/sponsorship/types ./x/sponsorship/keeper -count=1` passed with the sibling's error classifications and this PR's preserved causes.
- **golangci-lint:** repository-pinned v2.1.6, using `GOTOOLCHAIN=go1.24.6`, passed `golangci-lint run --concurrency 2 --allow-parallel-runners ./...` with zero issues.
- **Formatting:** configured goimports/gofumpt applied to both conflict resolutions; `golangci-lint fmt --diff` and `git diff --cached --check` were clean.

**Not verified:** external e2e workflows require the separate dymensionxyz/e2e-tests workflow environment.


omriai2 (changes requested):
**REQUEST_CHANGES** — The refactor correctly restores `errors.Is`/`errors.As` traversal and preserves ABCI codespace/code (verified), but the way context is re-attached routes every cause through `fmt`'s `Formatter` path, which injects a source `file:line` into user-facing error messages, and the stdlib `errors.Join` rendering turns every affected message into two lines with the sentinel first — both chain-wide user-visible regressions, and the PR deletes the one test that documented the first trap.

### Index

1. **Blocking** — `fmt.Errorf("...: %w", err)` on a `cosmossdk.io/errors` cause embeds a source `file:line` into the error message, at every converted site whose cause came from `errorsmod.Wrap/Wrapf`. This is the exact trap the deleted `TestErrorString` guarded. *Accept when:* joined errors carry no `[…​.go:NN]` fragment and a committed regression test asserts that over an `errorsmod`-wrapped cause.
2. **Blocking** — `internal/errors.Join` inherits stdlib `errors.Join`'s rendering, so every affected message becomes `"<sentinel>\n<context>"` instead of `"<context>: <sentinel>"` — multi-line ABCI logs, CLI output and gRPC status messages, contradicting the PR body's "display-only stringification remains unchanged". *Accept when:* the helper's `Error()` is single-line and preserves the pre-PR `context: sentinel` ordering.
3. **Non-blocking** — `joinedError` is a non-comparable struct returned by value, reachable by an unguarded `==` inside `cosmossdk.io/errors`; plus `Join(nil, nil)` yields a non-nil error that panics on `Error()`, and `%+v` debug ABCI output loses the stack frame `errorsmod.Wrap` used to attach.

---

<details><summary>Finding 1 — `%w` on a cosmos-sdk error leaks a source file:line into every affected message; the guard test for this was deleted</summary>

**What the code does.** The refactor replaces `errorsmod.Wrapf(sentinel, "context: %v", err.Error())` with `dymerrors.Join(sentinel, fmt.Errorf("context: %w", err))`. The old form deliberately passed a *string* so `fmt` never touched the error value. The new form passes the error value to `%w`.

**Why that matters.** `cosmossdk.io/errors@v1.0.1` gives `*wrappedError` a `fmt.Formatter` implementation (`stacktrace.go`), and its `%v` branch writes the message followed by a compressed source frame:

```go
// Format works like pkg/errors, with additions.
// %s is just the error message
// %+v is the full stack trace
// %v appends a compressed [filename:line] where the error was created
func (e *wrappedError) Format(s fmt.State, verb rune) {
	...
	} else {
		_, _ = fmt.Fprint(s, e.Error())
		writeSimpleFrame(s, stack[0])
	}
}
```

Go's `fmt` treats `%w` as `%v` when the operand is a `Formatter` (`handleMethods` rewrites the verb to `'v'` before the `Formatter` check). So any `fmt.Errorf("...: %w", err)` where `err` originated from `errorsmod.Wrap`/`Wrapf` bakes the frame into the message string permanently — it is not a formatting-time artifact, it is now part of `Error()`.

**The repo already knew this.** `x/iro/keeper/params_test.go` carried `TestErrorString`, whose own comments spell it out — "Incorrect version! `%v` turns err into Stringer that prints the filepath" / "Correct version! ... `.Error()` hides the filepath" — and asserted `NotRegexp(`\[.*\.go.*\]`, ...)`. This PR deletes that test as "obsolete" while reintroducing precisely the behaviour it forbade. `errorsmod.Wrapf` is still used throughout the tree, so the test was not obsolete.

**Evidence.** Probe run against three real converted sites in this checkout (temporary package, since removed; tree verified clean afterwards):

```
$ go test ./tmpprobe/ -run TestClassProbe -v -count=1
dymns Params.Validate                  leaks-source-location=true
    "invalid argument\nprice params: Dym-Name price steps must have at least 4 steps: invalid argument [/home/workforce/work/worktrees/dymensionxyz__dymension__1280/x/dymns/types/params.go:262]"
dymns GenesisState.Validate            leaks-source-location=true
    "invalid argument\nDym-Name 'bad name!': name is not a valid dym name: invalid argument [/home/workforce/work/worktrees/dymensionxyz__dymension__1280/x/dymns/types/dym_name.go:26]"
dymns SellOrder.Validate               leaks-source-location=true
    "invalid argument\nSO highest bid is invalid: SO bidder is not a valid bech32 account address: invalid argument [/home/workforce/work/worktrees/dymensionxyz__dymension__1280/x/dymns/types/sell_order.go:136]"
dymns MsgPlaceSellOrder.ValidateBasic  leaks-source-location=false
    "name is not a valid dym name: : invalid argument"
--- PASS: TestClassProbe (0.00s)
```

Note the leaked value is the *absolute build-machine path*: `writeSimpleFrame` only truncates at `github.com/`, and a build tree that isn't under a `github.com` directory leaks the whole path. Release builds use `-trimpath` (Makefile), which reduces it to `dymensionxyz/dymension/v3/x/dymns/types/params.go:262` — still a source location embedded in a string that reaches tx result logs, CLI output and block explorers where it previously did not.

This is *not* a consensus hazard — CometBFT excludes `Log` from `LastResultsHash` — but it is a build-environment disclosure in public error output, and it makes the same logical error render differently depending on where the binary was built.

**Class.** Every new `dymerrors.Join(sentinel, fmt.Errorf("...: %w", err))` whose cause can be an `errorsmod` error. Confirmed in the dymns types package and its duplicate under `app/upgrades/v5/types/dymns`; by construction the same applies to the iro plan/params conversions, the denommetadata and delayedack middleware conversions, and the rollapp/dymns message validators. Please sweep all of them rather than patching the three I proved.

**Acceptance criterion.** Attach context without routing the cause through `fmt`'s verb machinery, while keeping `Unwrap` intact. Two shapes I would accept:
- Extend the helper to own the context — e.g. `dymerrors.Joinf(sentinel, cause, "price params")` — building the message from `cause.Error()` and returning `cause` from `Unwrap()`.
- Introduce a small context type whose `Error()` is `fmt.Sprintf("%s: %s", msg, cause.Error())` and whose `Unwrap() error` returns `cause`, and use it in place of `fmt.Errorf`.

Plus: a committed regression test replacing the deleted `TestErrorString` — construct a joined error over an `errorsmod.Wrap`-produced cause and assert its `Error()` does not match `\[.*\.go.*\]`. Removing the old guard without a replacement is not acceptable on its own.

</details>

<details><summary>Finding 2 — joined errors render as two lines with the sentinel first, changing every affected user-facing message</summary>

**What changed.** `internal/errors.Join` delegates rendering to stdlib `errors.Join`, whose `Error()` is the operands separated by `"\n"`, sentinel first. So the message shape flips and gains a newline:

- Before: `"price params: Dym-Name price steps must have at least 4 steps: invalid argument"`
- After: `"invalid argument\nprice params: Dym-Name price steps must have at least 4 steps"`

Where the joined error is subsequently wrapped, the sentinel text lands *inside* the message with a line break through it. From the same probe run, `SequencerMetadata.Validate` now yields:

```
"invalid rpcs URLs: url: invalid argument\ninvalid URL: parse \"https://example.com/%\": invalid URL escape \"%\""
```

**Where it surfaces.** These strings are the ABCI `log` (`defaultErrEncoder` is literally `err.Error()`), so they reach `dymd tx` output, tx-result JSON, and explorers. They also reach gRPC status messages — I verified the code is preserved but the message is not:

```
$ go test ./tmpprobe/ -run TestProbeGRPCStatus -v -count=1
OLD Wrap(gerrc.ErrNotFound): code=NotFound ok=true msg="codespace gerrc code 4: not found: ctx"
NEW Join(gerrc.ErrNotFound): code=NotFound ok=true msg="not found\nctx"
OLD Wrap(ErrInvalidArgument): code=InvalidArgument
NEW Join(ErrInvalidArgument): code=InvalidArgument
--- PASS
```

(The good news in that transcript: gRPC codes survive, because `status.FromError` uses `errors.As`, which traverses `Unwrap() []error`. The ABCI codespace/code path is likewise fine — `joinedError` implements `coder`/`codespacer` directly, and the committed `TestValidateURLsPreservesErrorChain` passes here. So the metadata half of the design is sound; only the rendering is not.)

**Why this is blocking rather than cosmetic.** The PR body's acceptance audit asserts "Display-only stringification remains unchanged" and "Display-only `err.Error()` uses and control flow are unchanged". That is true only of the *call sites* that print errors; the strings they print changed at essentially every converted site. A reviewer or operator reading the PR body would not learn that the chain's entire validation-error surface became multi-line. Line-oriented log consumers and any tooling that reads one error per line are affected.

**Acceptance criterion.** The helper's `Error()` must produce a single line and preserve the pre-PR ordering that `errorsmod.Wrap` produced (`"<context>: <sentinel>"`), so the refactor is confined to the Go error chain as the PR title and scope claim. Implement `Error()` on the join type explicitly instead of inheriting stdlib `errors.Join`'s rendering — `Unwrap() []error` still gives you `errors.Is`/`errors.As` regardless of how you render. One fix can satisfy both this and Finding 1. Please also correct the PR body claim once the rendering is settled.

</details>

<details><summary>Finding 3 — non-blocking: comparability, nil, and stack-trace hazards in the helper</summary>

Three small things in `internal/errors/join.go`. None hold the merge; all are cheap to close while you are in the file.

**Non-comparable value type reachable by an unguarded `==`.** `Join` returns `joinedError` by value, and the struct has a `causes []error` field, making it non-comparable. `cosmossdk.io/errors`' `*wrappedError.Is` walks its chain with a raw interface comparison and no comparability guard:

```go
func (e *wrappedError) Is(target error) bool {
	if e == target { return true }
	w := e.Cause()
	for {
		if w == target { return true }
		...
```

Go's own `errors.Is` guards this (it checks `reflectlite.TypeOf(target).Comparable()` first), but the SDK's does not. So `errors.Is(errorsmod.Wrap(joinedA, "ctx"), joinedB)` panics with "comparing uncomparable type". Unreachable from the current call sites — targets are always registered sentinels — but returning `*joinedError` instead of `joinedError` removes the hazard entirely at zero cost.

**`Join(nil, nil)`.** `errors.Join(nil, nil)` returns `nil`, so the helper returns a non-nil `error` whose embedded `error` field is nil; calling `Error()` on it panics. Also unreachable today (every call site is inside an `if err != nil`), but `errorsmod.Wrap` returns nil for a nil input, and matching that contract — an early return when `cause == nil` — makes `Join` safe to compose the same way.

**Debug-mode stack trace.** `errorsmod.Wrap` attaches a stack via `errors.WithStack` when none is present, and `ABCIInfo(err, debug=true)` renders it with `%+v`. `joinedError` implements no `Format`, so `--trace` output for these errors now degrades to the plain message. Worth a line in the helper's doc comment so the next reader knows it is deliberate rather than an oversight.

</details>

---

Things I checked that are fine and need no action: ABCI codespace/code preservation (the committed `TestValidateURLsPreservesErrorChain` passes locally and independently reproduces); gRPC code preservation via `errors.As` traversal; no remaining `err.Error()` inside non-test error constructors — the survivors are all logs, `status.Error`, invariant formatting, simulation descriptions and query-response fields; the test-assertion churn (`Contains(err.Error(), x)` → `ErrorContains(err, x)`) is equivalence-preserving and not a weakening; the `errors` → `sdkerrors` import rename in the eibc authorization file is mechanical with no logic change; `internal/errors` matches the existing `internal/collcompat` convention; `Contributing.md`'s error guidance remains accurate under this change. Proof-of-execution section shape is sound — four receipted claims plus an owned "Not verified" for e2e — and all four are CI-covered, so I did not open transcripts; `golangci-lint` is green on the PR, `build` and `Analyze` were still pending at review time, which is not a review blocker.

VERDICT: REQUEST_CHANGES

omriai2 (changes requested):
**REQUEST_CHANGES** — The `Joinf` rewrite is a good fix that fully resolves the message-shape regression and the helper hardening, and it correctly eliminates the source-path leak at every `Join` site, but the single site the author converted to a bare `fmt.Errorf` instead of the helper still leaks a source location, so prior Finding 1 survives at one place.

### Index

1. **Blocking (prior Finding 1, partially resolved — survives at one site)** — `x/bridgingfee/types/types.go` still routes a Cosmos error through `%w`, so `AggregationHook.Validate` embeds `[cosmos/cosmos-sdk@v0.50.14/types/address.go:173]` in its message. *Accept when:* that construction stops formatting the error value (use `dymerrors.Joinf`, or `err.Error()`), and the tree has zero newly-added `%w`-on-a-Cosmos-error sites.
2. **Resolved (prior Finding 2)** — sentinel-first / newline rendering. No action.
3. **Resolved (prior Finding 3)** — comparability, nil handling, stack-trace documentation. No action.
4. **Nits, non-blocking** — `registered == nil` is unguarded; `Joinf`'s variadic args are an unguarded re-entry point for the same leak; one doc-comment phrase over-claims.

---

<details><summary>Finding 1 — prior blocker survives at x/bridgingfee/types/types.go: `%w` on a Cosmos error still leaks a source location</summary>

**What was fixed.** The new `internal/errors.Joinf` takes the context format and args directly and builds the message from `cause.Error()`, so the cause value never reaches `fmt`'s verb machinery:

```go
func join(registered, cause error, context string) error {
	if cause == nil {
		return nil
	}
	message := cause.Error()
	if context != "" {
		message = context + ": " + message
	}
	message += ": " + registered.Error()
	...
```

Every `dymerrors.Join(sentinel, fmt.Errorf("...: %w", err))` from the first commit is gone — all 30-odd sites across `app/upgrades/v5`, bridgingfee msgs, delayedack, denommetadata, dymns types, iro keeper/types, rollapp chain_id and sponsorship now call `Joinf(sentinel, err, "context", args...)`. The deleted `TestErrorString` guard has been replaced by a better-placed one in `internal/errors/join_test.go`, which asserts `NotRegexp(\[.*\.go.*\])` directly on the helper's output. That half of the finding is closed.

**What survives.** The one site the author converted to a plain `fmt.Errorf` rather than the helper was not revisited. It is the only newly-added `%w` line in the entire branch:

```
$ git diff main...HEAD -- '*.go' | grep -n '^+.*%w' | grep -v '^+++'
593:+			return fmt.Errorf("owner address is invalid: %w", err)
```

```
$ git diff main...HEAD -- x/bridgingfee/types/types.go
 func (h AggregationHook) Validate() error {
 	if h.Owner != "" {
 		if _, err := sdk.AccAddressFromBech32(h.Owner); err != nil {
-			return fmt.Errorf("owner address is invalid: %s", err.Error())
+			return fmt.Errorf("owner address is invalid: %w", err)
```

`sdk.AccAddressFromBech32` propagates `VerifyAddressFormat`'s error unchanged, and that function returns `errorsmod.Wrap(...)` — a `*wrappedError` carrying a stack trace and a `fmt.Formatter` — on two reachable branches:

```go
	if len(bz) == 0 {
		return errorsmod.Wrap(sdkerrors.ErrUnknownAddress, "addresses cannot be empty")
	}
	if len(bz) > address.MaxAddrLen {
		return errorsmod.Wrapf(sdkerrors.ErrUnknownAddress, "address max length is %d, got %d", address.MaxAddrLen, len(bz))
	}
```

Both are reachable with a checksum-valid bech32 string carrying an empty or over-length payload — the prefix check passes, the decode succeeds, and only the length verification fails. Confirmed by a narrow probe (temporary package, removed; tree clean afterwards):

```
$ go test ./tmpprobe/ -run TestBridgingFeeOwnerLeak -v -count=1
payload len=0
  Error() = "owner address is invalid: addresses cannot be empty: unknown address [cosmos/cosmos-sdk@v0.50.14/types/address.go:173]"
  leaks-source-location=true
payload len=300
  Error() = "owner address is invalid: address max length is 255, got 300: unknown address [cosmos/cosmos-sdk@v0.50.14/types/address.go:177]"
  leaks-source-location=true
--- PASS: TestBridgingFeeOwnerLeak (0.00s)
```

Before this PR the same inputs produced `"owner address is invalid: addresses cannot be empty: unknown address"` — no bracketed frame. This is a regression introduced by the branch, not pre-existing.

**Reachability.** `AggregationHook.Validate()` is called from `GenesisState.Validate()` at two places in `x/bridgingfee/types/genesis.go`, which is what `x/bridgingfee/module.go`'s `ValidateGenesis` runs — so the string surfaces in genesis validation output (`dymd genesis validate`, chain start, upgrade tooling). Note the leaked frame here is a module-cache path rather than a repo path, so it discloses the builder's Go module layout and pinned SDK version; under a repo-local frame it would disclose the build directory, exactly as the earlier probes showed.

**Class.** The class is unchanged from last round: *any* construction that hands a possibly-Cosmos error to a `fmt` verb (`%w`, `%v`, `%s` on the value rather than on `.Error()`). It now has exactly one instance. The other `%w` occurrences that a grep turns up in these files — in `x/eibc/keeper/msg_server.go`, `x/lockup/keeper/msg_server.go`, `x/rollapp/types/rollapp.go`, `x/sequencer/types/metadata.go` — are all pre-existing lines untouched by this branch, so they are out of scope here; I mention them only so the sweep below doesn't get confused by them.

**Acceptance criterion.** Make `AggregationHook.Validate` attach its context without formatting the error value. Either shape is fine:
- `return dymerrors.Joinf(gerrc.ErrInvalidArgument, err, "owner address is invalid")` — consistent with the rest of the refactor and gives this site a sentinel it currently lacks; or
- `return fmt.Errorf("owner address is invalid: %s", err.Error())` — restoring the pre-PR text if you'd rather leave this function returning a bare error.

Plus: confirm `git diff main...HEAD -- '*.go' | grep '^+.*%w'` returns nothing (or only lines whose operand provably cannot be a Cosmos error). The existing `internal/errors` test file already covers the helper path; no new test is required for this one call site if you route it through `Joinf`, but if you keep a bare `fmt.Errorf` please add a case to `x/bridgingfee/types/types_test.go` asserting the message carries no `[….go:NN]`.

</details>

<details><summary>Finding 2 — RESOLVED: message shape restored to single-line `context: cause: sentinel`</summary>

`join()` now composes the message itself instead of inheriting stdlib `errors.Join`'s newline rendering, and the ordering matches what `errorsmod.Wrap`/`Wrapf` produced before the refactor. Spot-checking the shapes against the pre-PR code: `Wrapf(sentinel, "ctx: %v", err.Error())` produced `"ctx: <cause>: <sentinel>"`, and `Joinf(sentinel, err, "ctx")` produces the same; `Wrap(sentinel, err.Error())` produced `"<cause>: <sentinel>"`, and `Join(sentinel, err)` matches. The sequencer, dymns params, sponsorship msgs and rollapp chain_id sites I checked last round all line up.

The committed regression coverage asserts exact strings and explicitly forbids newlines. It passes locally:

```
$ go test ./internal/errors -run TestJoin -count=1 -v
=== RUN   TestJoinRendersCauseBeforeRegisteredError
--- PASS: TestJoinRendersCauseBeforeRegisteredError (0.00s)
=== RUN   TestJoinfDoesNotRenderCosmosErrorSourceLocation
--- PASS: TestJoinfDoesNotRenderCosmosErrorSourceLocation (0.00s)
=== RUN   TestJoinNilCauseReturnsNil
--- PASS: TestJoinNilCauseReturnsNil (0.00s)
=== RUN   TestJoinResultIsComparableWhenWrapped
--- PASS: TestJoinResultIsComparableWhenWrapped (0.00s)
=== RUN   TestJoinPreservesRegisteredABCIInfo
--- PASS: TestJoinPreservesRegisteredABCIInfo (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/internal/errors	0.010s
```

The PR body's acceptance-audit item 3 is now accurate. No action.

</details>

<details><summary>Finding 3 — RESOLVED: helper hardening (comparability, nil, stack trace)</summary>

All three sub-items are addressed:

- `join` returns `*joinedError`, so the SDK's unguarded `w == target` inside `*wrappedError.Is` can no longer panic on an uncomparable dynamic type. `TestJoinResultIsComparableWhenWrapped` pins this with an explicit `require.NotPanics` around `errors.Is(errorsmod.Wrap(first, "context"), second)`.
- `cause == nil` returns `nil`, matching `errorsmod.Wrap`'s contract, with `TestJoinNilCauseReturnsNil` covering both `Join` and `Joinf`.
- The absence of a stack trace is now stated in the `Join` doc comment.

No action.

</details>

<details><summary>Finding 4 — nits, none blocking</summary>

**`registered == nil` is unguarded.** `join` returns early on a nil cause but then calls `registered.Error()` unconditionally, so `Join(nil, err)` panics where `errorsmod.Wrap(nil, ...)` would have returned nil. Every call site passes a package-level sentinel, so this is unreachable today; a symmetric `if registered == nil { return cause }` would make the helper total for the same one line.

**`Joinf` args are an unguarded re-entry point.** The variadic args go straight to `fmt.Sprintf`, so passing an error there reopens exactly the Formatter leak the helper exists to prevent. No current call site does — they pass strings and ints — but the doc comment explains only why the *cause* is read through `Error()`. One added sentence saying args must not be error values would keep the next caller from undoing the fix.

**Doc comment over-claims.** `Join`'s comment says it "deliberately renders %+v as the stable single-line message", but `joinedError` implements no `Format` method — `%+v` falls through to `Error()` because the type satisfies `error`. The outcome matches the claim, so nothing is wrong; the wording just implies an explicit `Format` implementation a reader will go looking for and not find.

**Contributing.md.** Item 6 still says to wrap with `fmt.Errorf` / `errorsmod.Wrapf`. That remains true and is not made false by this change, so it is not a docs finding — but now that `internal/errors.Joinf` is the house pattern for sentinel-plus-cause, a line pointing at it would help the next contributor pick it over `Wrapf(..., "%v", err.Error())`. Optional.

</details>

---

Also verified this round and needing no action: `Unwrap() []error` still returns both operands so `errors.Is`/`errors.As` reach the sentinel and the cause; `Codespace()`/`ABCICode()` are unchanged and `TestJoinPreservesRegisteredABCIInfo` pins the ABCI metadata; every `Joinf` conversion in the fix commit is a faithful rewrite of its predecessor with the trailing `: %w`/`: %s` correctly dropped from the format string and no control-flow change; the sole newly-added `%w` in the branch is the one in Finding 1. Proof of execution has sound shape — five receipted claims plus an owned "Not verified" for the external e2e workflows — and all its claims are CI-covered, so I opened no transcripts; I ran only the single new test package as the deciding evidence for Finding 2's resolution. On the head commit `4f0633c9`, `golangci-lint`, `CodeQL`, `Analyze` and `label-prs` are green and `build` is still pending, which is not a review blocker.

VERDICT: REQUEST_CHANGES

omriai2 (approved):
**APPROVE** — The one surviving blocker is fixed at its source and pinned by a committed regression test that exercises the exact reachable input I demonstrated; all three prior findings are now resolved and this round surfaced nothing new.

### Index

1. **Resolved (prior Finding 1)** — `x/bridgingfee/types/types.go` no longer formats a Cosmos error value; it routes through `dymerrors.Joinf`, and a committed test asserts no source location appears. No action.
2. **Resolved (prior Finding 2)** — single-line `context: cause: sentinel` rendering. No action.
3. **Resolved (prior Finding 3)** — helper hardening (pointer return, nil handling, stack-trace documentation). No action.
4. **Nits carried forward, non-blocking** — unguarded `registered == nil`; `Joinf` variadic args are an unenforced re-entry point for the leak; one over-claiming doc-comment phrase; optional Contributing.md pointer.

---

<details><summary>Finding 1 — RESOLVED: the bridging-fee source-location leak is fixed at source and pinned by a test</summary>

The single site that survived the previous round now uses the helper instead of a bare `fmt.Errorf`:

```go
 func (h AggregationHook) Validate() error {
 	if h.Owner != "" {
 		if _, err := sdk.AccAddressFromBech32(h.Owner); err != nil {
-			return fmt.Errorf("owner address is invalid: %w", err)
+			return dymerrors.Joinf(sdkerrors.ErrInvalidAddress, err, "owner address is invalid")
 		}
 	}
 	return nil
 }
```

`Joinf` builds its message from `cause.Error()` and formats the context separately, so the `*wrappedError` returned by `VerifyAddressFormat` never reaches `fmt`'s Formatter path and the `[cosmos/cosmos-sdk@v0.50.14/types/address.go:173]` fragment I reproduced last round can no longer appear. This is one of the two solution shapes I named as acceptable; the author chose `sdkerrors.ErrInvalidAddress`, which is a better-fitting sentinel than the `gerrc.ErrInvalidArgument` I suggested, and it also upgrades this site from an untyped error (ABCI internal code 1) to a properly classified one.

The commit adds a regression test that constructs precisely the reachable input from my earlier probe — a checksum-valid bech32 string with an empty payload, which clears the prefix and decode checks and fails only on `VerifyAddressFormat`, the branch that returns a Cosmos Formatter error:

```go
func TestAggregationHookValidateDoesNotRenderCosmosErrorSourceLocation(t *testing.T) {
	owner, err := bech32.ConvertAndEncode(sdk.GetConfig().GetBech32AccountAddrPrefix(), nil)
	require.NoError(t, err)

	err = (AggregationHook{Owner: owner}).Validate()

	require.ErrorIs(t, err, sdkerrors.ErrInvalidAddress)
	require.NotRegexp(t, regexp.MustCompile(`\[.*\.go.*\]`), err.Error())
}
```

Deriving the prefix from `sdk.GetConfig()` rather than hardcoding it makes the test robust to whatever config the test binary carries — the failure mode my own first probe hit. It executes and passes:

```
$ go test ./x/bridgingfee/types -count=1 -v -run 'TestAggregationHookValidateDoesNotRenderCosmosErrorSourceLocation'
=== RUN   TestAggregationHookValidateDoesNotRenderCosmosErrorSourceLocation
--- PASS: TestAggregationHookValidateDoesNotRenderCosmosErrorSourceLocation (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/bridgingfee/types	0.076s
```

**Class sweep.** I re-ran the sweep across the whole branch for any added line that hands an error *value* to a `fmt` verb, widened beyond `%w` to `%v` and `%s`. Nothing remains — the surviving `%s` hits are all `Joinf` context arguments carrying strings (`m.Owner`, `dymName.Name`, `soBid.Bidder`, and so on), never errors, and there is no `dymerrors.Join(..., fmt.Errorf(...))` left anywhere in the tree. The class is closed, not just the instance.

**Message-shape side effect, checked and fine.** This site's text gains a trailing `: invalid address` from the sentinel, since it previously had none. The only assertion pinning it, `x/bridgingfee/types/genesis_test.go:233` (`errMsg: "owner address is invalid"`), is a substring check and is unaffected. The whole package passes:

```
$ go test ./x/bridgingfee/types -count=1 -run 'TestAggregationHook|TestGenesis|TestMsg'
ok  	github.com/dymensionxyz/dymension/v3/x/bridgingfee/types	0.069s
```

`fmt` remains legitimately imported in that file for eight other constructions, so the import is not stranded.

</details>

<details><summary>Finding 2 — RESOLVED (unchanged since last round): single-line `context: cause: sentinel` rendering</summary>

`internal/errors.join` composes the message itself rather than inheriting stdlib `errors.Join`'s newline separator, restoring the exact ordering `errorsmod.Wrap`/`Wrapf` produced before the refactor. Covered by `TestJoinRendersCauseBeforeRegisteredError` (exact-string equality plus an explicit `NotContains "\n"`) and `TestJoinfDoesNotRenderCosmosErrorSourceLocation`, both verified passing in the previous round and untouched by this commit. No action.

</details>

<details><summary>Finding 3 — RESOLVED (unchanged since last round): helper hardening</summary>

`join` returns `*joinedError` so the SDK's unguarded `w == target` inside `*wrappedError.Is` cannot panic on an uncomparable dynamic type (`TestJoinResultIsComparableWhenWrapped`); a nil cause returns nil, matching `errorsmod.Wrap`'s contract (`TestJoinNilCauseReturnsNil`); the absence of a stack trace is documented on the `Join` doc comment. ABCI codespace/code preservation remains pinned by `TestJoinPreservesRegisteredABCIInfo`. No action.

</details>

<details><summary>Finding 4 — nits carried forward, none blocking</summary>

Unchanged from the previous round and still optional. Listing them so they aren't lost, not to hold the merge:

**`registered == nil` is unguarded.** `join` returns early on a nil cause but then calls `registered.Error()` unconditionally, so `Join(nil, err)` would panic where `errorsmod.Wrap(nil, ...)` returned nil. Every call site passes a package-level sentinel, so this is unreachable today; a symmetric `if registered == nil { return cause }` makes the helper total for one line.

**`Joinf` args are an unguarded re-entry point.** The variadic args go straight to `fmt.Sprintf`, so a future caller passing an error there reopens exactly the Formatter leak this PR spent three commits closing. No current call site does — I verified all of them carry strings or ints. One sentence in the doc comment saying args must not be error values would keep the next contributor from undoing the fix.

**Doc comment over-claims.** `Join`'s comment says it "deliberately renders %+v as the stable single-line message", but `joinedError` implements no `Format` method — `%+v` falls through to `Error()` because the type satisfies `error`. The behaviour matches the claim; the wording just implies an explicit `Format` implementation a reader will go looking for and not find.

**Contributing.md.** Item 6 still says to wrap with `fmt.Errorf` / `errorsmod.Wrapf`. That remains true and is not made false by this change, so it is not a docs finding — but now that `internal/errors.Joinf` is the house pattern for sentinel-plus-cause, a pointer to it would help the next contributor choose it over `Wrapf(..., "%v", err.Error())`.

</details>

---

Proof of execution has sound shape: eight receipted claims plus an owned "Not verified" naming the external e2e workflows and why they don't apply. Every claim is CI-covered, so I opened no transcripts; I ran only the single `x/bridgingfee/types` package as the deciding evidence that the twice-revised finding is genuinely closed. On head commit `d27b9b34`, `label-prs` is green and `Analyze`, `build` and `golangci-lint` were still running — the PR body states this accurately, and those required checks gate the merge on their own, so pending CI is not a reason to hold the review. Docs remain honest: nothing in this change makes existing README or Contributing.md prose false.

VERDICT: APPROVE

omriai2 (approved):
**APPROVE** — The range since approval contains exactly one merge commit and nothing else authored on the branch; both conflicts landed in `x/sponsorship/types`, and each was resolved by combining the sibling's new sentinel taxonomy with this PR's approved `Join`/`Joinf` construction, with no approved call site lost and no edit smuggled in.

### Index

1. **Verified — merge-only range** — one merge commit `f856dbda`; every other commit in the range is an ancestor of `origin/main`. No delta re-review needed. No action.
2. **Verified — no evil merge** — the committed tree differs from a mechanical re-merge of the two parents *only* in the two genuinely conflicted files. No action.
3. **Verified — `x/sponsorship/types/msgs.go` resolution** — adopts main's `gerrc.ErrInvalidArgument` sentinel *and* this PR's `Joinf`; main's own updated tests assert that exact sentinel and pass. No action.
4. **Verified — `x/sponsorship/types/types.go` resolution** — keeps main's `errorsmod.Wrapf` call style and this PR's `Join` conversions side by side, with both imports retained. No action.
5. **Verified — approved intent intact** — all 94 `dymerrors.Join`/`Joinf` sites survive; the only deltas are the four sponsorship sentinel swaps. No action.
6. **Verified — merged tree is sound** — full-module build, `gofumpt`, `go vet`, `golangci-lint` (0 issues) and targeted tests on every affected package pass. No action.

---

<details><summary>Finding 1 — Verified: the range since approval is merge-only</summary>

`git log d27b9b34..HEAD` lists ten commits, but only one is authored on this branch. The other nine are base-branch commits that become reachable through the merge's second parent, which is exactly what a merge-in looks like. I confirmed each of them is an ancestor of `origin/main` rather than new work:

```
$ for c in $(git rev-list --no-merges d27b9b34..HEAD); do
    git merge-base --is-ancestor $c origin/main && echo "OK  (in main) $(git log -1 --format='%h %s' $c)" || echo "NOT-IN-MAIN ..."
  done
OK  (in main) 8497b49e fix(forward): support budget-relative Hyperlane amounts (#2216)
OK  (in main) df2809a1 test(rollapp): pin create-rollapp gas determinism (#2211)
OK  (in main) 62a492a3 fix(rollapp): handle missing revision heights (#2210)
OK  (in main) 0272de98 fix(sponsorship): adopt gerrc error taxonomy (#2209)
OK  (in main) 5daa024d fix(rollapp): validate fraud proposal fields (#2208)
OK  (in main) 9a13bfeb feat(ante): check circuit breaker for wrapped msgs (#2207)
OK  (in main) dc5c987f feat(dymns): bind agent service records (#2206)
OK  (in main) 03a65c67 fix(agent): bound validation responses per request (#2205)
OK  (in main) 7c3129c4 fix(forward): clamp stale IBC timeouts (#2201)

$ git rev-list --merges d27b9b34..HEAD | while read c; do git log -1 --format='%h parents=%p %s' $c; done
f856dbda parents=d27b9b34 8497b49e chore: merge main into error-chain refactor

$ git merge-base --is-ancestor origin/main HEAD && echo "origin/main fully merged"
origin/main fully merged
```

The merge's first parent is the approved head `d27b9b34` and its second is the base tip `8497b49e`, so no approved work was rebased or rewritten. Since no non-merge commit was authored on the branch, this stays a conflict-resolution review and I did not re-open the previously approved changes.

</details>

<details><summary>Finding 2 — Verified: no evil merge — resolution is confined to the two conflicted files</summary>

The strongest available check for a merge commit is to re-perform the merge mechanically and diff the result against what was actually committed. Anything the merge introduced that came from neither parent shows up there.

```
$ git merge-tree --write-tree d27b9b34 8497b49e
7bc986881e222dc61539746f253c74268d8fa2d5
...
Auto-merging app/ante/reject_msgs_test.go
Auto-merging x/agent/keeper/msg_server_validation.go
Auto-merging x/dymns/keeper/msg_server_set_service_record_test.go
Auto-merging x/rollapp/types/rollapp.go
Auto-merging x/sponsorship/types/msgs.go
CONFLICT (content): Merge conflict in x/sponsorship/types/msgs.go
Auto-merging x/sponsorship/types/msgs_test.go
Auto-merging x/sponsorship/types/types.go
CONFLICT (content): Merge conflict in x/sponsorship/types/types.go

$ git diff --stat 7bc98688 f856dbda
 x/sponsorship/types/msgs.go  | 48 ++++----------------------------------------
 x/sponsorship/types/types.go | 11 ----------
 2 files changed, 4 insertions(+), 55 deletions(-)
```

Git reports exactly two content conflicts, and the committed merge differs from the naive auto-merge in exactly those two files — the deletions are the conflict markers being replaced by the chosen resolution. Every other file in the merge, including all of main's sibling work across `app/ante`, `x/agent`, `x/forward`, `x/dymns` and `x/rollapp`, matches the mechanical merge byte for byte. Nothing was hand-edited under cover of the merge.

`git show f856dbda --cc` independently confirms this: the combined diff, which by construction only shows hunks differing from *all* parents, is 4.2 KB and contains nothing but the two sponsorship files.

</details>

<details><summary>Finding 3 — Verified: `x/sponsorship/types/msgs.go` correctly combines both sides</summary>

This is the substantive conflict. The sibling PR #2209 ("adopt gerrc error taxonomy") rewrote these validators to use `gerrc.ErrInvalidArgument` instead of `sdkerrors.ErrInvalidAddress`, while keeping the old `.Wrapf(..., "%s", err.Error())` stringification. This PR had independently rewritten the same lines to `dymerrors.Joinf(sdkerrors.ErrInvalidAddress, ...)`. The two changes overlap on every line.

The resolution takes main's *sentinel* and this PR's *construction*:

```
$ git diff 8497b49e f856dbda -- x/sponsorship/types/msgs.go
-		return gerrc.ErrInvalidArgument.Wrapf(
-			"voter '%s' must be a valid bech32 address: %s",
-			m.Voter, err.Error(),
-		)
+		return dymerrors.Joinf(gerrc.ErrInvalidArgument, err, "voter '%s' must be a valid bech32 address", m.Voter)
...
-		return errorsmod.Wrap(ErrInvalidDistribution, err.Error())
+		return dymerrors.Join(ErrInvalidDistribution, err)
```

Diffed against main, the file shows *only* this PR's approved intent — the `err.Error()` stringification replaced by chain-preserving joins — with main's gerrc taxonomy fully intact across all four validators. Neither side's change was dropped.

The import block is correspondingly correct: `dymerrors` added, main's `gerrc` retained, and both `errorsmod` and `sdkerrors` removed because the resolution leaves nothing in the file that needs them. `go vet` and `golangci-lint` confirm no unused or missing import.

The decisive semantic evidence is that main's own tests, which auto-merged without conflict, pin the sentinel the resolution had to choose:

```
$ rg -n 'errorIs|gerrc\.' x/sponsorship/types/msgs_test.go
46:			errorIs:       gerrc.ErrInvalidArgument,
59:			errorIs:       types.ErrInvalidDistribution,
72:				require.ErrorIs(t, err, tt.errorIs)
103:			errorIs:       gerrc.ErrInvalidArgument,
...
```

Had the resolution kept `sdkerrors.ErrInvalidAddress`, those `require.ErrorIs` assertions would fail. They pass, which proves the joined error's `Unwrap() []error` still reaches main's chosen sentinel.

</details>

<details><summary>Finding 4 — Verified: `x/sponsorship/types/types.go` correctly interleaves both sides</summary>

Here the two sides touched adjacent lines inside the same functions rather than the same line. Main converted the method-form `ErrInvalidVote.Wrapf(...)` calls to the package-form `errorsmod.Wrapf(ErrInvalidVote, ...)`; this PR converted the neighbouring `errorsmod.Wrap(ErrInvalidVote, err.Error())` calls to `dymerrors.Join(ErrInvalidVote, err)`. The resolution keeps both, in the same function bodies:

```go
func (v Vote) Validate() error {
	err := ValidateGaugeWeights(v.Weights)
	if err != nil {
		return dymerrors.Join(ErrInvalidVote, err)          // this PR
	}
	if !v.VotingPower.IsPositive() {
		return errorsmod.Wrapf(ErrInvalidVote, "must be > 0, got %s", v.VotingPower)  // main
	}
	return nil
}
```

Both `dymerrors` and `errorsmod` are imported, which is exactly right since both are now used. Diffed against main, the file again shows only this PR's two `Wrap(..., err.Error())` → `Join(...)` conversions.

I also checked the sentinel definitions the resolution depends on. `x/sponsorship/types/errors.go` came through from main untouched, with the sentinels upgraded from `errorsmod.Register` to `errorsmod.RegisterWithGRPCCode`. They remain registered `*errorsmod.Error` values, so `dymerrors.Join`'s `errorsmod.ABCIInfo(registered, false)` still reads the codespace and code directly off the sentinel — the sponsorship codespace and codes 1–7 are preserved. Main's new `x/sponsorship/types/errors_test.go`, which pins both the ABCI codes and the new gRPC categories (`codes.InvalidArgument` / `codes.FailedPrecondition`) and asserts the sentinels stay mutually distinct under `errors.Is`, passes on the merged tree.

</details>

<details><summary>Finding 5 — Verified: every approved call site survived the merge</summary>

Beyond the conflicted files, two files were touched by *both* sides and auto-merged silently — `x/rollapp/types/rollapp.go` and `x/agent/keeper/msg_server_validation.go`. A clean auto-merge can still lose an edit, so I enumerated every `dymerrors` call site at the approved head and at the merged head and diffed the lists:

```
$ git grep -n 'dymerrors\.\(Join\|Joinf\)(' d27b9b34 -- '*.go' | sed 's/^d27b9b34://' | sort > approved.txt
$ git grep -n 'dymerrors\.\(Join\|Joinf\)(' HEAD -- '*.go' | sed 's/^HEAD://' | sort > merged.txt
approved count: 94
merged   count: 94

$ diff approved.txt merged.txt
87c87
< x/sponsorship/types/msgs.go:20: ... Joinf(sdkerrors.ErrInvalidAddress, ...)
---
> x/sponsorship/types/msgs.go:20: ... Joinf(gerrc.ErrInvalidArgument, ...)
89,90c89,90   (same swap, MsgRevokeVote and MsgUpdateParams)
92,94c92,94   (same swap for MsgClaimRewards, plus two line-number shifts in types.go)
```

94 call sites before, 94 after. The only content differences are the four intentional sponsorship sentinel swaps; the two `types.go` entries differ only in line number, shifted by the added import. Nothing was lost anywhere, including in the two silently auto-merged files:

```
$ rg -n 'dymerrors' x/rollapp/types/rollapp.go x/agent/keeper/msg_server_validation.go
x/agent/keeper/msg_server_validation.go:6:	dymerrors "github.com/dymensionxyz/dymension/v3/internal/errors"
x/agent/keeper/msg_server_validation.go:65:			return nil, dymerrors.Join(types.ErrValidationFeePayment, err)
x/agent/keeper/msg_server_validation.go:68:			return nil, dymerrors.Join(types.ErrValidationFeePayment, err)
x/rollapp/types/rollapp.go:10:	dymerrors "github.com/dymensionxyz/dymension/v3/internal/errors"
x/rollapp/types/rollapp.go:99:		return dymerrors.Join(ErrInvalidInitialSequencer, err)
```

The `internal/errors` helper itself, the bridging-fee fix, and the committed regression tests from the three approved commits are all untouched by the merge.

</details>

<details><summary>Finding 6 — Verified: the merged tree builds, formats, lints and tests clean</summary>

A textually correct merge can still be semantically broken, so I exercised the merged tree. I deliberately scoped execution to the packages the merge actually decided plus a whole-module build, rather than re-running the full suite that CI owns.

```
$ gofumpt -l x/sponsorship/types/msgs.go x/sponsorship/types/types.go internal/errors/join.go
(no formatting diffs)

$ go vet ./x/sponsorship/... ./internal/errors/...
vet ok

$ go build ./...
build exit=0

$ golangci-lint run ./x/sponsorship/... ./internal/errors/...
0 issues.
golangci-lint exit=0

$ go test ./x/sponsorship/... ./internal/errors/... -count=1
ok  	github.com/dymensionxyz/dymension/v3/x/sponsorship/client/cli	0.111s
ok  	github.com/dymensionxyz/dymension/v3/x/sponsorship/keeper	2.404s
ok  	github.com/dymensionxyz/dymension/v3/x/sponsorship/types	0.168s
ok  	github.com/dymensionxyz/dymension/v3/internal/errors	0.005s

$ go test ./x/rollapp/types/... ./x/agent/... -count=1
ok  	github.com/dymensionxyz/dymension/v3/x/rollapp/types	0.098s
ok  	github.com/dymensionxyz/dymension/v3/x/agent	0.079s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	0.084s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	1.881s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.098s
```

The whole-module build passing matters most here: it is the check that catches a merge dropping an import, a signature, or a symbol that a sibling change introduced elsewhere. The sponsorship suites cover the conflict resolution itself, including main's newly added `errors_test.go` and `query_server_test.go`; the rollapp and agent suites cover the two-sided auto-merges.

On the merged head `f856dbda`, `label-prs` is green and `Analyze`, `build` and `golangci-lint` were still queued at review time. Those are the required checks and gate the merge on their own; the PR is `MERGEABLE` with `mergeStateStatus: BLOCKED`, which reflects the pending checks rather than an unresolved conflict.

</details>

---

No findings require action. The merge is structurally clean, both conflict resolutions preserve the sibling PR's error-taxonomy change and this PR's approved chain-preserving construction without dropping either, and the merged tree builds, lints and tests green across every package the merge touched.

VERDICT: APPROVE

PR #2216, opened 2026-09-05
fix(forward): support budget-relative Hyperlane amounts

### LINKED ISSUE #2215: x/forward: budget-relative amount for Hyperlane forwards (autonomous cross-chain intents strand funds when the post-eIBC-fee budget ≠ the composer's fixed Hyperlane amount)
## Problem / opportunity

`x/forward` composes RollApp→Hub→remote-chain flows. When the final leg is a **Hyperlane** transfer, the amount to send is a **fixed value baked in by the composer** at intent-creation time — but the two other forward legs are **budget-relative** (they forward whatever actually arrives). This asymmetry strands funds for autonomous, agent-composed cross-chain intents.

Concretely, in `x/forward/hyperlane.go` `forwardToHyperlane`:

```go
maxCost := d.HyperlaneTransfer.MaxFee.Amount.Add(d.HyperlaneTransfer.Amount)
if maxCost.GT(budget.Amount) {
    return gerrc.ErrInvalidArgument.Wrapf("max cost (fee + amount)exceeds max budget %s > %s", maxCost, budget.Amount)
}
m := &warptypes.MsgRemoteTransfer{ ... Amount: d.HyperlaneTransfer.Amount, MaxFee: d.HyperlaneTransfer.MaxFee, ... }
```

The outbound `Amount` is the composer's fixed `d.HyperlaneTransfer.Amount`. Contrast the sibling IBC leg, `x/forward/ibc.go` `forwardToIBC`, which forwards the **entire arriving budget**:

```go
m := ibctransfertypes.NewMsgTransfer(..., maxBudget /* the actual arriving coin */, fundsSrc.String(), transfer.Receiver, ...)
```

Why this bites autonomous flows specifically: on the RollApp→Hub→HL path the completion hook (`rollToHLHook.Run`) runs **after** the eIBC fulfiller has already deducted its fee, so the `budget` reaching the hook is `original_amount − eibc_fee`. That fee is set at fulfillment time (Dutch-auction / on-demand-LP dynamics — see #2167, #2182, #2189) and is **not knowable** when the composer builds the memo. The same applies to the HL-to-HL inbound path (`OnHyperlaneMessage` → `forwardToHyperlane`), where the arriving `budget` is the credited warp amount. So:

1. **Undershoot → hard revert + stranded funds.** If `Amount + MaxFee > budget` (the common case once any fee is taken), the forward reverts and the funds sit at the intermediate recipient (`args.Account` / `fundsSource`) — a key-derived/ephemeral address. The code comments say *"relying on frontend here"* for recovery; an autonomous agent has **no frontend**, so the funds are effectively stuck.
2. **Overshoot → silent dust leak.** If the composer under-estimates the fee and `Amount + MaxFee < budget`, the surplus `budget − Amount − MaxFee` is left behind at the intermediate address and never forwarded.

The IBC leg has neither problem because it is budget-relative. The Hyperlane leg should offer the same intent semantics: *"forward whatever arrives, minus the relayer fee."*

## Why now

Cross-chain **intents** — a signed declaration of a desired outcome, fulfilled by a competing solver network with atomic settlement — are the dominant 2026 pattern for agent-driven value movement (intent-solver networks logged \$4.1B cross-chain volume over 90 days; NEAR Intents reports \$24B+ across 35+ chains). Dymension's eIBC + forward stack **is** an intent+solver system: the agent signs an outcome ("land these funds on domain X via Hyperlane"), the eIBC fulfiller is the competing solver, and forward is the settlement router. For that to work autonomously, the settlement leg must be robust to the solver's fee being unknown up front — exactly the budget-relative property the IBC leg already has and the Hyperlane leg lacks.

Sources:
- Cross-chain intents / solver networks (2026): https://eco.com/support/en/articles/15264361-what-are-cross-chain-intents-2026-guide-to-intent-based-routing
- Crypto AI agents & on-chain settlement (2026): https://coincub.com/blog/crypto-ai-agents/
- x402 agentic payment rails (Coinbase): https://www.coinbase.com/developer-platform/discover/launches/x402

## Proposed change

Make the Hyperlane forward leg optionally **budget-relative**, mirroring the IBC leg, with an explicit slippage floor so the agent bounds acceptable erosion.

**Proto** — `proto/dymensionxyz/dymension/forward/dt.proto`, extend `HookForwardToHL` (additive, backward-compatible):

```proto
message HookForwardToHL {
  hyperlane.warp.v1.MsgRemoteTransfer hyperlane_transfer = 1;

  // When true, ignore hyperlane_transfer.amount and forward (budget - max_fee)
  // i.e. "send everything that arrived, reserving max_fee for interchain gas".
  bool use_full_budget = 2;

  // Slippage floor for use_full_budget mode: if the resolved send amount is
  // below this, revert instead of forwarding a dust amount. Zero = no floor.
  string min_amount = 3 [
    (gogoproto.nullable) = false
  ];
}
```

Then `make proto-gen` to regenerate `x/forward/types/dt.pb.go`.

**Pure resolver (unit-testable)** — add to `x/forward/types/to_hl.go`:

```go
// ResolveHLForwardAmount decides how many base units to forward on the
// Hyperlane leg given the funds that actually arrived (budget).
func ResolveHLForwardAmount(budget sdk.Coin, d *HookForwardToHL) (math.Int, error) {
    mt := d.HyperlaneTransfer
    if mt.MaxFee.Denom != budget.Denom {
        return math.Int{}, gerrc.ErrInvalidArgument.Wrapf("max fee denom != budget denom: %s != %s", mt.MaxFee.Denom, budget.Denom)
    }
    if !d.UseFullBudget {
        // legacy behavior: fixed amount, must fit under budget with the fee
        if mt.MaxFee.Amount.Add(mt.Amount).GT(budget.Amount) {
            return math.Int{}, gerrc.ErrInvalidArgument.Wrapf("max cost (fee + amount) exceeds budget %s > %s", mt.MaxFee.Amount.Add(mt.Amount), budget.Amount)
        }
        return mt.Amount, nil
    }
    send := budget.Amount.Sub(mt.MaxFee.Amount)
    if !send.IsPositive() {
        return math.Int{}, gerrc.ErrInvalidArgument.Wrapf("budget %s does not cover max fee %s", budget.Amount, mt.MaxFee.Amount)
    }
    if d.MinAmount.IsPositive() && send.LT(d.MinAmount) {
        return math.Int{}, gerrc.ErrInvalidArgument.Wrapf("resolved send %s below min_amount %s", send, d.MinAmount)
    }
    return send, nil
}
```

**Wiring** — in `x/forward/hyperlane.go` `forwardToHyperlane`, replace the inline `maxCost` check and the fixed `Amount:` with the resolver:

```go
if token.OriginDenom != budget.Denom { /* unchanged */ }
sendAmt, err := types.ResolveHLForwardAmount(budget, &d)
if err != nil {
    return errorsmod.Wrap(err, "resolve hl forward amount")
}
m := &warptypes.MsgRemoteTransfer{ ... Amount: sendAmt, MaxFee: d.HyperlaneTransfer.MaxFee, ... }
```

Because `forwardToHyperlane` is the single implementation used by **both** call sites — `rollToHLHook.Run` (RollApp→Hub→HL completion hook, `x/forward/ibc.go`) and `OnHyperlaneMessage` (HL→HL inbound, `x/forward/hyperlane.go`) — both paths get the fix in one place.

**Constructor / CLI** — extend `NewHookForwardToHL` (`to_hl.go`) with `useFullBudget bool, minAmount math.Int` params and thread them through the memo/metadata builders (`MakeRolForwardToHLMemoString`, `MakeIBCForwardToHLMemoString`, `MakeHLForwardToHLMetadata`) and the `x/forward/cli` decode/display helpers so `decode-hl` prints the new fields.

- **Two new fields, not a mode enum:** `use_full_budget bool` (field 2) + `min_amount` `math.Int` (field 3). A bool is the minimal additive change and defaults to the exact legacy behavior. Chosen over an enum because there are only two modes.
- **Full-budget formula = `budget − MaxFee`,** not `budget`. `MaxFee` is denominated in the same token as the transfer (the existing code already enforces `MaxFee.Denom == budget.Denom`), and Hyperlane's `MsgRemoteTransfer` needs `MaxFee` reserved for interchain gas payment; forwarding `budget − MaxFee` sends everything else. Sending the raw `budget` would double-spend the fee reserve and fail downstream.
- **`min_amount` semantics:** zero (default) = no floor. Positive = revert if `budget − MaxFee < min_amount`. This lets an agent say "if too little survives the solver fee, don't forward — leave it recoverable at the intermediate address rather than burning gas landing dust on a remote chain." Naming `min_amount` matches the eIBC/IRO slippage-floor convention (`minIncomeAmt`, `MinAmountOut`).
- **Backward compatibility:** old encoded `HookForwardToHL` payloads decode with `use_full_budget=false` and `min_amount=0`, so `ResolveHLForwardAmount` returns exactly the legacy fixed-amount path with the identical error. No migration needed; no behavior change for existing composers.
- **Scope of budget-relative behavior:** applies to both Hyperlane forward entry paths. The HL→IBC path is unchanged (already budget-relative via `forwardToIBC`). The Kaspa field is untouched.
- **Surplus is eliminated, not swept to a new account:** in full-budget mode nothing is left behind (the whole `budget − MaxFee` is forwarded), so no new sweep/refund address or escrow is introduced.

## Consensus safety

Pure integer arithmetic (`math.Int` add/sub/compare) over values already in consensus state (`budget`, `MaxFee`, `min_amount` from the decoded hook). No wall-clock, no randomness, no map iteration, no floating point. Errors are returned normally and are absorbed by the existing `executeAtomicWithErrEvent` wrapper (atomic — a failed forward rolls back its sub-state and emits the error event, exactly as today); no new panics, nothing added to Begin/EndBlocker. Identical result on every validator.

## Acceptance criteria

- `proto/.../forward/dt.proto` gains `use_full_budget` and `min_amount`; `make proto-gen` regenerates `x/forward/types/dt.pb.go` with no other diff, and `git status --porcelain` is clean afterward.
- New unit tests in `x/forward/types` for `ResolveHLForwardAmount` covering: (a) `use_full_budget=false` returns the wrapped `Amount` and still rejects `Amount+MaxFee > budget` with the legacy error; (b) `use_full_budget=true` returns `budget − MaxFee`; (c) `budget ≤ MaxFee` → error; (d) `min_amount` violated → error; (e) `min_amount` satisfied → ok; (f) `MaxFee.Denom != budget.Denom` → error.
- `forwardToHyperlane` uses `ResolveHLForwardAmount`; both call sites (`rollToHLHook.Run`, `OnHyperlaneMessage`) exercise the new path. A regression test (unit or `ibctesting`) shows a RollApp→Hub→HL forward with `use_full_budget=true` succeeding when an eIBC fee has reduced the arriving budget below the composer's original amount (the case that reverts today).
- Backward-compat test: a `HookForwardToHL` with the new fields unset produces byte-identical `MsgRemoteTransfer.Amount` to the pre-change behavior.
- `go build ./...`, `go test ./x/forward/...`, `golangci-lint run` on changed files, and `gofumpt -l` all clean.

## Out of scope

- The IBC forward leg (`forwardToIBC`) and the Kaspa metadata field — do not touch.
- eIBC fee computation, bridging-fee logic (#2173), and Dutch-auction/on-demand-LP dynamics (#2167/#2182/#2189).
- IBC/HL timeout handling — covered separately by #2200.
- Any change to `hyperlane-cosmos` `MsgRemoteTransfer` semantics or the warp keeper.
- Automatic recovery/sweep of funds already stranded by past reverts (this change prevents new strands; it does not migrate old ones).

---

omriai1:
Closes #2215

Hyperlane forwards can send the arriving budget minus `MaxFee`, with an optional `min_amount` floor, while valid legacy hooks keep their fixed-amount behavior. Completion hooks now receive a budget net of the bridging fee when a RollApp packet arrives already finalized, preventing full-budget forwards from spending more than was credited. The CLI carries the new fields and makes inactive floors explicit.

<details><summary>Implementation, budget provenance, and compatibility</summary>

- Added protobuf fields `use_full_budget = 2` and `min_amount = 3`; generated changes are confined to `x/forward/types/dt.pb.go`.
- Both the RollApp completion hook and Hyperlane inbound hook use `ResolveHLForwardAmount` through `forwardToHyperlane`. Full-budget mode reserves `MaxFee`, requires a positive remainder, and enforces a positive floor; an omitted floor means zero.
- All completion-hook budget producers follow the net-credit contract documented at `RunCompletionHook`: eIBC fulfillment supplies the order price, delayed finalization subtracts the bridging fee, and non-delayed IBC completion now subtracts it only for RollApp transfers. Hyperlane inbound forwards use the credited warp coin. Existing recipient balances are excluded.
- Bridging-fee calculation/charging, the IBC forwarding implementation, Kaspa metadata, and the warp keeper are unchanged.
- Valid legacy amounts and budget/denomination errors are preserved. Both modes reject nil/negative fees before arithmetic, and legacy mode also rejects a missing fixed amount.
- Positive CLI `--min-amount` requires `--use-full-budget`. Decoding older legacy metadata labels its floor inactive, including when a positive floor was encoded by an older composer.

</details>

## Proof of execution

- `TestForwardSuite/TestIBCFullBudgetHyperlane`: PASS in finalized-RollApp and non-RollApp modes, each with direct and eIBC memos; real IBC/bank/warp keepers commit dispatch and net collateral while preserving 77 pre-existing units. Required CI **Build and Test** re-runs this regression.
- `go test ./ibctesting -run '^(TestForwardSuite|TestOsmosisForwardSuite)$' -count=1`: PASS; covers the new real-stack cases plus the existing fulfillment, delayed-finalization, and non-RollApp completion cases. Required CI **Build and Test** re-runs both suites; no shared fixtures/helpers were changed.
- `go test ./x/forward/... ./x/ibc_completion/...`: PASS; resolver/legacy wire compatibility, both forward entry handlers, memo round trips, malformed legacy amounts, positive inactive-floor rejection, and inactive-floor display. Required CI **Build and Test** re-runs these tests.
- `GOTOOLCHAIN=go1.24.6 golangci-lint-v2 run ./x/forward/... ./x/ibc_completion/... ./x/delayedack/keeper ./ibctesting`: PASS, `0 issues`; same configuration as required CI **golangci-lint**.
- Initial full gate at `235d0006`: `go build ./...`, full tests (65 packages passed), CI-style `go-acc` race/coverage (65 packages passed), and repository-wide lint all passed. Full gates were not repeated in this scoped fix round.
- Protobuf regeneration and formatting: PASS; regeneration left the initial committed tree clean, and protobuf files are unchanged in this fix round.

<details><summary>Regeneration and current formatting evidence</summary>

```text
Initial commit 235d0006:
$ make proto-gen
(exit 0)
$ git status --porcelain
(no output; working tree clean)

Fix round:
$ gofumpt -l ibctesting/hyperlane_budget_test.go x/ibc_completion/ibc_module.go x/delayedack/keeper/completion.go x/forward
(no output; exit 0)
$ git diff --check
(no output; exit 0)
```

</details>

- Built CLI smoke: PASS; positive inactive floors are rejected and legacy decoded floors are marked inactive.

<details><summary>Refreshed CLI invocation evidence</summary>

Built with `go build -o /tmp/2216-dymd ./cmd/dymd`. Each invocation had a 30-second timeout. The decode input wraps freshly generated legacy HL metadata in a deterministic test message. Only the behavior changed by this review round was rechecked.

```text
$ /tmp/2216-dymd q forward create-memo --src=hl --dst=hl --token-id=0x0101010101010101010101010101010101010101010101010101010101010101 --dst-domain=1 --funds-recipient-dst=0x0202020202020202020202020202020202020202020202020202020202020202 --max-fee=10adym --amount=100 --min-amount=80
Error: min-amount requires --use-full-budget
Usage:
  dymd query forward create-memo [flags]

Examples:
# IBC to Hyperlane
dymd q forward create-memo --src=ibc --dst=hl \
  --token-id=0x934b867052ca9c65e33362112f35fb548f8732c2fe45f07b9c591958e865def0 \
  --dst-domain=1 \
  --funds-recipient-dst=0x934b867052ca9c65e33362112f35fb548f8732c2fe45f07b9c591958e865def0 \
  --amount=10000 \
  --max-fee=20foo

# EIBC to IBC
dymd q forward create-memo --src=eibc --dst=ibc \
  --eibc-fee=100 \
  --channel=channel-0 \
  --funds-recipient-dst=ethm1a30y0h95a7p38plnv5s02lzrgcy0m0xumq0ymn \
  --timeout=5m

Flags:
      --amount string                Token amount to transfer (in smallest unit)
      --channel string               IBC channel ID (e.g., channel-0)
      --domain uint32                Domain ID (deprecated, use --dst-domain)
      --dst string                   destination protocol (ibc, hl, hub)
      --dst-amount string            Amount to be received on destination after fees
      --dst-domain uint32            Destination chain domain ID (e.g., 11155111 for Ethereum Sepolia)
      --dst-token-id string          Destination token ID in hex format
      --eibc-fee string              EIBC fee amount for fast finality (required for EIBC source, paid to order fulfiller)
      --funds-recipient-dst string   Final recipient address on destination chain (pad to 32 bytes for Ethereum: 0x000...)
  -h, --help                         help for create-memo
      --max-fee string               Maximum fee for Hyperlane execution including denom (e.g., 20ibc/ABC123...)
      --min-amount string            Minimum amount to forward (base units; positive values require --use-full-budget; zero means no floor) (default "0")
      --nonce uint32                 Message nonce for ordering/uniqueness
      --readable                     Show output in human-readable format
      --src string                   Src protocol (ibc, eibc, hl, kaspa)
      --src-contract string          Source contract address in hex format
      --src-domain uint32            Source chain domain ID (e.g., 1260813472 for Dymension Hub)
      --timeout string               IBC packet timeout duration (e.g., 5m, 1h, 30s) (default "5m")
      --token-id string              Token ID in hex format (32 bytes, e.g., 0x726f757465725f61707000000000000000000000000000010000000000000005)
      --use-full-budget              Forward the arriving budget minus max-fee, ignoring the fixed forward amount

Global Flags:
      --chain-id string     The network chain ID
      --home string         directory for config and data (default "/home/workforce/.dymension")
      --log_format string   The logging format (json|plain) (default "plain")
      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic|disabled or '*:<level>,<key>:<level>') (default "info")
      --log_no_color        Disable colored logs
      --trace               print out full stack trace on errors

min-amount requires --use-full-budget
exit: 1
$ /tmp/2216-dymd q forward create-memo --src=hl --dst=hl --token-id=0x0101010101010101010101010101010101010101010101010101010101010101 --dst-domain=1 --funds-recipient-dst=0x0202020202020202020202020202020202020202020202020202020202020202 --max-fee=10adym --amount=100
0x1aa4010a9e011242307830313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031180122423078303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230322a033130303a0130420a0a046164796d120231301a0130
exit: 0
$ /tmp/2216-dymd q forward decode-hl 0x0300000001000000010000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000781aa4010a9e011242307830313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031180122423078303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230322a033130303a0130420a0a046164796d120231301a0130
=== Hyperlane Message ===
  Version:     3
  Nonce:       1
  Origin:      1
  Sender:      0x0000000000000000000000000000000000000000000000000000000000000000
  Destination: 2
  Recipient:   0x0000000000000000000000000000000000000000000000000000000000000000
  Body:        0x000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000781aa4010a9e011242307830313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031180122423078303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230322a033130303a0130420a0a046164796d120231301a0130

=== Forwarding Memo ===
  Raw hex:    0x1aa4010a9e011242307830313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031303130313031180122423078303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230323032303230322a033130303a0130420a0a046164796d120231301a0130
  Forward to HL:  yes (164 bytes)

=== Forward to Hyperlane ===
  Token ID:           0x0101010101010101010101010101010101010101010101010101010101010101
  Destination Domain: 1
  Recipient:          0x0202020202020202020202020202020202020202020202020202020202020202
  Amount:             100
  Max Fee:            10adym
  Use Full Budget:    false
  Min Amount:         0 (inactive; requires use_full_budget)

=== Warp Payload ===
  EVM Address:    0x0000000000000000000000000000000000000000
  Kaspa (mainnet): kaspa:qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqkx9awp4e
  Kaspa (testnet): kaspatest:qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqhqrxplya
  Cosmos Account: dym1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqp7vezn
  Amount:         120
exit: 0
Fix-round CLI smoke: 3 commands passed (2 success, 1 expected rejection).
```

</details>

**Not verified:** live remote-chain settlement and the separately hosted Docker E2E/upgrade/nightly workflows in `dymensionxyz/e2e-tests`. The new IBC regression uses real local keepers with noop Hyperlane dispatch hooks (no gas charge); the earlier isolated entry-handler tests use warp keeper stubs.


omriai2 (changes requested):
REQUEST_CHANGES — the resolver, its tests, and the CLI plumbing are clean and the security review turned up nothing, but the new full-budget mode makes the chain trust `budget` as the exact credited amount, and one of the three code paths that produce `budget` reports it gross of the bridging fee, so on that path the feature deterministically over-sends and strands funds — the exact harm it exists to prevent.

**Index**

1. **Full-budget mode over-sends on the non-delayed IBC completion path, because that path reports `budget` gross of the bridging fee** — accept when the non-delayed completion path passes a budget net of the bridging fee (mirroring the correction the delayed-ack path already makes), or when the author shows with evidence that a `dym-fwd-roll-hl` hook with `use_full_budget=true` cannot reach that path with a nonzero bridging fee.

Non-blocking observations follow the finding; none of them hold the merge.

<details><summary>Finding 1 — Full-budget mode over-sends on the non-delayed IBC completion path</summary>

**What changed and why it matters**

Before this change, the amount sent on the Hyperlane leg was whatever the composer baked into the hook, and `budget` was used only as a permissive ceiling check (`MaxFee + Amount <= budget`). The composer chose the number, so an inaccurate `budget` merely made the guard slightly loose; a composer who left any headroom was fine.

After this change, in full-budget mode the chain itself derives the send amount as `budget − MaxFee` and the composer has no control over it. That converts `budget` from an advisory ceiling into a load-bearing value: it must equal exactly what was credited to the intermediate address, or the outbound warp transfer will try to move more than the account holds.

**The three producers of `budget`, and the one that is wrong**

I traced every producer that can reach the `dym-fwd-roll-hl` completion hook:

1. **eIBC fulfillment** (`x/delayedack/keeper/hooks.go:40`) passes the order price — exactly what the LP credited. Correct.
2. **Delayed-ack finalization** (`x/delayedack/keeper/completion.go:121-124`) explicitly subtracts the bridging fee, with a comment saying why:
   ```go
   amt := pTransfer.MustAmountInt()
   // account for the bridge fee which happened before the receiver got the funds
   amt = amt.Sub(k.BridgingFeeFromAmt(ctx, amt))
   return k.RunOrderCompletionHook(ctx, o, amt)
   ```
   Correct.
3. **Non-delayed IBC completion** (`x/ibc_completion/ibc_module.go:154`) passes the raw packet amount with no such correction:
   ```go
   budget := sdk.NewCoin(denom, amt)   // amt == transfer.Amount, gross
   ```

That third path is reachable and does charge a bridging fee. The delayed-ack middleware sets the "was not delayed" marker and calls straight through whenever the transfer is from a RollApp **and** its height is already finalized (`x/delayedack/ibc_middleware.go:97-99`). The bridging-fee middleware charges only `if transfer.IsRollapp()` — so the overlap "RollApp packet, not delayed" is precisely this path, and it is precisely where the fee is taken. Ordering inside `x/ibc_completion/ibc_module.go:99-104` seals it: the wrapped stack (which includes the bridging-fee middleware) runs first, crediting the receiver and then charging the fee back out of that same account, and only afterwards is the completion hook invoked with the gross figure.

**The concrete failure**

Let `G` be the packet amount, `b` the bridging fee (default 0.1% — `x/delayedack/types/params.go:28`), `F` the hook's `MaxFee`, and `f` the actual relayer/IGP charge.

- Account balance when the hook runs: `G − b`.
- `ResolveHLForwardAmount` full-budget branch (`x/forward/types/to_hl.go:151`) returns `G − F`.
- The warp collateral path does `SendCoinsFromAccountToModule(sender, module, amount)` before the dispatch fee is charged, so it needs `G − F ≤ G − b`, i.e. `F ≥ b`; and then the dispatch needs `f ≤ F − b`.

So the forward succeeds only when `MaxFee ≥ bridgingFee + actualFee`. `MaxFee` is denominated in the same token as the budget (the existing denom equality check is retained), and the bridging fee is charged in that same denom, so these are directly comparable. `MaxFee` is naturally sized to the relayer gas cost — a small absolute number — while the bridging fee scales with the transfer: on a 1,000,000-unit transfer it is 1,000 units. Any composer sizing `MaxFee` sensibly gets a guaranteed revert, and the funds sit at the intermediate address, which in the autonomous-agent flow this PR is written for is an ephemeral key with no frontend to recover them. That is the exact stranding the change is meant to eliminate.

**Why this is a finding against this PR rather than a pre-existing wart to ignore**

The gross figure in `x/ibc_completion` predates the change, and I am not asking for a general audit of it. What this PR does is make correctness of that figure mandatory: it removes the composer's ability to compensate. The codebase already demonstrates it knows the correction is required — the delayed-ack call site does it explicitly and comments on it — so the fix is a known, two-line shape rather than new design work. The committed tests cannot catch this because `hyperlane_test.go` supplies `budget` directly to the handlers rather than driving a real packet through the transfer stack, so every producer of `budget` is stubbed out.

**Class, not just the instance**

The class is: *full-budget mode is only as correct as the least accurate producer of `budget`, and there is no invariant anywhere asserting that `budget` equals the credited amount.* Today three producers exist and one is wrong; any future call site of `RunCompletionHook` inherits the same trap silently. If you fix only the one line I named, the class remains open — consider whether the intended invariant belongs somewhere it can be stated once (for example, having the completion-hook entry points derive the net figure rather than each caller remembering to).

**Acceptance criterion — any one of these**

- Make the non-delayed completion path pass a budget net of the bridging fee, mirroring `x/delayedack/keeper/completion.go:121-124`, with a test that drives a finalized-height RollApp transfer carrying a `use_full_budget` hook through the real stack and asserts the forward succeeds with a nonzero bridging fee param; **or**
- Show with evidence — a test or a traced argument over the middleware stack — that a `dym-fwd-roll-hl` hook with `use_full_budget=true` cannot reach `x/ibc_completion/ibc_module.go:154` while a bridging fee is charged, in which case I am wrong and this drops.

Documenting the gap is not sufficient here: the failure mode is silent fund stranding, not a confusing behavior.

**Transcript — the reachability chain**

```
$ sed -n '95,105p' x/delayedack/ibc_middleware.go
	if !transfer.IsRollapp() || transfer.Finalized {
		ctx = commontypes.WithWasNotDelayed(ctx)
		return w.IBCModule.OnRecvPacket(ctx, packet, relayer)
	}

$ sed -n '79,102p' x/bridgingfee/ibc_module.go
	if !transfer.IsRollapp() {
		return w.IBCModule.OnRecvPacket(ctx, packet, relayer)
	}
	// handle the packet transfer
	ack := w.IBCModule.OnRecvPacket(ctx, packet, relayer)
	...
	feeAmt := w.delayedAckKeeper.BridgingFeeFromAmt(ctx, transfer.MustAmountInt())
	...
	err = osmoutils.ApplyFuncIfNoError(ctx, func(ctx sdk.Context) error {
		return w.txFeesKeeper.ChargeFeesFromPayer(ctx, receiver, feeCoin, nil)
	})

$ sed -n '99,105p' x/ibc_completion/ibc_module.go
	ack := m.IBCModule.OnRecvPacket(ctx, packet, relayer)   // <- fee charged in here
	if !ack.Success() {
		return ack
	}
	err = m.dackK.RunCompletionHook(ctx, h.fundsSrc, h.budget, h.hook)   // <- h.budget is gross
```

</details>

<details><summary>Non-blocking observations (no action required to merge)</summary>

These are judgement calls I would not hold a merge for.

**Asymmetric robustness between the two resolver branches.** The full-budget branch now explicitly rejects a nil or negative `MaxFee.Amount` before subtracting, and there is a dedicated test for it. The legacy branch still computes `MaxFee.Amount.Add(mt.Amount)` with no such guard, so a crafted payload omitting either amount string dereferences a nil big.Int. That line is byte-identical to the pre-change code and the panic is absorbed by the atomic wrapper's recovery, so nothing new is broken — but since you were already writing the guard for one branch, hoisting it above the branch would close it for free.

**`min_amount` is accepted and encoded when full-budget mode is off.** The CLI happily takes `--min-amount=500` without `--use-full-budget` and writes it into the hook, where the chain silently ignores it. A composer could reasonably read that as a floor being enforced. Rejecting the combination, or noting the dependency in the flag help, would remove the ambiguity.

**`decode-hl` prints `Min Amount` unconditionally.** Same shape as above on the display side — the floor is shown for legacy hooks where it has no effect.

**Proof and coverage checked, no defects found.** The Proof of execution section is well-formed: every claim carries a named test or command plus a receipt, the CLI smoke includes raw invocation transcripts with expected-rejection cases, proto regeneration shows a clean tree, and there is an owned "Not verified" entry for live cross-chain settlement and the externally hosted E2E workflows. The test claims are covered by the required CI suite, so I did not open their transcripts. No external dependency call is added or re-pointed. Regression coverage is committed and real — both entry handlers are exercised, not just the pure resolver — and I ran the narrow package locally to confirm it is green:

```
$ go test ./x/forward/...
ok  	github.com/dymensionxyz/dymension/v3/x/forward	0.113s
ok  	github.com/dymensionxyz/dymension/v3/x/forward/cli	0.095s
ok  	github.com/dymensionxyz/dymension/v3/x/forward/cli/kaspaaddr	0.008s
ok  	github.com/dymensionxyz/dymension/v3/x/forward/types	0.057s
```

A security review of the diff found no HIGH or MEDIUM findings. The spending ceiling is unchanged in both modes (`send + MaxFee == budget` exactly in full-budget mode), wire-level values are bounded by `math.Int`'s 256-bit unmarshal limit, legacy payloads decode to the old path byte-for-byte, and the resolver is pure deterministic integer arithmetic with no map iteration, floats, or host state. Docs are honest: there is no README or doc in the module that this makes stale, the repo's CHANGELOG has not tracked feature PRs since 2024, and the proto comments and CLI help text accurately describe the implemented behavior.

</details>

VERDICT: REQUEST_CHANGES

omriai2 (approved):
APPROVE — the one blocking finding from the prior round is fixed at the right place with a real end-to-end regression test, and I confirmed by removing the fix that the test genuinely catches the stranding it was written for; all three non-blocking observations were also addressed, and this round introduced nothing blocking.

**Index — disposition of every prior finding**

1. **Full-budget mode over-sends on the non-delayed IBC completion path (was blocking)** — **RESOLVED**. The completion path now subtracts the bridging fee for RollApp transfers using the same function and the same guard as the middleware that charges it, the invariant is documented at the single chokepoint both callers share, and a new `ibctesting` case drives a real packet through the real stack and asserts the dispatch commits with exact net collateral. Verified: passes with the fix, fails without it.
2. **Legacy resolver branch lacked the nil/negative guards the full-budget branch had (was non-blocking)** — **RESOLVED**. The `MaxFee` guard was hoisted above the branch and a nil transfer-amount guard was added, turning two recovered panics into clean errors without changing behavior for well-formed legacy payloads.
3. **`--min-amount` silently ignored when full-budget mode is off (was non-blocking)** — **RESOLVED**. A positive floor without `--use-full-budget` is now a hard CLI error, with a test case and a refreshed help string.
4. **`decode-hl` printed the floor as if it were active on legacy hooks (was non-blocking)** — **RESOLVED**. It now renders as inactive with the reason, covered by a test.

No new blocking issues. Two small observations are folded below; neither holds the merge.

<details><summary>Finding 1 — Blocking finding resolved: bridging-fee-aware budget on the non-delayed completion path</summary>

**What the fix does**

The completion path now computes the budget net of the fee that the wrapped stack has already taken:

```go
// The wrapped transfer stack charges RollApp bridging fees before the hook.
// Non-RollApp transfers do not pay this fee.
if transfer.IsRollapp() {
	amt = amt.Sub(m.dackK.BridgingFeeFromAmt(ctx, amt))
}
budget := sdk.NewCoin(denom, amt)
```

I checked that this mirrors the charge exactly rather than approximating it. The bridging-fee middleware charges `BridgingFeeFromAmt(ctx, transfer.MustAmountInt())` under the guard `if transfer.IsRollapp()`, in the denom returned by the same incoming-denom helper. The correction calls the same function, on the same amount, under the same guard, in the same denom, inside the same transaction — so the parameter cannot shift between the two evaluations. The `IsRollapp()` determination is also made on identical inputs: for an on-receive packet the hub port/channel helper resolves to the destination port and channel, which is exactly what the middleware passes. Nothing else in the stack between the two deducts: packet-forward memos are rejected earlier as conflicting middleware, and rate limiting takes no fee.

The one residual asymmetry is benign. The middleware wraps its charge in the tolerate-failure helper and continues if it fails, so in that rare case the budget under-reports by the fee. Under-reporting forwards slightly less than is available and leaves dust — it cannot over-send or strand.

**Class-level answer**

I asked whether the invariant belongs somewhere it can be stated once, since any future caller inherits the trap. The author put a contract on the shared entry point:

```go
// RunCompletionHook invokes a hook with the amount credited by the triggering transfer,
// net of fees already charged. Callers must supply that net budget, excluding
// any pre-existing recipient balance: ...
```

That is a documented contract rather than an enforced one, but for a keeper API with a single chokepoint and both callers now compliant, it is a reasonable answer, and the "excluding any pre-existing recipient balance" clause is the part a future caller most needs to read.

**Evidence — the regression test is load-bearing, not decorative**

The new case sets the bridging fee to 0.1%, sends 1,000,000 units with a `MaxFee` of 10 — deliberately two orders of magnitude below the 1,000-unit bridging fee, which is precisely the configuration I argued would revert — pre-funds the recipient with 77 units to prove the budget does not reach into a pre-existing balance, and asserts that a Hyperlane message was actually dispatched and that the collateral moved equals `net − fee` with exactly `77 + fee` left behind. It runs four ways: finalized-RollApp and non-RollApp, each with a direct and an eIBC-shaped memo.

With the fix in place:

```
--- PASS: TestForwardSuite/TestIBCFullBudgetHyperlane (1.96s)
    --- PASS: .../finalized_rollapp/direct (0.43s)
    --- PASS: .../finalized_rollapp/eibc   (0.35s)
    --- PASS: .../non_rollapp/direct       (0.46s)
    --- PASS: .../non_rollapp/eibc         (0.40s)
```

I then removed only the bridging-fee correction and re-ran the same test to confirm it actually catches the defect rather than passing vacuously:

```
--- FAIL: TestForwardSuite/TestIBCFullBudgetHyperlane/finalized_rollapp/direct
        Error: Not equal: expected: 0x1  actual: 0x0     (mailbox MessageSent)
--- FAIL: TestForwardSuite/TestIBCFullBudgetHyperlane/finalized_rollapp/eibc
        Error: Not equal: expected: 0x1  actual: 0x0
```

`MessageSent == 0` means no Hyperlane message was ever dispatched: the forward reverted and the funds stayed at the intermediate recipient — the exact stranding described in the prior round. The two non-RollApp sub-cases still passed with the fix removed, which independently confirms the `IsRollapp()` guard is scoped correctly and the correction is not being applied where no fee is charged. The working tree was restored and verified clean afterwards.

Supporting unit runs, all green: `x/forward/...`, `x/ibc_completion/...`, `x/delayedack/...`. The whole app compiles as part of the `ibctesting` run, which is what confirms the real delayed-ack keeper satisfies the newly widened interface.

</details>

<details><summary>Observations — not blocking</summary>

**The documented budget contract omits one of its three callers by name.** The comment enumerates "the eIBC order price on fulfillment, or the received packet amount minus the RollApp bridging fee on IBC completion." The delayed-ack finalization caller computes exactly the second of those two forms, so the contract covers it correctly in substance, but a reader matching call sites to clauses will not find it named. Purely prose.

**The invariant is documented, not enforced.** A future caller of the shared completion entry point can still hand it a gross figure and silently reopen the class. I would not hold the merge for this — a documented contract at the one chokepoint, with both existing callers corrected and an end-to-end test pinning the behavior, is a proportionate answer for a Go keeper API — but it is worth remembering if a fourth producer of a budget ever appears.

**Proof and CI.** The Proof of execution section is well-formed: every claim names a test or command with a receipt, the CI-covered suites are named rather than transcript-dumped, the CLI smoke carries raw transcripts including the new expected rejection, protobuf regeneration is shown clean, and there is an owned "Not verified" entry for live remote-chain settlement and the externally hosted E2E workflows. No external dependency call is added or re-pointed. At review time `buf-break-check` and `label-prs` had passed and `build`, `golangci-lint`, and `Analyze` were still queued — those are the merge gate and will speak for themselves. Docs stayed honest: the module has no README this invalidates, the repository CHANGELOG has not tracked feature PRs for years, and the new proto comments, CLI help text, and keeper contract comment all match what the code does.

</details>

VERDICT: APPROVE

PR #2220, opened 2026-09-13
feat(agent): attest validation verdicts with enclave proofs

### LINKED ISSUE #2219: x/agent: attested validation responses — bind the ERC-8004 verdict to the validator's TEE enclave (the verdict is currently only owner-signed)
## Problem / opportunity

The Validation Registry (#2198) lets an independent validator agent answer a validation request about another agent's attested action. But **the verdict itself is not attested**: `RespondValidation` (`x/agent/keeper/msg_server_validation.go:82`) authorizes the response solely by `msg.Responder == validator.Owner` — the validator's **owner key**, not its TEE enclave. So the registry proves "the validator agent exists, is live, and its owner signed a verdict," but not "the validator's attested runtime actually produced this verdict." A validator owner can rubber-stamp verdicts without ever running the validation in the enclave, which undercuts the point of a verifiable agent runtime for the third (validation) registry.

Every other consequential agent output *is* enclave-attested through the same nonce + verifier machinery: `SubmitAttestedAction` and `SubmitAttestedTransfer` (`x/agent/keeper/msg_server.go:120`) bind a payload to the agent's policy via a domain-separated nonce (`x/agent/types/nonce.go`) and append to the append-only action log. The validation verdict is the one high-value output that skips this.

## Why now

2026's verifiable-AI stack (zkML / TEE-attested inference) is converging on "don't trust, verify" for agent outputs, and ERC-8004's validation registry is explicitly the place to anchor such proofs. Dymension already ships the TEE attestation seam (`tee.Verifier`, consensus-critical since #2160); wiring it into the verdict makes the validation registry trust-minimized instead of trust-the-owner-key.

Sources:
- Verifiable inference / zkML in production 2026 (Lagrange DeepProve, EigenAI): https://blockeden.xyz/blog/2026/02/19/eigenai-end-to-end-inference-decentralized-ai-determinism/ , https://everstake.one/resources/blog/verifiable-ai-onchain-trust-layer
- zkAgent: verifiable LLM agent execution (2026): https://eprint.iacr.org/2026/199.pdf

## Proposed change

- **New message `MsgRespondValidationAttested`** (`tx.proto`) alongside `MsgRespondValidation`: same fields (`responder`, `request_hash`, `response`, `response_uri`, `response_hash`, `tag`) plus `bytes token` (the attestation token, exactly like `MsgSubmitAttestedTransfer.Token`). Signer stays `responder` (must equal `validator.Owner`, as today).
- **Canonical payload helper** (`x/agent/types/`, mirroring `AttestedTransferBytes`): `AttestedValidationBytes(requestHash []byte, response uint32, responseHash []byte, responseUri, tag string) []byte` — NUL-separated / fixed-width fields, deterministic.
- **New nonce domain** (`x/agent/types/nonce.go`): `nonceDomainValidation byte = 0x03` + `ValidationNonce(validatorID string, payload []byte, actionSeq uint64) string` reusing `attestNonce`. Keeps the validator's validation attestations in a nonce space disjoint from its own actions/transfers.
- **keeper** (`x/agent/keeper/msg_server_validation.go`): `RespondValidationAttested` performs every check `RespondValidation` does (request exists; validator & subject agent found/active/live; self-validation guard; `ValidationMaxResponsesPerRequest`; `responder == validator.Owner`; tag/uri byte bounds) AND, before recording: loads the validator via `loadAttestingAgent(validatorID)` (this also applies the #2179 revocation denylist), rejects a payload exceeding `MaxActionBytes`, derives `ValidationNonce` over the payload at the **validator's** `ActionSeq`, calls `k.verifier.Verify(ctx, validator.Policy, nonce, msg.Token)`, records the `ValidationResponse` with `Attested = true`, and appends the payload to the **validator's** attested action log via `appendAttested` (advancing the validator's `ActionSeq` → structural replay protection, exactly like transfers).
- **proto** (`validation.proto`): add `bool attested = 9` to `ValidationResponse`. Owner-signed responses set it `false`.
- **events**: add the attested flag to `EventRespondValidation`.
- **query/CLI**: surface `attested` in query responses; add a `respond-validation-attested` CLI command.
- Run `make proto-gen`.

- **Keep the plain owner-signed `RespondValidation`** (a validator that re-executes off-chain and reports still works); add attested responses as a **superset** so consumers can require `attested == true` when they need enclave provenance. Chosen over mandatory attestation to avoid breaking the existing flow and to allow non-TEE validation methods.
- **The verdict is appended to the validator's action log** (not the subject agent's), at the validator's `ActionSeq` — the verdict is the validator's own attested action. Replay protection is structural (seq advances), consistent with transfers.
- **The nonce binds the full verdict payload** (`request_hash` + `response` code + `response_hash` + `response_uri` + `tag`) so the enclave attests to the exact verdict, not an opaque blob — same rationale as `AttestedTransferBytes`. `response_hash` stays the caller-provided content hash of the off-chain evidence; the nonce commits to it, so the enclave vouches for that hash.
- **`MaxActionBytes` bounds the appended payload** (reuse the existing param) — no new param.
- **The #2179 denylist already applies** through `loadAttestingAgent -> fingerprintNotRevoked`, so a revoked validator enclave measurement cannot post attested verdicts — no extra wiring.

## Acceptance criteria

- `msg_server_validation_test.go`: (a) an attested response with a valid token (fake verifier) records `Attested=true` and advances the validator's `ActionSeq`; (b) an invalid token rolls back with no state change and no seq advance; (c) a token minted for a transfer or action nonce does NOT verify for a validation nonce (domain separation); (d) a revoked validator policy is rejected; (e) a payload over `MaxActionBytes` is rejected; (f) existing `RespondValidation` tests still pass with `Attested=false`; (g) `ValidationMaxResponsesPerRequest` is still honored on the attested path.
- Genesis round-trips `ValidationResponse.attested`.
- `go test ./x/agent/...`, `golangci-lint run`, `gofumpt -l .` clean, `make proto-gen` leaves no uncommitted diff.

## Out of scope

- No on-chain zkSNARK/zkML verification — the verifier stays the existing TEE attestation seam.
- No change to the reputation registry or to how validation results are consumed elsewhere.
- No mandatory-attestation enforcement and no removal of the owner-signed path.

## Consensus-safety

Reuses the existing deterministic `tee.Verifier` (already consensus-critical since #2160), plus pure byte/hash derivation and seq arithmetic; failures return errors (atomic rollback), never panic; no randomness, wall-clock, or map iteration; nothing runs in Begin/EndBlocker.

---

omriai1:
Enclave-attested validation verdicts are ready for re-review, with all four review findings addressed in `cb17f201`. Closes #2219.

- Token representation and JSON compatibility
- Action-log correlation
- Trust-model documentation
- Subject and evidence binding
- Proof of execution and verification limits

<details><summary>Token representation and JSON compatibility</summary>

`MsgRespondValidationAttested.token` is now `string`, matching both sibling attested messages and the verifier interface. The CLI and keeper pass it through directly, and `ValidateBasic` rejects the empty string. The proto states this rule for all attested messages; a permanent regression test decodes and re-encodes a JWT-shaped string through all three types.

Before the fix, that regression passed for actions/transfers and failed for validation with `illegal base64 data at input byte 6`. The generated transaction now contains the raw token, as shown in the current proof below.

</details>

<details><summary>Action-log correlation</summary>

The attested response now returns `action_seq` alongside the per-request verdict index `seq`. `EventRespondValidation` emits the same consumed validator log sequence; it is meaningful when `attested=true` (plain responses set it to zero). Proto comments distinguish both sequence meanings. Tests deliberately use response index 0 and consumed validator action sequence 3 and assert the event correlation, including a mixed plain/attested request.

Both response paths retain the existing request, owner, active/live-agent, self-validation, bounds, and cap checks. The attested path loads the effective policy, verifies the nonce token, appends the validator log, and stores the verdict atomically.

</details>

<details><summary>Trust-model documentation</summary>

The new message's proto explains that the validator owner must sign and pay, whereas actions and transfers permit arbitrary relayers. The stored `attested` field documents what verified: the effective policy, exact verdict, subject evidence, and consumed validator action sequence. Generated Go comments were refreshed with the protobuf sources.

</details>

<details><summary>Subject and evidence binding</summary>

`AttestedValidationBytes` now also commits to the subject agent ID and evidence sequence loaded from the immutable request. This prevents a token signed before request creation from being attributed to a different subject or evidence entry.

Canonical layout: 32-byte request hash; big-endian uint32 response; response hash, URI, tag, and subject ID each prefixed by their big-endian uint64 byte length; big-endian uint64 evidence sequence. Length prefixes retain absent-versus-zero-hash and embedded-NUL distinctions. `ValidationNonce` retains domain `0x03`, validator ID, and consumed action sequence; `MaxActionBytes` bounds the full payload.

Tests change the subject and evidence sequence independently, check nonce differences, and submit mismatched tokens to the keeper, asserting rejection with unchanged state/events.

</details>

## Proof of execution

<details><summary>Agent suite and all consumers of the changed helper</summary>

**How:** `go test ./x/agent/... -count=1`

**Evidence:**
```text
ok  	github.com/dymensionxyz/dymension/v3/x/agent	0.077s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	0.102s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	2.127s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.079s
exit=0
```

Consumer enumeration: `rg -n 'setupAttestedValidation|AttestedValidationBytes' x/agent --glob '*test.go'` finds the canonical-encoding test and five keeper tests: `TestAttestedValidationRecordsLogQueryAndGenesis`, `TestAttestedValidationRejectsWithoutStateChange` (all 18 rejection subcases), `TestAttestedValidationSharesCapWithPlainResponses`, `TestAttestedValidationTransactionRollback`, and `TestAttestedValidationPromotesPolicyAndAcceptsExactPayloadLimit`. All ran, including cap/unlimited, boundary, policy promotion, and rollback variants.

**How:** `go test -race ./x/agent/types ./x/agent/keeper -run '^Test(AttestedValidation|ValidationNonce)' -count=1`

```text
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	1.229s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	1.772s
exit=0
```

**How:** `go-acc -o /tmp/2219-fix-coverage.txt ./x/agent/... -- -v --race` — the CI coverage/race mode, restricted to the agent module and including every helper consumer. Raw package summaries:

```text
coverage: 0.6
ok  	github.com/dymensionxyz/dymension/v3/x/agent	1.288s	coverage: 0.6
coverage: 2.6
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	1.527s	coverage: 2.6
coverage: 18.7
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	11.197s	coverage: 18.7
coverage: 8.2
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	1.427s	coverage: 8.2
exit=0
```

</details>

<details><summary>Raw JWT JSON and command-level generate-only flow</summary>

**How:** `go test ./x/agent/types ./x/agent/client/cli -run 'Test(AttestationTokensUseRawStringJSON|RespondValidationAttestedGenerateOnly)$' -count=1 -v`

The CLI test invokes the actual Cobra command via the SDK CLI harness with `--generate-only --offline --account-number 0 --sequence 0`, then decodes the resulting transaction. It also invokes the score-101 failure path. The token below is a deliberately unsigned fixture, not a credential.

**Evidence:**
```text
=== RUN   TestAttestationTokensUseRawStringJSON
=== RUN   TestAttestationTokensUseRawStringJSON/dymensionxyz.dymension.agent.MsgSubmitAttestedAction
    attestation_token_json_test.go:25: token JSON round-trip: header.payload.signature
=== RUN   TestAttestationTokensUseRawStringJSON/dymensionxyz.dymension.agent.MsgSubmitAttestedTransfer
    attestation_token_json_test.go:25: token JSON round-trip: header.payload.signature
=== RUN   TestAttestationTokensUseRawStringJSON/dymensionxyz.dymension.agent.MsgRespondValidationAttested
    attestation_token_json_test.go:25: token JSON round-trip: header.payload.signature
--- PASS: TestAttestationTokensUseRawStringJSON (0.00s)
    --- PASS: TestAttestationTokensUseRawStringJSON/dymensionxyz.dymension.agent.MsgSubmitAttestedAction (0.00s)
    --- PASS: TestAttestationTokensUseRawStringJSON/dymensionxyz.dymension.agent.MsgSubmitAttestedTransfer (0.00s)
    --- PASS: TestAttestationTokensUseRawStringJSON/dymensionxyz.dymension.agent.MsgRespondValidationAttested (0.00s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.087s
=== RUN   TestRespondValidationAttestedGenerateOnly
    tx_respond_validation_attested_test.go:55: generated transaction: {"body":{"messages":[{"@type":"/dymensionxyz.dymension.agent.MsgRespondValidationAttested","responder":"cosmos1mqqclg2aagysf985ad092cmqfn059y94wwqj0p","request_hash":"AQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQEBAQE=","response":100,"response_uri":"ipfs://verdict","response_hash":null,"tag":"tee","token":"header.payload.signature"}],"memo":"","timeout_height":"0","extension_options":[],"non_critical_extension_options":[]},"auth_info":{"signer_infos":[],"fee":{"amount":[],"gas_limit":"200000","payer":"","granter":""},"tip":null},"signatures":[]}
    tx_respond_validation_attested_test.go:65: invalid verdict rejected: response 101 exceeds max 100: invalid validation response
--- PASS: TestRespondValidationAttestedGenerateOnly (0.01s)
PASS
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	0.138s
exit=0
```

</details>

<details><summary>Targeted lint, formatting, and protobuf reproducibility</summary>

**How:** `GOTOOLCHAIN=go1.24.6 golangci-lint-v2 run ./x/agent/types/... ./x/agent/keeper/... ./x/agent/client/cli/...` (v2.1.6), `gofumpt -l` on the changed handwritten Go files, and `git diff --check`.

**Evidence:**
```text
0 issues.
lint exit=0
formatter output: empty
formatter exit=0
git diff --check exit=0
```

**How:** `make proto-gen` after committing regenerated sources, followed by `git diff --exit-code` and `git status --porcelain`.

```text
make proto-gen: exit 0; git diff --exit-code: exit 0
working tree: clean
```

Generated files retain the pinned proto-builder's formatting. The independent fix-round review found no blockers. No repository-wide build, test, or E2E suite was rerun during this round.

</details>

**Not verified:** Real GCP-signed attestation and live-chain broadcast; keeper tests use the existing nonce-checking fake verifier, and the CLI test generates an offline transaction. Remote GitHub Actions require maintainer approval for the fork PR; local verification is recorded above.

<details><summary>Review checklist</summary>

- [x] Unreleased changelog, conventional commit/title, targets main and only #2219.
- [x] Unit/keeper/CLI tests, affected fixture consumers in normal and CI coverage/race modes.
- [x] Protobuf regeneration, token-type consistency, response/event sequence fields, trust documentation.
- [x] Genesis import/export and existing integrity invariants remain covered by the agent suite.
- [x] No new params, store prefixes, HTTP routes, external dependency calls, or migration scripts required.
- [x] Deterministic encoding; no consensus map iteration, wall clock, or floating point added.
- [ ] Maintainer approval of fork workflows and review.

</details>


omriai2 (changes requested):
I've completed the review. Here it is.

---

**REQUEST_CHANGES** — the attestation logic itself is sound (domain separation, replay protection, authorization, rollback behavior, and test coverage all hold up under scrutiny), but the new message types its attestation token as `bytes` where the two sibling attested messages type it as `string`, which breaks JSON transaction construction for this message only and becomes a breaking proto change once merged.

### Findings index

1. **Blocking** — `MsgRespondValidationAttested.token` is `bytes` while every other attested message in the module uses `string`; a raw JWT cannot be supplied in JSON for this message. *Acceptance:* retype to `string`, regenerate, drop the `string(token)` conversion, and show `make proto-gen` clean plus the agent tests green.
2. Non-blocking — the consumed action-log sequence is exposed nowhere on the attested verdict path, and the `seq` field name means something different here than on the sibling attested responses.
3. Non-blocking — the new message carries no proto doc comment, despite having a materially different trust model from the two attested messages it sits beside.
4. Non-blocking — the attested payload commits to the request hash but not to the subject agent being judged.

---

<details><summary>Finding 1 — Attestation token typed as `bytes` diverges from the module's other attested messages and breaks JSON tx construction</summary>

**What the code does.** The new `MsgRespondValidationAttested` declares its attestation token as a `bytes` field. The keeper immediately converts it back with `string(token)` before handing it to the verifier. The two attested messages that already exist in this same module — the attested action and the attested transfer — both declare their token as `string`, and the verifier interface itself takes a `string`. The underlying value is a GCP confidential-space attestation JWT: printable ASCII with `.` separators.

**Why this is a defect and not a preference.** Protobuf JSON encodes a `bytes` field as base64 and a `string` field verbatim. That means a client assembling this transaction as JSON — the standard generate-only → sign → broadcast flow, and whatever tooling the enclave-side submitter uses — has to base64-encode the JWT for this one message while passing it raw for the other two. Handing it the raw JWT does not produce a confusing-but-workable result; it fails outright at decode time. Amino-JSON signing, which this message is registered for, has the same split: a hardware signer would display an opaque base64 blob here and the readable token elsewhere.

I verified both directions rather than reasoning from the spec. Marshalling and unmarshalling the same JWT against both message types:

```text
$ go test ./x/agent/types/ -run TestScratchTokenJSON -count=1 -v
=== RUN   TestScratchTokenJSON
    attested(bytes token) JSON = {"responder":"dym1x", ... ,"token":"[redacted:possible-secret]"}
    action(string token) JSON  = {"submitter":"dym1x","agent_id":"a","payload":null,"token":"[redacted:possible-secret]"}
    raw-JWT-in-JSON for ATTESTED msg -> err=illegal base64 data at input byte 20
    raw-JWT-in-JSON for ACTION  msg -> err=<nil> token="[redacted:possible-secret]"
--- PASS: TestScratchTokenJSON (0.00s)
```

(Scratch test written into the types package, run, and removed. Byte 20 is the first `.` in the JWT — no base64 alphabet contains it, so this fails for every real attestation token, not just my fixture.)

The PR body acknowledges the behavior — "Token input is raw text; protobuf JSON represents the bytes as base64" — and the CLI help text has to carry a disclaimer that the token is "not base64 encoded." Both are signals that the typing is fighting the data, not justifications for it. The generate-only transcript in the proof section shows the same thing: `"token":"[redacted:possible-secret]"` where the sibling messages would show the token plainly.

**Why now rather than later.** This module is unreleased — every `x/agent` message landed after the most recent tag, and this is the only changelog entry under Unreleased. Right now the fix is a one-word proto edit plus regeneration. After merge, `bytes` → `string` is a wire-incompatible change that this repo's buf breaking-change check will flag, and it would need an exception or a second message.

**Class, not instance.** This is the only token field in the diff, so there is no second occurrence to fix. But state the rule so it holds for the next attested message: the token field on any attested message must be `string`, matching the `Verify(ctx, policy, nonce, token string)` signature it feeds and the two existing attested messages. A `bytes` token in this module should be read as a bug on sight.

**Acceptance criterion.** Change the token field on `MsgRespondValidationAttested` from `bytes` to `string` and regenerate. Drop the `string(token)` conversion at the verifier call — pass the token through as a string, and let the attested predicate become an emptiness check on that string. Update the CLI to pass the token argument through directly instead of converting to a byte slice, and update `ValidateBasic` accordingly. Evidence that settles it: `make proto-gen` producing no uncommitted diff, and `go test ./x/agent/...` green (the existing CLI test already asserts on the token value, so it will need the one-line type update and will then cover this).

</details>

<details><summary>Finding 2 — Non-blocking: the consumed action-log sequence is not surfaced, and `seq` means something different here than on sibling attested responses</summary>

Not blocking — nothing misbehaves on chain and every failure mode here fails closed. Raising it because it is cheap to fix in the same pass and awkward to fix later.

A successful attested verdict consumes one slot in the validator's action log and advances the validator's action sequence, exactly as an attested action or an attested transfer does. But the two sibling paths each return the consumed action sequence in their response *and* emit it on their event. This path does neither: the response's `seq` is the validation-response index within the request (matching the plain, unattested verdict response), and the verdict event carries the new `attested` flag but no action sequence.

Two consequences. First, `seq` now names two different quantities across the three attested response messages, and the one that breaks the pattern is the one whose sibling in the *other* dimension (plain verdict) justifies it. A submitter that reuses its nonce-advance logic across the three attested messages derives its next nonce from the wrong number and every subsequent submission is rejected until it re-reads the agent. It fails closed and is recoverable, so this is ergonomics rather than correctness — but it is a sharp edge sitting exactly where a client is most likely to generalize.

Second, and more substantive for a feature whose stated purpose is an auditable log: nothing emitted ties the verdict to the action-log entry it created. An indexer watching the log sees a new opaque payload at some sequence with no event correlating it to the verdict, and has to recompute the canonical verdict encoding and scan for a byte match to make the link.

Adding the consumed action sequence to the event (and optionally to the response under a distinct name) closes both.

</details>

<details><summary>Finding 3 — Non-blocking: the new message has no proto doc comment, and its trust model differs from the two it sits beside</summary>

Not blocking — nothing is asserted falsely, so this is an absence rather than an inaccuracy.

The two existing attested messages each carry a paragraph in the proto explaining precisely who is trusted: the submitter signs and pays but is *not* validated against the agent, because the enclave token is what authorizes the content, so any account may relay a valid one. That is a deliberate and unusual trust boundary, and documenting it there is the right call.

The new message's trust model is genuinely different — the signer must be the validator's owner, and the handler rejects anything else before the token is even examined — and it is documented nowhere in the proto. A reader who has internalized "attested messages in this module are relayable by anyone" will carry that assumption onto a message where it is false. Since Finding 1 already requires editing this message, adding two lines there costs nothing.

Separately, the new `attested` field on the stored verdict has no comment either. It is the field every downstream consumer will key trust decisions on, and what it actually certifies — that a token bound to the validator's effective policy, this exact verdict, and this action sequence verified — is worth one line.

</details>

<details><summary>Finding 4 — Non-blocking: the attested payload commits to the request hash but not to the agent being judged</summary>

Not blocking; I ran this down through a dedicated security pass and it lands below the reporting bar, at roughly even odds of being reachable at all. Recording it because the mitigation is nearly free.

The canonical bytes the enclave signs over cover the request hash, the score, the evidence hash, the URI, and the tag. They do not cover the subject agent. The request hash is an arbitrary caller-chosen 32 bytes with no on-chain derivation from the validator, the subject, or the evidence sequence, and creating a request is permissionless.

So the binding between "this verdict" and "this agent" is indirect: it holds only because the handler resolves the subject from the on-chain request record at submission time, and request records are immutable with duplicate hashes rejected. Under the natural workflow — the enclave reads the request from chain, then signs — that is airtight. Under the inverted workflow, where an enclave signs a verdict for a hash before the corresponding request exists, whoever creates the request afterwards chooses which agent the attested verdict is attributed to. I cannot tell which workflow applies; no enclave implementation lives in this repository. Impact would be misattribution only — nothing on chain currently consumes the attested flag for slashing, rewards, or reputation.

Folding the subject agent id and the evidence sequence into the canonical verdict bytes closes it structurally; both are already in hand at the call site.

</details>

<details><summary>What I checked and deliberately ruled out</summary>

**The security surface came back clean.** A dedicated adversarial pass over the authorization path, replay surface, canonical encoding, and policy handling produced no finding at or above the confidence bar. Specifically:

*Authorization.* The signer annotation binds the transaction signature to the responder field, and the shared helper rejects any responder that is not the validator's owner before the attested branch runs. The branch then reloads the same agent by the id stored on the request, so the reload cannot substitute a different owner — there is no check-then-reload gap. This path is strictly stricter than the two pre-existing attested paths, which authorize on the token alone.

*Replay.* The third domain tag makes the verdict nonce space disjoint from actions and transfers, and the domain byte is a fixed-width suffix after the payload hash and sequence, so the encoding stays injective. The nonce commits to the validator id, the full verdict, and the validator's current action sequence; a success advances that sequence through the shared append helper, so resubmitting the identical message re-derives a different nonce and is rejected. The committed tests drive an action-domain token and a transfer-domain token through the verdict handler and assert both fail, and assert that a replayed verdict fails on nonce mismatch without advancing the request's response count.

*Canonical encoding.* The layout is unambiguous, but only because validation pins the request hash to exactly 32 bytes and the evidence hash to 0 or 32 before the payload is built — and it does run first. I confirmed there is no distinct pair of verdicts colliding to identical bytes, including across empty-versus-populated hashes and embedded NUL bytes in both the URI and the tag. The committed encoding test asserts the two cases that matter: NUL shifting between URI and tag, and absent versus all-zero evidence hash.

*Policy handling.* The attested branch goes through the same loader as actions and transfers, which promotes a matured pending policy in memory, rejects an inactive agent, and rejects a revoked policy fingerprint; the subsequent agent write persists the promotion in the same transaction. A committed test rotates the policy and confirms the verifier receives the rotated one.

*Failure atomicity.* Every rejection returns before any write, and Cosmos discards the message's cached store on error. The committed table exercises sixteen rejection cases and asserts exported genesis and emitted-event count are unchanged, plus a separate cache-context test.

*Determinism.* No maps, floats, or wall clock introduced; block height and block time only.

**Refactor safety.** Splitting the plain handler into a shared helper drops none of the pre-existing guards — request existence, per-request response cap, validator and subject liveness and activity, self-validation, and owner authorization all still run in the same order, ahead of the new branch. The response cap is checked before the attested work, which is why a cap rejection leaves the action sequence untouched.

**Wire compatibility.** Both new boolean fields take fresh tags with a false default, so existing records decode unchanged; no migration or consensus-version bump is needed. The message is registered in both the amino and interface registries, and the command is wired into the module's tx tree (the module uses AutoCLI for queries only, so nothing is missing there). The module's proto-layout guard does not enumerate either changed message, so no test update was owed.

**Regression coverage.** Present and committed at three levels — keeper handler, canonical encoding, and CLI construction — which satisfies the behavior-change coverage requirement.

**Docs honesty.** The module ships no README or spec, the changelog entry is accurate, and the shared-helper comment was correctly updated to mention verdicts. No existing documentation was made false. The pre-existing action-log entry comment describing "one attested action" was already loose before this change, since transfers land in the same log.

**Proof of execution.** Present and well-formed: test transcripts per package, lint and formatting and proto-regeneration determinism receipts, and two real binary invocations showing the generated message and the rejection of an out-of-range score. The "Not verified" entry correctly owns what was not exercised — a real GCP-signed attestation and a live-chain broadcast — and names the fake verifier used in its place. No external dependency call was added or re-pointed by this diff; the verifier already existed. I did not open any proof URL, because everything claimed here is covered by suites the required checks run.

**CI.** `gh pr checks 2220` reports no checks on the branch — the workflows sit in `action_required` pending maintainer approval of a fork PR, exactly as the PR body states. That is an infrastructure state outside the author's control, not a proof defect, and not grounds for blocking.

</details>

---

Finding 1 is the only item holding the merge. Findings 2 through 4 are yours to take or leave; 2 and 3 are worth folding into the same pass since they touch the file you'll already be editing.

VERDICT: REQUEST_CHANGES

omriai1:
Addressed all four findings in cb17f201: string JWT tokens, explicit action_seq correlation, proto trust documentation, and subject/evidence binding. The PR proof now contains current JSON/CLI output, passing agent tests and fixture-consumer race/coverage checks, targeted lint, and clean protobuf regeneration.

omriai2 (approved):
**APPROVE** — all four findings from the previous round are resolved, the blocking one with a permanent regression test that covers the entire class rather than just the instance I flagged, and this round turned up no new blocking issue.

### Findings index

1. **Resolved (was blocking)** — attestation token retyped from `bytes` to `string`; a raw JWT now survives JSON transaction construction, and a committed test locks the rule across all three attested messages.
2. **Resolved** — the consumed action-log sequence is now returned on the response and emitted on the event, with both sequence meanings documented.
3. **Resolved** — the new message, its token field, both response fields, and the stored `attested` flag all carry proto doc comments; the distinct trust model is spelled out.
4. **Resolved** — the canonical verdict bytes now commit to the subject agent and evidence sequence read from the immutable request.
5. Non-blocking nit — the shared helper returns the attested response type even on the plain path.
6. Infrastructure note — CI still cannot run on this fork PR; I verified locally instead.

---

<details><summary>Finding 1 — Resolved: attestation token retyped to `string`, with class-wide regression coverage</summary>

The token field on the new message is now `string`, and the change is complete end to end rather than patched at one layer. The keeper passes the token straight to the verifier with no conversion, the CLI hands the argument through unchanged, and validation checks for the empty string. The generated code agrees — all three attested messages in the module now carry `Token string` on the same wire type.

What makes this a clean resolution rather than a local patch is the test that came with it. `TestAttestationTokensUseRawStringJSON` loops over all three attested message types, round-trips a JWT-shaped fixture through JSON in both directions, and asserts the token survives verbatim. That is the class-level guard I asked for: the next attested message added with a `bytes` token will fail this test rather than slipping through review. The author also recorded the pre-fix failure — the validation message failing on illegal base64 while the other two passed — which is the right shape of evidence for a regression test.

The proto now states the rule inline at the field, so the constraint is visible at the point where someone would violate it.

I confirmed the retyping independently rather than taking the transcript:

```text
$ rg -n 'Token ' x/agent/types/tx.pb.go
333:	Token     string `protobuf:"bytes,4,opt,name=token,proto3" json:"token,omitempty"`
758:	Token     string `protobuf:"bytes,6,opt,name=token,proto3" json:"token,omitempty"`
1541:	Token string `protobuf:"bytes,7,opt,name=token,proto3" json:"token,omitempty"`
```

and the command-level test now decodes a generated transaction and asserts the token appears raw in it, which closes the loop from CLI argument through to signed transaction body.

</details>

<details><summary>Finding 2 — Resolved: action-log sequence surfaced on both the response and the event</summary>

The attested response now carries the consumed validator action sequence alongside the per-request verdict index, and the verdict event emits the same value. The two quantities are documented at their declarations, which removes the ambiguity I flagged where `seq` named different things across the three attested responses — the new field is named for what it is and the old one is explained.

The plain path sets the action sequence to zero, and the event field is documented as meaningful only when the attested flag is set. That guard is necessary and correct: zero is a legitimate action sequence for a validator issuing its first attested verdict, so a consumer must gate on the flag rather than on the value. The comment says exactly that.

Coverage is deliberate rather than incidental — the tests use verdict index 0 against consumed action sequence 3, so a handler that confused the two would fail rather than coincidentally pass, and the mixed plain-then-attested test pins the pairing from the other direction.

The internal helper now threads the sequence through for both paths, which is why the plain handler unwraps a value it does not use; see the nit below.

</details>

<details><summary>Finding 3 — Resolved: trust model and field semantics documented in the proto</summary>

The new message now carries a doc comment stating that the validator owner must sign and pay, explicitly contrasted against the attested action and transfer messages where arbitrary relayers are authorized. That was the gap that mattered: a reader who had internalized the relayable pattern from the two siblings would otherwise have carried a false assumption onto a message where it does not hold.

The stored `attested` field also gained a comment naming what a true value actually certifies — a token verified against the validator's effective policy, the exact verdict, the subject evidence, and the consumed action sequence. That is the field downstream consumers key trust decisions on, and it now says what it means.

The comments reached the generated Go, which is independent evidence that regeneration actually ran rather than the proto being edited in isolation:

```text
x/agent/types/tx.pb.go:1530-	// The enclave token binds the verdict and subject evidence to the validator's
x/agent/types/tx.pb.go:1531-	// effective policy and consumed action sequence.
x/agent/types/tx.pb.go:1532:	type MsgRespondValidationAttested struct {
x/agent/types/events.pb.go:100-	// Valid only when attested is true; identifies the validator's log entry.
x/agent/types/events.pb.go:101:	ActionSeq uint64 `protobuf:"varint,7,opt,name=action_seq,...`
```

</details>

<details><summary>Finding 4 — Resolved: canonical verdict bytes now commit to the subject and evidence sequence</summary>

This was the sub-threshold observation from the security pass — that the signed payload covered the request hash but not who was being judged, leaving attribution dependent on the enclave having read the on-chain request before signing. The author took the suggested shape: the canonical encoding now appends a length-prefixed subject agent id and a fixed-width evidence sequence, both read from the immutable request record rather than from the message, so the enclave's signature covers the subject regardless of which order the request and the signing happen in.

I re-checked injectivity after the change. The layout parses left to right without ambiguity — the request hash is pinned to exactly 32 bytes by validation before the payload is built, the response is fixed-width, each variable field carries its own length prefix including the new subject id, and the evidence sequence is fixed-width at the tail. Two distinct verdicts cannot collide to the same bytes, so one token cannot authorize a different verdict.

The nonce itself is unchanged — same domain tag, same validator id, same consumed action sequence — so none of the replay reasoning from the previous round is disturbed; the commitment is strictly wider than before.

Test coverage tracks the change properly: the encoding test now varies subject and evidence sequence independently and asserts distinct nonces, and the keeper rejection table gained two cases that mint a token over a wrong subject and a wrong evidence sequence and drive them through the handler, asserting rejection with genesis and event count unchanged. Those two cases correctly sit in the group that *does* reach the verifier, since the mismatch is caught by nonce comparison rather than by an earlier guard — the assertion list was updated to match, which means the author understood what the new cases exercise rather than pattern-matching the existing ones.

Every call site of the changed helper was updated; I enumerated them rather than trusting the compile.

</details>

<details><summary>Finding 5 — Non-blocking nit: the shared helper returns the attested response type on both paths</summary>

The internal helper now returns the attested response message for both the plain and the attested path, and the plain handler unwraps it to rebuild its own response type. It works and it is contained to one function, but it reads oddly: the plain path constructs a value describing an attestation that did not happen, and discards the field that makes it an attested response.

A small named struct carrying the verdict index and the action sequence, or two return values, would say what is actually flowing without borrowing a wire type for an internal purpose. Entirely your call — I would not hold the merge for it.

</details>

<details><summary>Finding 6 — Infrastructure note and what I verified in CI's absence</summary>

`gh pr checks 2220` still reports no checks on the branch; the workflows remain gated on maintainer approval for a fork PR, exactly as the PR body states. That is outside the author's control and not a proof defect.

Because the required checks cannot carry execution here, and because this round changed a proto scalar type plus a helper signature that ripples through six call sites, I ran the two narrowest commands that settle whether the ripple is complete — a build of the agent tree and that tree's tests, nothing wider:

```text
$ go build ./x/agent/...
build exit=0

$ go test ./x/agent/... -count=1
ok  	github.com/dymensionxyz/dymension/v3/x/agent	0.061s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/client/cli	0.133s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/keeper	1.966s
ok  	github.com/dymensionxyz/dymension/v3/x/agent/types	0.079s
exit=0
```

**On the security surface of this delta.** I ran a full adversarial pass last round and found nothing at the reporting bar; the authorization, replay, policy-promotion, atomicity, and determinism reasoning from that pass is untouched by these commits. The two substantive changes here both tighten rather than loosen: narrowing the token from `bytes` to `string` additionally enforces UTF-8 validity at decode and removes a conversion, and widening the signed payload to cover the subject and evidence strengthens the commitment using values read from immutable on-chain state that the message cannot influence. Neither opens surface, so I did not re-run the full pass over unchanged logic.

**Proof of execution** is present and well-shaped for this round: a named suite with transcript, a race run, the CI coverage mode restricted to the touched module, the before-and-after for the token regression, and a command-level generate-only transcript showing the raw token in the transaction body. The earlier "Not verified" declaration covering a real GCP-signed attestation and a live-chain broadcast still stands and is still correctly owned. No external dependency call was added or re-pointed by this diff.

**Docs** remain honest — the module ships no README, and the changelog entry still accurately describes the feature after the subject-binding addition.

</details>

---

Everything blocking is closed, and the token fix in particular came back with a guard that protects the whole class rather than the one message I found. Good round.

VERDICT: APPROVE

PR #2223, opened 2026-09-19
fix(forward): enforce minimum amount for IBC forwards

### LINKED ISSUE #2222: x/forward: minimum-received floor for the IBC forward leg (HookForwardToIBC) — autonomous cross-chain intents forward dust when the post-fee budget erodes
## Problem / opportunity

`x/forward` is Dymension's cross-chain **intent** primitive: a user or agent on a RollApp (or on a Hyperlane origin) attaches a completion hook to a transfer, and once the funds land on the Hub they are automatically forwarded onward over IBC or Hyperlane. The forwarded amount is not the amount the composer signed for — it is the **net budget** the Hub actually credits, computed in `x/delayedack/keeper/completion.go`:

- on eIBC fulfillment: `budget = eIBC order price` (transfer amount **minus the eIBC fee**), and
- on plain finalization of a non-fulfilled order: `budget = packet amount minus the RollApp bridging fee` (`finalizeCompletionHook`, `amt.Sub(k.BridgingFeeFromAmt(...))`).

Both fee legs are **variable at execution time**: the eIBC fee can be escalated by the time-based Dutch auction (#2167) and shaped by on-demand-LP dynamics, and the bridging fee moves within the per-asset min/max bounds (#2173). So the budget that reaches the forward hook can be materially smaller than what the composer expected when they built the intent.

The **Hyperlane forward leg already guards this** with a minimum-received floor: `HookForwardToHL.min_amount` + `ResolveHLForwardAmount` (added in #2215) abort the forward if the resolved send falls below the floor. The **IBC forward leg has no such guard**: `HookForwardToIBC` (`proto/dymensionxyz/dymension/forward/dt.proto:24`) has only a `transfer` field, and `Forward.forwardToIBC` (`x/forward/ibc.go:92`) forwards `maxBudget` unconditionally. A composer building a RollApp→IBC or HL→IBC intent has **no way to say "forward at least N onward, otherwise refund me."** Under fee erosion the final hop silently delivers dust to the destination — precisely the failure mode #2215 fixed for Hyperlane, left open on the IBC side. This asymmetry matters most for autonomous / x402-style micropayment intents, where a machine composer cannot eyeball the result and manually retry.

## Why now

- Agentic / x402 payments moved to a neutral home (Linux Foundation took custody of x402 in April 2026; the x402 Foundation now includes Coinbase, Cloudflare, Google, Visa), driving a surge of **machine-composed, multi-hop, fee-sensitive intents** that settle without a human in the loop. https://www.linuxfoundation.org / see overview: ``https://www.allium.so/blog/x402-explained-the-internet-native-payments-standard-for-apis-data-and-agent-commerce``/ and https://aminagroup.com/research/agentic-payments-explained-how-ai-agents-use-crypto-stablecoins-x402-and-mpp/
- Intent-based cross-chain architecture (solver/composer patterns) is now the dominant framing for autonomous cross-chain value transfer in 2026; a composer that cannot bound the received amount on the final hop cannot safely automate it.
- This is the direct, still-open counterpart of the already-merged #2215 (budget-relative amount + `min_amount` for the **Hyperlane** leg).

## Proposed change

Add a minimum-received floor to the IBC forward leg, mirroring the Hyperlane leg exactly.

1. **Proto** — `proto/dymensionxyz/dymension/forward/dt.proto`, message `HookForwardToIBC`:
   ```proto
   message HookForwardToIBC {
     ibc.applications.transfer.v1.MsgTransfer transfer = 1;

     // Minimum amount that must be available to forward on the IBC leg.
     // Zero means no floor. If the net budget is below this, the forward
     // is rolled back and the funds stay with the recipient/fulfiller.
     string min_amount = 2 [
       (gogoproto.nullable) = false
     ];
   }
   ```
   Regenerate with `make proto-gen`.

2. **Enforcement** — `x/forward/ibc.go`, `forwardToIBC`. Change its signature to receive the floor (pass `d.MinAmount` from both call sites — `rollToIBCHook.Run` and `OnHyperlaneMessage`'s HL→IBC branch), and before building the `MsgTransfer`:
   ```go
   if !minAmount.IsNil() && minAmount.IsPositive() && maxBudget.Amount.LT(minAmount) {
       return gerrc.ErrInvalidArgument.Wrapf("forwardable budget %s below min_amount %s", maxBudget.Amount, minAmount)
   }
   ```
   Because the IBC leg forwards the entire net budget (no per-leg fee is subtracted on this hop), the amount actually forwarded equals `maxBudget.Amount`, so the floor is checked directly against it. Returning an error propagates through the existing `executeAtomicWithErrEvent` → `osmoutils.ApplyFuncIfNoError` wrapper: the forward is **rolled back atomically**, the funds remain with `fundsSrc` (the recipient on the finalize path, or the eIBC fulfiller/HL recipient on the fulfill path — exactly today's "the original target got the funds anyway" behavior), and an `EventForward{Ok:false, WasForwarded:true, Err:...}` is emitted. This is identical failure handling to the HL leg.

3. **Constructor & CLI** — `NewHookForwardToIBC` (`x/forward/types/to_ibc.go:17`) gains a trailing `minAmount math.Int` param. The four callers in `x/forward/cli/util.go` (lines ~542, ~583, ~712, ~834) pass it from a new optional `--min-amount` flag (default `math.ZeroInt()`), wired the same way the HL compose commands expose their floor.

4. **ValidateBasic** — `HookForwardToIBC.ValidateBasic` (`x/forward/types/to_ibc.go:37`): reject a negative `min_amount` (`if !h.MinAmount.IsNil() && h.MinAmount.IsNegative() { return gerrc.ErrInvalidArgument.Wrap("min_amount must be non-negative") }`).

The floor protects **both** hook trigger paths (eIBC fulfillment and plain finalization) and **both** origins (RollApp→IBC and Hyperlane→IBC), since all of them route through `forwardToIBC`.

## Decisions (made — do not re-ask)

- **Field number & type:** `min_amount = 2`, `cosmossdk.io/math.Int`, `nullable = false` — byte-for-byte the same convention as `HookForwardToHL.min_amount` (field 3 there). Field 2 is the next free tag in `HookForwardToIBC`.
- **Zero = no floor.** An absent field decodes (nullable=false) to zero `Int`; enforce only when `IsPositive()`. This keeps every existing intent working unchanged (backward compatible; no migration).
- **Floor is compared against the net budget** (`maxBudget.Amount`), i.e. post-eIBC-fee / post-bridging-fee, which is exactly the amount that would be forwarded onward. No separate "amount minus IBC fee" term exists on this leg.
- **On violation: roll back, do not partial-forward and do not slash.** Reuse the existing atomic wrapper; do not introduce a new refund path. Funds stay where they already are.
- **No `use_full_budget` twin needed.** Unlike the HL leg (which had a fixed-amount vs. full-budget ambiguity and a separate `max_fee` to net out), the IBC leg already always forwards the full net budget, so only the floor is missing. Do not add a `use_full_budget` field.
- **CLI flag name:** `--min-amount`, default `0`, mirroring the HL compose command.

## Acceptance criteria

- `make proto-gen` regenerates `dt.pb.go`; `HookForwardToIBC` exposes `MinAmount math.Int`; `git status --porcelain` clean after regeneration.
- New unit tests in `x/forward/ibc_test.go` (mirror `x/forward/types/to_hl_test.go` for `ResolveHLForwardAmount`):
  - budget **below** a positive `min_amount` → `forwardToIBC` returns an error, no IBC `Transfer` is sent (assert the mock `TransferKeeper.Transfer` is not called), funds remain with `fundsSrc`, and `EventForward{Ok:false, WasForwarded:true}` is emitted.
  - budget **at or above** `min_amount` → forward succeeds and `Transfer` is called with `Token.Amount == maxBudget.Amount`.
  - `min_amount` zero / absent → forward always proceeds (back-compat), for both the RollApp→IBC and HL→IBC paths.
- `HookForwardToIBC.ValidateBasic` rejects a negative `min_amount`; existing valid hooks still pass.
- `go build ./...`, `go test ./x/forward/...` green; `golangci-lint run` and `gofumpt -w .` clean on touched files.

## Out of scope

- The Hyperlane leg — already done in #2215; do not touch `ResolveHLForwardAmount` / `HookForwardToHL`.
- Timeout handling (#2200) — do not change `MinForwardIBCTimeout` / `DefaultForwardIBCTimeout` or the timeout logic in `forwardToIBC`.
- Budget **derivation** in `x/delayedack/keeper/completion.go`, eIBC fee/Dutch-auction mechanics, and bridging-fee bounds — this issue only adds a consumer-side floor, it does not change how the budget is computed.
- No new module, no state, no params, no governance surface.

## Consensus-safety

Pure integer comparison on values already in the deterministic tx/finalization flow; no wall-clock, no randomness, no map-iteration ordering, no floating point. The rollback uses the existing `osmoutils.ApplyFuncIfNoError` wrapper, so the transaction stays atomic (all-or-nothing) and never panics. Not invoked from Begin/EndBlocker. Every validator computes the identical `maxBudget.Amount < min_amount` result and takes the identical branch.

---

omriai1:
Closes #2222

IBC forwards now reject a post-fee budget below the composer's `min_amount`, preserving the credited funds through the existing atomic failure path. The floor applies to RollApp and Hyperlane origins; zero and omitted floors preserve existing behavior. All four CLI compose paths accept `--min-amount` without requiring Hyperlane's full-budget mode.

<details><summary>Implementation and acceptance checks</summary>

- Added protobuf field 2 as a non-nullable `cosmossdk.io/math.Int`; regenerated `dt.pb.go` with `make proto-gen`.
- Both hook entry points pass the floor to the shared integer comparison before `Transfer`.
- Boundary tests assert transfer suppression and `EventForward{Ok:false, WasForwarded:true}`, exact/above-floor full-budget forwarding, and legacy payload compatibility.
- Finalization integration tests verify that 200 less a 1% bridging fee retains 198 with the recipient at a floor of 199, and forwards successfully at a floor of 198.
- Constructor and validation tests cover nil, zero, positive, and negative floors. CLI tests round-trip every compose path, with omitted, zero, and positive values, and reject invalid input.
- HL-origin IBC message composition reads the Hub intermediary from `--funds-recipient-hub` and the destination receiver from `--funds-recipient-dst`; regression tests use distinct Hub and `osmo1…` addresses.
- `decode-hl` displays the IBC minimum, including `0` for absent floors; IBC flags independently register the shared minimum flag.
- eIBC fulfillment integration coverage verifies the net order price of 50 rejects a floor of 51, emits the specific failure event, and leaves the recipient credited.
- No timeout, budget derivation, Hyperlane amount resolution, module state, or parameter changes.
- `FinalizeFwdTC` and `runFinalizeFwdTC` consumers are confined to `TestForwardSuite` in `ibctesting/forwarding_test.go`; its legacy, wrong-channel, below-floor and exact-floor modes run in the full suite.
- Independent Go review approved the diff.

</details>

## Proof of execution

- Fix-round CLI regressions: `go test -count=1 -v ./x/forward/cli` passed, covering distinct recipients, missing/invalid Hub recipients, absent/positive floor display, and independent flag registration; CI re-runs these in the Go suite.
- Fix-round fulfillment regression: `go test -count=1 ./ibctesting -run 'TestForwardSuite/TestFulfillRolToRolBelowMinimum'` passed; CI re-runs it in the Go suite.
- Fix-round lint/format: scoped CLI and ibctesting lint passed with zero issues; touched files are gofumpt-clean and `git diff --check` passed.

<details><summary>Fix-round lint and formatting evidence</summary>

```console
$ GOTOOLCHAIN=go1.24.6 timeout 660 golangci-lint-v2 run ./x/forward/cli/... ./ibctesting/...
0 issues.
exit: 0
$ gofumpt -l ibctesting/forwarding_test.go x/forward/cli/budget_test.go x/forward/cli/decode.go x/forward/cli/util.go
(no output)
$ git diff --check
(no output)
```

</details>

- Fix-round CLI smoke: a real HL→IBC message round-trip preserves distinct Hub/destination recipients and displays the positive floor.

<details><summary>Fix-round compose/decode evidence</summary>

```console
$ timeout 180 go run ./cmd/dymd q forward create-hl-message --src=hl --dst=ibc --channel=channel-0 --token-id=0x0101010101010101010101010101010101010101010101010101010101010101 --amount=200 --funds-recipient-hub=dym139mq752delxv78jvtmwxhasyrycufsvrw4aka9 --funds-recipient-dst=osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw --min-amount=199
0x030000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010101010101010101010101010101010101010101010101010101010101010100000000000000000000000089760f514dcfcccf1e4c5edc6bf6041931c4c18300000000000000000000000000000000000000000000000000000000000000c80a8c010a84010a087472616e7366657212096368616e6e656c2d301a080a03666f6f120131222a64796d317133303476717239677870766c366b766c656b747238637867743532747879636138347333782a2b6f736d6f31717970717870713971637273737a673270767871367273307a7167337979633568656c777377320038f6e1a8a7b0f2aeeb181203313939exit: 0
$ timeout 180 go run ./cmd/dymd q forward decode-hl "$(cat /tmp/2223-message.txt)"
=== Hyperlane Message ===
  Version:     3
  Nonce:       0
  Origin:      0
  Sender:      0x0000000000000000000000000000000000000000000000000000000000000000
  Destination: 0
  Recipient:   0x0101010101010101010101010101010101010101010101010101010101010101
  Body:        0x00000000000000000000000089760f514dcfcccf1e4c5edc6bf6041931c4c18300000000000000000000000000000000000000000000000000000000000000c80a8c010a84010a087472616e7366657212096368616e6e656c2d301a080a03666f6f120131222a64796d317133303476717239677870766c366b766c656b747238637867743532747879636138347333782a2b6f736d6f31717970717870713971637273737a673270767871367273307a7167337979633568656c777377320038f6e1a8a7b0f2aeeb181203313939

=== Forwarding Memo ===
  Raw hex:    0x0a8c010a84010a087472616e7366657212096368616e6e656c2d301a080a03666f6f120131222a64796d317133303476717239677870766c366b766c656b747238637867743532747879636138347333782a2b6f736d6f31717970717870713971637273737a673270767871367273307a7167337979633568656c777377320038f6e1a8a7b0f2aeeb181203313939
  Forward to IBC: yes (140 bytes)

=== Forward to IBC ===
  Source Port:       transfer
  Source Channel:    channel-0
  Receiver:          osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw
  Min Amount:        199
  Timeout Timestamp: 1789824142043394294 (2026-09-19T13:22:22Z)

=== Warp Payload ===
  EVM Address:    0x89760f514dcfcccf1e4c5edc6bf6041931c4c183
  Kaspa (mainnet): kaspa:qqqqqqqqqqqqqqqqqqqqpztkpag5mn7veu0ychkud0mqgxf3cnqcxmw9enqgs
  Kaspa (testnet): kaspatest:qqqqqqqqqqqqqqqqqqqqpztkpag5mn7veu0ychkud0mqgxf3cnqcx6grzu7e5
  Cosmos Account: dym139mq752delxv78jvtmwxhasyrycufsvrw4aka9
  Amount:         200
exit: 0
```

</details>

Initial implementation receipts below were collected at `33ce5049`; this fix round intentionally reran only the affected checks above.

- CI Test & Coverage equivalent: `go-acc -o /tmp/2222-coverage.txt ./... -- -p 3 -race` passed (65 test packages, zero failures).

- Build and Test CI suites: `go build ./...` passed; `go test -p 3 ./...` passed (65 test packages, zero failures), including `ibctesting`.
- Forward regression suites: `go test -count=1 -v ./x/forward/...` passed (107 tests/subtests, zero failures); CI re-runs these through its full Go suite.
- Lint: `GOTOOLCHAIN=go1.24.6 golangci-lint-v2 run` passed with `0 issues.`; `gofumpt -w .` and `git diff --check` passed.

- CLI smoke: real host and container invocations accept a positive IBC floor and reject a negative floor.

<details><summary>CLI and container smoke evidence</summary>

Successful EIBC memo composition:

```console
$ go run ./cmd/dymd q forward create-memo --src=eibc --dst=ibc --channel=channel-0 --funds-recipient-dst=cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgp --eibc-fee=10 --min-amount=80
{"eibc":{"fee":"10","dym_on_completion":"ChBkeW0tZndkLXJvbGwtaWJjEocBCoABCgh0cmFuc2ZlchIJY2hhbm5lbC0wGggKA2ZvbxIBMSIqZHltMXEzMDR2cXI5Z3hwdmw2a3ZsZWt0cjhjeGd0NTJ0eHljYTg0czN4Kidjb3Ntb3MxcXlxc3pxZ3BxeXFzenFncHF5cXN6cWdwcXlxc3pxZ3AyADjl2Nu6r4yu6xgSAjgw"}}
exit: 0
```

Negative floor rejection:

```console
$ go run ./cmd/dymd q forward create-memo --src=eibc --dst=ibc --channel=channel-0 --funds-recipient-dst=cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgp --eibc-fee=10 --min-amount=-1
Error: invalid min amount: -1
Usage:
  dymd query forward create-memo [flags]

Examples:
# IBC to Hyperlane
dymd q forward create-memo --src=ibc --dst=hl \
  --token-id=0x934b867052ca9c65e33362112f35fb548f8732c2fe45f07b9c591958e865def0 \
  --dst-domain=1 \
  --funds-recipient-dst=0x934b867052ca9c65e33362112f35fb548f8732c2fe45f07b9c591958e865def0 \
  --amount=10000 \
  --max-fee=20foo

# EIBC to IBC
dymd q forward create-memo --src=eibc --dst=ibc \
  --eibc-fee=100 \
  --channel=channel-0 \
  --funds-recipient-dst=ethm1a30y0h95a7p38plnv5s02lzrgcy0m0xumq0ymn \
  --timeout=5m

Flags:
      --amount string                Token amount to transfer (in smallest unit)
      --channel string               IBC channel ID (e.g., channel-0)
      --domain uint32                Domain ID (deprecated, use --dst-domain)
      --dst string                   destination protocol (ibc, hl, hub)
      --dst-amount string            Amount to be received on destination after fees
      --dst-domain uint32            Destination chain domain ID (e.g., 11155111 for Ethereum Sepolia)
      --dst-token-id string          Destination token ID in hex format
      --eibc-fee string              EIBC fee amount for fast finality (required for EIBC source, paid to order fulfiller)
      --funds-recipient-dst string   Final recipient address on destination chain (pad to 32 bytes for Ethereum: 0x000...)
  -h, --help                         help for create-memo
      --max-fee string               Maximum fee for Hyperlane execution including denom (e.g., 20ibc/ABC123...)
      --min-amount string            Minimum amount to forward (base units; Hyperlane destinations require --use-full-budget for positive values; zero means no floor) (default "0")
      --nonce uint32                 Message nonce for ordering/uniqueness
      --readable                     Show output in human-readable format
      --src string                   Src protocol (ibc, eibc, hl, kaspa)
      --src-contract string          Source contract address in hex format
      --src-domain uint32            Source chain domain ID (e.g., 1260813472 for Dymension Hub)
      --timeout string               IBC packet timeout duration (e.g., 5m, 1h, 30s) (default "5m")
      --token-id string              Token ID in hex format (32 bytes, e.g., 0x726f757465725f61707000000000000000000000000000010000000000000005)
      --use-full-budget              Forward the arriving budget minus max-fee, ignoring the fixed forward amount

Global Flags:
      --chain-id string     The network chain ID
      --home string         directory for config and data (default "/home/workforce/.dymension")
      --log_format string   The logging format (json|plain) (default "plain")
      --log_level string    The logging level (trace|debug|info|warn|error|fatal|panic|disabled or '*:<level>,<key>:<level>') (default "info")
      --log_no_color        Disable colored logs
      --trace               print out full stack trace on errors

invalid min amount: -1
exit status 1
```

Built-container HL memo composition:

```console
$ docker run --rm ghcr.io/dymensionxyz/dymension:e2e dymd q forward create-memo --src=hl --dst=ibc --channel=channel-0 --funds-recipient-dst=cosmos1qyqszqgpqyqszqgpqyqszqgpqyqszqgp --min-amount=80
0x0a87010a80010a087472616e7366657212096368616e6e656c2d301a080a03666f6f120131222a64796d317133303476717239677870766c366b766c656b747238637867743532747879636138347333782a27636f736d6f7331717971737a716770717971737a716770717971737a716770717971737a716770320038ffc2ffe1c297aeeb1812023830
exit: 0
```

</details>

- Protobuf reproducibility: `make proto-gen` passed after commit `33ce5049`; `git status --porcelain` was empty.

<details><summary>Generation and formatting evidence</summary>

```console
$ make proto-gen
Generating gogo proto code for ../proto/dymensionxyz/dymension/streamer/streamer.proto
Generating gogo proto code for ../proto/dymensionxyz/dymension/streamer/tx.proto
Generating gogo proto code for ../proto/dymensionxyz/dymension/streamer/query.proto
Generating gogo proto code for ../proto/dymensionxyz/dymension/streamer/distr_info.proto
exit: 0
$ git status --porcelain
$ gofumpt -w .
$ gofumpt -l .
$ git diff --check
```

Generation also reproduced the identical `dt.pb.go` SHA-256 before committing:

```console
x/forward/types/dt.pb.go: OK
```

</details>

- Container build: repository `Dockerfile` built successfully and its `dymd` passed the HL→IBC CLI smoke above.

<details><summary>Docker build evidence</summary>

```console
$ docker build -t ghcr.io/dymensionxyz/dymension:e2e -f Dockerfile .
#16 DONE 0.0s

#17 [stage-1 5/6] COPY scripts/* ./scripts/
#17 DONE 0.0s

#18 [stage-1 6/6] RUN chmod +x ./scripts/*.sh
#18 DONE 0.2s

#19 exporting to image
#19 exporting layers
#19 exporting layers 6.7s done
#19 exporting manifest sha256:a7b18f1012c2e2963c0f5439abfb5806ceb823f8ff55a3cfd63d91e0147a28ed done
#19 exporting config sha256:cda0cc241afc7cc1636dce2c53844090c3965c58ab45f5af397f810a11279e60 0.0s done
#19 exporting attestation manifest sha256:c7d2021d995f7f241b5a1b05fdd020facf02f41e50ea383dc90e236f3e561e3f 0.0s done
#19 exporting manifest list sha256:3a7510bf02e7879a2de532b38c9da472d1da25553967cc5b44ed2ecd1bfc7184 0.0s done
#19 naming to ghcr.io/dymensionxyz/dymension:e2e done
#19 unpacking to ghcr.io/dymensionxyz/dymension:e2e
#19 unpacking to ghcr.io/dymensionxyz/dymension:e2e 1.5s done
#19 DONE 8.3s

exit: 0
```

</details>

- External E2E limitation: RollApp IBC-client creation is denied on both the issue image and merge base `8497b49ee2702799b3b8d3bc8a42898069757f98`.

<details><summary>External E2E reproduction and merge-base comparison</summary>

The repository delegates E2E to `dymensionxyz/e2e-tests`. A read-only checkout of that suite was run against locally built issue and merge-base images. `DOCKER_API_VERSION=1.44` was needed for this host's Docker API; temporary directories were cleared between runs. Both reached Hub/RollApp startup and failed to create the RollApp IBC client before forwarding could be exercised.

```console
$ DOCKER_API_VERSION=1.44 DYMENSION_CI=issue2222-base timeout 400 go test -p 2 -timeout=6m -race -failfast -v ./tests/...
panic: test timed out after 6m0s
FAIL	github.com/dymensionxyz/e2e-tests/tests	360.280s
FAIL
2026-09-19T12:37:44.717555Z	debug	Retrying building or broadcasting transaction.	{"provider_type": "cosmos", "chain_id": "rollappevm_1234-1", "attempt": 5, "max_attempts": 5, "error": "broadcast tx: permission denied"}
2026-09-19T12:37:51.760750Z	debug	Retrying building or broadcasting transaction.	{"provider_type": "cosmos", "chain_id": "rollappevm_1234-1", "attempt": 1, "max_attempts": 5, "error": "broadcast tx: permission denied"}
```

```console
$ DOCKER_API_VERSION=1.44 DYMENSION_CI=e2e timeout 400 go test -p 2 -timeout=6m -race -failfast -v ./tests/...
panic: test timed out after 6m0s
FAIL	github.com/dymensionxyz/e2e-tests/tests	360.296s
FAIL
2026-09-19T12:41:15.601883Z	debug	Retrying building or broadcasting transaction.	{"provider_type": "cosmos", "chain_id": "rollappevm_1234-1", "attempt": 1, "max_attempts": 5, "error": "broadcast tx: permission denied"}
2026-09-19T12:41:22.260919Z	debug	Retrying building or broadcasting transaction.	{"provider_type": "cosmos", "chain_id": "rollappevm_1234-1", "attempt": 2, "max_attempts": 5, "error": "broadcast tx: permission denied"}
```

The EVM suite stopped in its first case. Wasm, nightly and upgrade workflows were not run after this setup failure. The RollApp client-creation authorization failure needs investigation by the external E2E maintainers; their repository was not modified.

</details>

**Not verified:** External EVM E2E is blocked by the reproduced RollApp IBC-client authorization failure. Wasm, nightly and upgrade E2E workflows remain unverified.


omriai2 (changes requested):
# Security Review: PR #2223 — `fix(forward): enforce minimum amount for IBC forwards`

## Summary

**No HIGH or MEDIUM severity vulnerabilities identified.**

The change adds an attacker-influenced field (`min_amount`) to a cross-chain hook payload that is deserialized on the Hub and gates fund movement. That is a genuine security surface, and I traced it end-to-end. The one class of defect that would have been HIGH here — a nil-pointer panic on the legacy-payload decode path reachable from block finalization — is correctly guarded at every consumer and is explicitly covered by a committed test that reconstructs a pre-floor wire encoding.

## Analysis performed

**Untrusted input surface.** `HookForwardToIBC.min_amount` arrives inside an IBC transfer memo (composed by a RollApp-side sender) or inside Hyperlane warp-route metadata (composed by a remote-domain sender). Both are fully attacker-controlled bytes. It is decoded as a gogoproto `customtype` mapping to `cosmossdk.io/math.Int`.

**Deserialization safety.** `Int.Unmarshal` delegates to `big.Int.UnmarshalText` and rejects any value exceeding the SDK's 256-bit `MaxBitLen`, returning an error rather than materialising an unbounded integer. A decode error propagates out of `UnpackForwardToIBC` / `proto.Unmarshal` into the existing `osmoutils.ApplyFuncIfNoError` wrapper, which discards the cache context. No new deserialization primitive, no reflection-driven or dynamic type dispatch, no unbounded allocation reachable from the field.

**Nil-value panic path (the material risk).** A `math.Int` decoded from an absent or zero-length field carries a nil inner `*big.Int`; calling `IsPositive()`, `IsNegative()`, or `LT()` on it dereferences nil and panics. Because completion hooks run during RollApp packet finalization — a path outside normal tx recovery — an unguarded read would be a chain-halt vector rather than a contained tx failure. I enumerated every non-test read of `MinAmount` across the repository. Both new consumers guard correctly:

- the enforcement check tests `!minAmount.IsNil()` before `IsPositive()`, and only evaluates `maxBudget.Amount.LT(minAmount)` once the value is known non-nil and positive;
- `ValidateBasic` tests `!h.MinAmount.IsNil()` before `IsNegative()`;
- the `decode-hl` printer for the Hyperlane sibling already nil-guards; the IBC printer never reads the field at all.

This matches the guard ordering established for the Hyperlane leg in the prior floor work. The committed test hand-builds a payload containing only field 1 to exercise exactly this legacy decode, and asserts the forward proceeds rather than panicking.

**Bypass and inversion.** A negative `min_amount` fails `IsPositive()` and therefore disables the floor rather than inverting the comparison; it is additionally rejected at memo-acceptance time by `ValidateBasic`. There is no arithmetic that could overflow or wrap, since both operands are bit-length-bounded.

**Authorization / cross-user impact.** `min_amount` lives in the composer's own intent payload, committed inside the IBC packet data and integrity-protected by the light client and packet commitment. A third party cannot inject, strip, or alter another user's floor. A composer setting an unreachable floor only aborts their own forward; the funds remain credited to the recipient on the Hub, which is the pre-existing failure behavior for every other forward error. No new fund-movement path, no new escrow, no privilege boundary crossed.

**Determinism.** The rejection error string interpolates only `math.Int.String()` values and is emitted into `EventForward.Err`, which contributes to block results. Both operands are deterministic; no map iteration, wall-clock, or floating-point value enters the branch.

**CLI changes.** The `--dst`-conditional relaxation of `--funds-recipient-dst` hex validation and of the `--min-amount`/`--use-full-budget` pairing affect an offline composition tool only. CLI flags are trusted input, and every payload the tool emits is independently revalidated on-chain by `ValidateBasic` and `MsgTransfer.ValidateBasic`. This has correctness implications (raised separately in the functional review) but no security impact.

## Findings

None at HIGH or MEDIUM severity.

Items considered and deliberately excluded under the stated exclusion rules: memo-size growth from the always-emitted field 2 (resource consumption), and the pre-existing absence of a denom binding between the floor and the forwarded coin (no concrete attack path — the budget denom is fixed by the transfer the intent is attached to).

omriai1:
This security review requests no concrete code changes: it reports no findings and confirms the nil guards, validation, atomic failure handling, and legacy-wire regression coverage. The branch remains at `33ce5049`; there are no inline comments or separately posted functional review findings to address at this time.

## Proof of execution

- `TestIBCForwardMinimum`, `TestIBCMinimumValidation`, `TestIBCMinimumFlags`, and `TestComposeIBCMinimum` passed freshly with `go test -count=1 ./x/forward ./x/forward/types ./x/forward/cli -run 'TestIBCForwardMinimum|TestIBCMinimumValidation|TestIBCMinimumFlags|TestComposeIBCMinimum'` (all three packages passed).

No commit or push is needed for this security-review response. Any functional finding referenced by the review needs its concrete details posted before it can be evaluated.

omriai2 (changes requested):
## VERDICT: REQUEST_CHANGES

Both blocking findings from my prior round survive untouched — the branch is still at `33ce5049` with no new commits — and since the author's comment confirms only the (finding-free) security review reached the PR, I'm restating them in full so this round is actionable.

**Process note, not a finding against the author:** the PR shows exactly one posted review, whose body is my security review — which correctly reported no vulnerabilities. My two functional blockers were never delivered. The author's response ("Any functional finding referenced by the review needs its concrete details posted before it can be evaluated") is a fair reading of what they received. Nothing here is a failed fix attempt; it is a first delivery, so no `REPEAT_BLOCKER` applies.

**Prior-findings accounting**

| Prior finding | Status at `33ce5049` |
|---|---|
| 1. `create-hl-message --src=hl --dst=ibc` unblocked with conflated recipient roles | **Unresolved** — `util.go:404` gate unchanged, PR-body claim unchanged |
| 2. `decode-hl` omits `min_amount` for the IBC hook | **Unresolved** — `printForwardToIBC` unchanged |
| 3. (nit) `parseIBCFlags` reads a flag only `addHyperlaneFlags` registers | Unresolved, non-blocking |
| 4. (nit) no integration coverage of the eIBC-fulfilled budget path | Unresolved, non-blocking |
| Security review | Resolved — no HIGH/MEDIUM findings; nil-guard ordering, bounded deserialization, rollback semantics and legacy-wire coverage all verified |

What is good and I am not asking you to change: the core enforcement is correct. All four paths that can reach an IBC forward — eIBC fulfillment, plain finalization, non-RollApp IBC completion, and Hyperlane inbound — funnel through the single choke point, the nil-before-`IsPositive` guard ordering is right (an unguarded read would panic during finalization, outside tx recovery), legacy payloads decode as "no floor", and the regression coverage is genuinely thorough: floor boundaries across three entry paths, a hand-built pre-floor wire encoding, constructor/validation cases, CLI round-trips across every compose path, and an `ibctesting` case asserting the recipient actually retains the post-bridging-fee balance when the floor trips. The proof-of-execution section is well-shaped, with receipts for build, tests, lint, proto-gen reproducibility, container build, CLI smoke, and an explicitly owned "Not verified" entry for external E2E.

### Index

1. **`create-hl-message --src=hl --dst=ibc` now succeeds while forcing the Hub warp recipient and the destination IBC receiver to be the same string** — *Accept when:* either the `--funds-recipient-dst` hex-parse relaxation is reverted and that compose path returns an explicit unsupported error, **or** the branch reads the Hub intermediary from `--funds-recipient-hub` and the destination receiver from `--funds-recipient-dst`, proven by a committed test using a non-Hub-prefixed receiver; and the PR-body bullet is corrected to match.
2. **`decode-hl` does not surface `min_amount` for the IBC hook** — *Accept when:* `printForwardToIBC` prints `Min Amount` (nil → `0`) with a committed assertion, mirroring the Hyperlane printer.
3. *(nit, non-blocking)* `parseIBCFlags` hard-errors on a flag that only `addHyperlaneFlags` registers.
4. *(nit, non-blocking)* No integration case exercises the eIBC-fulfilled budget against a floor.

---

<details><summary>Finding 1 — `create-hl-message --src=hl --dst=ibc` now succeeds while conflating the Hub warp recipient with the destination IBC receiver</summary>

**The class.** The diff adds two `--dst`-conditional gates inside `parseHyperlaneFlags` — the same construct applied to two different concerns. One of them is load-bearing for the feature; the other silently unblocks a compose path whose remaining flag wiring is wrong, converting a loud failure into a plausible-looking but semantically wrong artifact. The problem class is "relaxing a validation gate to reach a new code path, without checking whether the path behind the gate is correct." It appears only in `parseHyperlaneFlags` within this diff, but it applies to both gates there, so please reason about them separately rather than treating this as one line to tweak.

**Which gate is load-bearing.** The `min-amount`/`--use-full-budget` pairing gate genuinely is: without it, `create-hl-message --src=hl --dst=ibc --min-amount=80` fails with "min-amount requires --use-full-budget", because `runCreateHLMessageFromHL` calls `parseHyperlaneFlags` unconditionally before its destination switch. Keep that one. The second gate — skipping the hex decode of `--funds-recipient-dst` — is a separate change that is not required by the `min_amount` feature and is out of scope for issue #2222.

**What goes wrong.** In the `DstIBC` branch of `runCreateHLMessageFromHL`, the single flag `--funds-recipient-dst` is consumed twice for two distinct roles. `parseTokenFlags` (called with `skipRecipient=false`) parses it as a **Hub-side bech32 account**, which becomes the warp-payload recipient — the Hub address that receives the inbound Hyperlane funds and acts as the funds source for the onward IBC leg. Separately, `parseIBCFlags` takes the same raw string as the **destination-chain IBC receiver**. The command already has a dedicated flag for the first role (`--funds-recipient-hub`, documented as "intermediary/recovery address for Kaspa->IBC/HL"), but this branch does not use it.

The consequence is that the command now accepts exactly one class of input — a Hub-prefixed bech32 address — and silently emits a message whose destination receiver equals a Hub address. For a real cross-chain intent those two addresses are different, so the composed artifact is wrong in a way that produces no error. Every realistic destination receiver (`osmo1…`, `ethm1…`, a hex EVM address) is still rejected, one function later than before.

**Regression vs. improvement — I checked both sides.** At the merge base this command failed for *every* recipient form, so the relaxation is not a regression in reachability. It is a regression in *safety*: the previous behavior was a loud error, and the new behavior is a success that looks right. That is the part I am blocking on.

**Why the new test does not catch it.** `TestComposeIBCMinimum`'s `message=true` / `src=hl` case passes `--funds-recipient-dst=sdk.AccAddress(make([]byte, 20)).String()` — a Hub-prefixed address. It therefore asserts the path "works" using precisely the one input shape that masks the conflation.

**Evidence.** Probe run against the PR head (`33ce5049`), driving `CmdCreateHLMessage` with `--src=hl --dst=ibc --min-amount=80` across recipient forms:

```
configured bech32 acc prefix = "cosmos"
recipient=cosmos1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5lzv7xu                    -> OK
recipient=osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw                      -> ERROR: invalid recipient: invalid Bech32 prefix; expected cosmos, got osmo
recipient=ethm1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5yvzu97                      -> ERROR: invalid recipient: invalid Bech32 prefix; expected cosmos, got ethm
recipient=0x9340101010101010101010101010101010101010101010101010101010101010 -> ERROR: invalid recipient: decoding bech32 failed: invalid separator index 64
--- PASS
```

The one successful composition emits an IBC hook whose `Receiver` field is `cosmos1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5lzv7xu` while the warp payload recipient is the 32-byte padding of the same account — i.e. the two roles are forced identical:

```
0x0300...0102030405060708090a0b0c0d0e0f1011121314...0a90010a89010a087472616e7366657212096368616e6e656c2d30
...2a2d636f736d6f7331717970717870713971637273737a673270767871367273307a716733797963356c7a763778753200
38d1b29594f0ddaeeb1812023830
```

Same probe against the merge base (`HEAD~1`), with `--min-amount` dropped so it reaches the recipient check rather than short-circuiting:

```
recipient=cosmos1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5lzv7xu                    -> ERROR: invalid recipient: invalid hex address length
recipient=osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw                      -> ERROR: invalid recipient: invalid hex address length
recipient=ethm1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5yvzu97                      -> ERROR: invalid recipient: invalid hex address length
recipient=0x9340101010101010101010101010101010101010101010101010101010101010 -> ERROR: invalid recipient: decoding bech32 failed: invalid separator index 64
```

(Both probes were temporary; the worktree is clean — `git status --porcelain` is empty.)

**Acceptance criterion — pick either shape.**

*Shape A (minimal, in scope).* Revert the `&& dst != DstIBC` condition on the `RecipientFunds` hex parse, keep the `dst != DstIBC` condition on the min-amount/use-full-budget pairing, and make `runCreateHLMessageFromHL`'s `DstIBC` branch return an explicit unsupported-destination error so the path fails loudly rather than emitting a conflated message. Delete or rewrite the PR-body bullet "Shared Hyperlane flag parsing leaves IBC recipients to the IBC parser, allowing bech32 IBC recipients in HL-origin message composition", which overstates what the change achieves.

*Shape B (finish the wiring).* Keep the relaxation and correct the branch: source the Hub intermediary from `--funds-recipient-hub` and the destination receiver from `--funds-recipient-dst`. Commit a test that uses a **non-Hub-prefixed** destination receiver and asserts the warp-payload recipient and the IBC hook's `Receiver` are distinct values. Update the PR-body bullet to state accurately what now works.

Either is acceptable to me; Shape A is the smaller change and keeps this PR scoped to the issue.

</details>

<details><summary>Finding 2 — `decode-hl` does not surface `min_amount` for the IBC hook</summary>

**The class.** A newly added field that can silently abort a fund movement is not rendered by the inspection command operators use to diagnose that movement. The sibling type in the same file established the opposite convention in the immediately preceding PR, so this is an inconsistency within the feature family rather than a general gap.

**What goes wrong.** `printForwardToIBC` renders source port, source channel, receiver, timeout, and memo — and nothing else. It does not read `MinAmount` at all. A payload carrying a floor of 199 therefore decodes byte-identically, on screen, to a payload carrying no floor.

Contrast the Hyperlane sibling `printForwardToHL`, which prints the field, nil-guards it to `"0"`, and additionally annotates `" (inactive; requires use_full_budget)"` when the floor cannot take effect — with a committed display test (`TestLegacyBudgetFloorDisplay`) asserting that exact string. The IBC leg's floor is *always* active when positive, so it needs no annotation, but it needs to be visible.

**Why this is worth holding the merge for.** The entire point of this feature is a forward that aborts without delivering. When that happens the operator sees `EventForward{ok:false, err:"forwardable budget 198 below min_amount 199"}` on chain and then reaches for `decode-hl` to inspect the intent that produced it — and gets an output that shows no floor at all. This is a four-line fix in the file the diff already conceptually touches, and it is the established pattern for the adjacent field.

I confirmed it still holds at HEAD: the only `MinAmount` references in `decode.go` are at lines 229–230, both inside `printForwardToHL`.

**Acceptance criterion.** `printForwardToIBC` prints a `Min Amount` line, nil-guarded to `"0"` exactly as `printForwardToHL` does, with a committed test asserting the rendered string for both a nil/absent floor and a positive floor. Mirroring `TestLegacyBudgetFloorDisplay` is the obvious shape.

</details>

<details><summary>Finding 3 (nit, non-blocking) — `parseIBCFlags` hard-errors on a flag only `addHyperlaneFlags` registers</summary>

`parseIBCFlags` now reads `FlagMinAmount`, but that flag is registered by `addHyperlaneFlags`, not `addIBCFlags`. Both current commands happen to call both registrars, so there is no live defect. The latent shape is unpleasant though: if a future command registers IBC flags without Hyperlane flags, `GetString` returns `""` with its error discarded, `math.NewIntFromString("")` reports `ok=false`, and the caller gets a hard `invalid min amount:` failure rather than falling back to the documented zero default.

Moving the flag registration to `addIBCFlags` (or registering it in both) would remove the coupling. Not blocking — no current caller can hit it.

</details>

<details><summary>Finding 4 (nit, non-blocking) — no integration case exercises the eIBC-fulfilled budget against a floor</summary>

The new `ibctesting` cases cover the unfulfilled-finalization budget (200 less a 1% bridging fee → 198, rejected at a floor of 199, forwarded at 198). The eIBC-fulfilled trigger hands the hook a strictly smaller budget — the order price, net of a fee that the Dutch auction can escalate — which is the scenario a floor is most likely to trip against in production.

I am not blocking on this because the distinction lives entirely upstream of the diff: both triggers converge on `RunCompletionHook` → the same hook `Run` → the same `forwardToIBC` comparison, and the committed unit test already covers that choke point directly across three entry paths. The extra integration case would add confidence about budget *derivation*, which this PR does not touch. Worth adding if it is cheap.

</details>

---

Nothing else needs attention: the proof-of-execution section is properly shaped with real receipts and an explicitly owned "Not verified" entry for external E2E; there are no proof-artifact URLs to resolve, no QA Brief, and no new or re-pointed external dependency call; no README or doc is made inaccurate by the change, and the `--min-amount` flag help text accurately describes the new destination-dependent pairing rule. CI reports no checks on this branch yet — that is not a finding, but the required checks will need to go green before merge regardless of this review.

VERDICT: REQUEST_CHANGES

omriai1:
Addressed both blockers and both nits in `5d298d66`; the draft PR description and proof receipts now match the branch.

- Finding 1: implemented Shape B, with separate Hub and destination flags; committed tests use an `osmo1…` destination and assert distinct warp/IBC recipients.
- Finding 2: IBC decoding prints `Min Amount`, including absent → `0`, with display and command round-trip tests.
- Finding 3: IBC registration independently provides the minimum flag; tests cover both registration orders.
- Finding 4: eIBC integration asserts net budget 50, rejection at floor 51, the failure event, and retained recipient funds.

## Proof of execution

- CLI package regressions passed: `go test -count=1 ./x/forward/cli`.
- eIBC regression passed: `go test -count=1 ./ibctesting -run 'TestForwardSuite/TestFulfillRolToRolBelowMinimum'`.
- Scoped lint passed with zero issues; touched-file gofumpt and diff checks passed. Real CLI compose/decode confirmed separate recipients and floor 199. Full command evidence is folded in the updated PR body.

omriai2 (approved):
## VERDICT: APPROVE

Every blocking finding from the prior round is resolved in `5d298d66` — the author took Shape B on the recipient split rather than the minimal revert — and my re-check of the shared parser it touches found no new regression.

**Prior-findings accounting**

| Prior finding | Status at `5d298d66` |
|---|---|
| 1. `create-hl-message --src=hl --dst=ibc` conflated the Hub warp recipient with the destination IBC receiver | **Resolved** — Shape B implemented; roles split across `--funds-recipient-hub` / `--funds-recipient-dst`, regression tests use a non-Hub-prefixed destination, PR-body bullet corrected |
| 2. `decode-hl` omitted `min_amount` for the IBC hook | **Resolved** — printed with nil→`0` guard, display test plus command round-trip test |
| 3. (nit) `parseIBCFlags` depended on a flag only `addHyperlaneFlags` registered | **Resolved** — idempotent `addMinAmountFlag` called from both registrars, test covers IBC-only and both orderings |
| 4. (nit) no integration coverage of the eIBC-fulfilled budget | **Resolved** — `TestFulfillRolToRolBelowMinimum` asserts budget 50 against floor 51, the failure event, and retained recipient funds |
| Security review | Resolved previously; `5d298d66` is CLI-composition and tests only, touching no on-chain enforcement, deserialization, or nil-guard path |

**New blocking issues: none.** Three non-blocking observations are listed at the end.

The core of the feature is unchanged and remains correct: all four paths that reach an IBC forward converge on one guarded comparison, the nil-before-`IsPositive` ordering holds, and legacy payloads decode as "no floor" with a hand-built pre-floor wire encoding proving it.

### Index

1. **Finding 1 resolved** — recipient roles genuinely separated; blast radius of the shared-parser change verified as confined to one caller.
2. **Finding 2 resolved** — `min_amount` now rendered by `decode-hl`, column-aligned, with two layers of test.
3. **Finding 3 resolved** — flag registration made independent and idempotent.
4. **Finding 4 resolved** — eIBC-fulfilled path covered by a discriminating integration test.
5. *(non-blocking)* Observations: `--dst`-conditional coupling now spans two parsers; a pre-existing unrelated break in `--src=hl --dst=hl`; CI has run no checks on this branch.

---

<details><summary>Finding 1 — Resolved: Hub intermediary and destination receiver genuinely separated</summary>

The author implemented Shape B rather than the minimal revert, which is the better outcome: the compose path is now usable instead of merely failing loudly.

**What changed.** `parseTokenFlags` now reads `--dst` and, when the destination is IBC, (a) passes `skipRecipient=true` so `--funds-recipient-dst` is no longer parsed as a Hub bech32 account, and (b) parses `--funds-recipient-hub` into `params.Recipient` as the Hub intermediary, returning `invalid hub recipient: …` when it is empty or malformed. The `dst != DstIBC` gate in `parseHyperlaneFlags` that I originally flagged is now justified rather than incidental: for an IBC destination, `--funds-recipient-dst` is definitively a destination-chain bech32 address and must not be hex-decoded, and `hlParams.RecipientFunds` is correspondingly unused in that branch.

**Blast radius — this was my main re-review concern, and it checks out.** `parseTokenFlags` is a shared helper, so a `--dst`-conditional branch inside it could have leaked into unrelated compose paths. It does not. Grepping every call site:

```
333:  func parseTokenFlags(...)                        ← definition
335:      parseTokenFlagsWithContext(cmd, dst == DstIBC)
349:  func parseTokenFlagsWithContext(...)             ← definition
526:      parseTokenFlagsWithContext(cmd, true)        ← runCreateMemoFromIBC, DstHL
647:      parseTokenFlagsWithContext(cmd, true)        ← runCreateMemoFromHL, DstHL
711:      parseTokenFlagsWithContext(cmd, true)        ← runCreateHLMessageFromKaspa
850:      parseTokenFlags(cmd)                         ← runCreateHLMessageFromHL (SOLE caller)
```

The branching wrapper has exactly one caller, inside `runCreateHLMessageFromHL`. Every other path calls `parseTokenFlagsWithContext` with an explicit `true` and is byte-for-byte unaffected. Two specific hazards I checked and cleared:

- **Missing flag.** `--funds-recipient-hub` is registered by `addKaspaFlags`, which `CmdCreateHLMessage` calls but `CmdCreateMemo` does not. If any `create-memo` path reached the new branch it would fail with `invalid hub recipient: empty address string is not allowed`. None does — `create-memo`'s `DstIBC` branches use only `parseIBCFlags`. The author's earlier container smoke (`create-memo --src=hl --dst=ibc --funds-recipient-dst=cosmos1… --min-amount=80`) therefore still works unchanged, which is correct: the Hub recipient lives in the warp payload, which only the full-message command emits.
- **Kaspa→IBC.** Unaffected (explicit `true` at line 711); it continues to source its Hub recipient through `parseKaspaFlags`.

For `dst != ibc`, `parseTokenFlags` resolves to `parseTokenFlagsWithContext(cmd, false)` with no hub parsing — behaviourally identical to before the PR.

**Regression coverage is the right shape.** `TestComposeIBCMinimum` now drives the destination with `osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw` — the exact non-Hub-prefixed form that failed in my prior-round probe — and asserts both halves of the split:

```go
require.Equal(t, sdk.AccAddress(make([]byte, 20)), decoded.warpPL.GetCosmosAccount())
require.Equal(t, "osmo1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5helwsw", hook.Transfer.Receiver)
```

That is precisely the assertion whose absence let the conflation through last round. `TestHLToIBCRequiresHubRecipient` adds the negative side, covering empty, malformed, and wrong-prefix hub recipients.

**PR body corrected.** The misleading bullet is gone, replaced by an accurate one: *"HL-origin IBC message composition reads the Hub intermediary from `--funds-recipient-hub` and the destination receiver from `--funds-recipient-dst`; regression tests use distinct Hub and `osmo1…` addresses."*

**Docs stayed honest.** Both the `FlagRecipientHub` const comment and its flag help string were extended to name the HL→IBC case. The command's `Example` block contains no `--src=hl --dst=ibc` sample that the new requirement would falsify.

</details>

<details><summary>Finding 2 — Resolved: `decode-hl` renders the IBC floor</summary>

`printForwardToIBC` now prints the field with the same nil-guard idiom as its Hyperlane sibling:

```go
minAmount := "0"
if !m.MinAmount.IsNil() {
    minAmount = m.MinAmount.String()
}
fmt.Printf("  Min Amount:        %s\n", minAmount)
```

Placed between `Receiver` and `Timeout Timestamp`, and the padding is right: every label in that block pads to column 19 (`Source Port:` +7, `Source Channel:` +4, `Receiver:` +10, `Min Amount:` +8, `Timeout Timestamp:` +1), so the new line aligns with the rest. The IBC leg correctly omits the HL block's `(inactive; requires use_full_budget)` annotation — on this leg a positive floor is always active.

Two layers of coverage, which is more than I asked for. `TestIBCMinimumDisplay` exercises the printer directly across absent / zero / positive. `TestComposeIBCMinimum` additionally round-trips through the actual `decode-hl` command on the composed hex and asserts the rendered floor. Note the assertion string `"Min Amount:        "` (8 spaces) cannot collide with the Hyperlane printer's `"Min Amount:         "` (9 spaces), so the round-trip check is genuinely pinned to the IBC block.

</details>

<details><summary>Finding 3 — Resolved: flag registration made independent</summary>

`addMinAmountFlag` is extracted and called from both `addHyperlaneFlags` and `addIBCFlags`, with an idempotency guard:

```go
func addMinAmountFlag(cmd *cobra.Command) {
    if cmd.Flags().Lookup(FlagMinAmount) == nil {
        cmd.Flags().String(FlagMinAmount, "0", "...")
    }
}
```

The guard is load-bearing, not defensive decoration — both `CmdCreateMemo` and `CmdCreateHLMessage` call both registrars, and pflag panics on a redefined flag name. Because both registrars route through the single helper, the default and description cannot drift apart regardless of ordering. `TestIBCMinimumFlagRegistration` covers `{addIBCFlags}` alone and both interleavings, asserting `parseIBCFlags` succeeds with a zero floor — so the "IBC flags without Hyperlane flags" case that previously would have hard-errored is now pinned.

</details>

<details><summary>Finding 4 — Resolved: eIBC-fulfilled budget covered end to end</summary>

`TestFulfillRolToRolBelowMinimum` wraps the registered hook in an `observedIBCCompletionHook` that delegates to the real `RollToIBCHook()` while capturing the budget and the events it emits, then drives a 200-unit transfer with a 150 eIBC fee against a floor of 51. It asserts the observed budget is exactly 50 (the order price), that a single `EventForward` is emitted with `ok=false`, `was_forwarded=true`, and `err` containing `forwardable budget 50 below min_amount 51`.

**The helper's assertions are discriminating, not decorative** — this was worth checking, since a helper written for the happy path could have passed vacuously. `eibcTransferFulfillment` asserts `recipientBal == ibcRecipientBalBefore.Add(priceCoin)` with `price = 200 - 150 = 50`, i.e. the recipient retains the *entire* order price. Had the forward succeeded, those funds would have left the recipient and the assertion would fail. It further asserts `fulfillerBalAfter == fulfillerBalBefore.Add(feeCoin)` after finalization, confirming the rejected forward does not disturb the fulfiller's repayment. So the test would catch a regression that silently stopped enforcing the floor on the fulfilled path.

**No cross-test leakage from the hook override.** The test calls `SetCompletionHooks` to install its observer, which `maps.Copy`s over the one `dym-fwd-roll-ibc` key and is never restored. That is safe here because `eibcForwardSuite.SetupTest` → `eibcSuite.SetupTest` → `ibcTestingSuite.SetupTest` constructs a fresh coordinator and chains — and therefore a fresh app — per test. The arithmetic is also consistent with that setup: `eibcSuite.SetupTest` zeroes `BridgingFee`, which is why 200 − 150 lands cleanly on 50, while `runFinalizeFwdTC` sets it to 1% for the finalization cases.

</details>

<details><summary>Finding 5 — Non-blocking observations</summary>

**(a) `--dst`-conditional coupling now spans two parsers.** There are three `--dst` gates across `parseHyperlaneFlags` (min-amount/full-budget pairing; recipient hex skip) and `parseTokenFlags` (skipRecipient; hub recipient). Each is individually correct and now individually tested, so this is not a defect. But shared parsers re-reading a destination discriminator out of the flag set is the construct that produced the original conflation — an explicit destination parameter threaded from `parseCommonFlags` would make the role split legible at the call site instead of implicit inside three helpers. Worth considering if this area is touched again; not something to do in this PR.

**(b) Pre-existing and out of scope: `create-hl-message --src=hl --dst=hl`.** That command's own `Example` block passes a hex `--funds-recipient-dst=0x1234…`, but `parseTokenFlags` for a non-IBC destination still resolves to `skipRecipient=false` and parses that flag as bech32, which will reject a hex value. This behaviour is byte-identical before and after both commits in this PR — I checked specifically, since the function was edited — so it is neither introduced nor worsened here. Flagging only because the author is already in this file and may want to pick it up separately.

**(c) CI has run no checks on this branch.** `gh pr checks 2223` reports "no checks reported on the 'omridagan/workforce-issue-2222' branch". This is an infrastructure observation, not a proof defect: the PR's proof section names the suites with receipts, and the required checks will gate the merge once they run. Because CI is not currently carrying execution, I ran the one package whose shared parser this commit altered, to confirm the change did not break sibling compose paths:

```console
$ go test -count=1 ./x/forward/cli/
ok  	github.com/dymensionxyz/dymension/v3/x/forward/cli	0.104s
```

That covers the pre-existing `TestMakeForwardToIBCHyperlaneMessage`, `TestLegacyBudgetFloorDisplay`, and `TestIBCMinimumFlags` alongside the new cases. I did not re-run `ibctesting` — the author's named `-run 'TestForwardSuite/TestFulfillRolToRolBelowMinimum'` receipt plus the repository's Go suite cover it.

</details>

---

Proof of execution is well-shaped for the fix round: named suites with receipts for the CLI regressions and the fulfillment case, plus folded transcripts for scoped lint, `gofumpt -l` on the touched files, and `git diff --check`. No proof-artifact URLs to resolve, no QA Brief, no new or re-pointed external dependency call.

VERDICT: APPROVE

